Issue #4, opened 2015-09-23
Improve initial documentation


aggFTW:

alope107:
We should improve the example notebook as part of the documentation.

aggFTW:
Yes... in regards to the notebook and instructions on how to use the magics, I was waiting for you to refactor the API and then do it. Do you think that's fair?

aggFTW:
We need a set of notebooks:
* Spark magic usage
* Pyspark kernel simple usage
* Scala kernel simple usage

Issue #3, opened 2015-09-23
Travis support to have tests run


aggFTW:
Enable Travis support

PR #10, opened 2015-09-24
README file

aggFTW:
Added proposal as README file.

Please take a look @alope107


aggFTW:
@alope107 Sorry for the bad pull request name. I intended to send two separate pull requests, but apparently I'm a rookie in selecting the right branch in my enlistment from Pycharm... :)

I can't seem to get Travis to work for 3.3. I believe it's the packaging I'm doing. I'll look at it tomorrow, unless you have an idea of what's going on.

Thanks!

alope107:
No worries :)
But in the future, it definitely is helpful to have smaller more focused PRs. I think this would make sense as 4 different PRs: one for the travis YAML, one for the README, one for the client manager and one for style fixes.  This makes it a lot easier to review and easier to pin down what's causing tests to fail.

I think your test failure here is due to relative imports.  See http://stackoverflow.com/questions/12172791/changes-in-import-statement-python3

.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: python
+python:
+    - "2.7"
+    - "3.3"
+install:
+    pip install -r requirements.txt -e .
+script: nosetests
```
alope107: Why not use the test script we already have?

remotespark/livyclientfactory.py
```diff
@@ -0,0 +1,34 @@
...
+from base64 import b64encode
+
+from log import Log
+from connectionstringutil import get_connection_string_elements
+from livysession import LivySession
+from livyclient import LivyClient
+from reliablehttpclient import ReliableHttpClient
+
+
+class LivyClientFactory(object):
```
alope107: Why do we need a separate factory class? I think we can move all of this logic to the construction of the LivyClient itself.

remotespark/RemoteSparkMagics.py
```diff
@@ -1,30 +1,29 @@
...
-from livysession import LivySession
-from livyclient import LivyClient
-from reliablehttpclient import ReliableHttpClient
+from clientmanager import ClientManager
+from livyclientfactory import LivyClientFactory
 from log import Log
 								
+
 @magics_class
 class RemoteSparkMagics(Magics):
```
alope107: Nice, this class is much cleaner with all of the client management moved out.

.travis.yml
aggFTW: Travis was complaining about the file not being executable. Decided to do this to unblock me.

remotespark/livyclientfactory.py
aggFTW: This is so that the classes are testable. Basically, if we move the logic to the LivyClient, we cannot have it use a dummy LivySession or make the LivySession use a dummy client.

Futhermore, it helps make the magics class testable by just passing it a dummy factory that produces mockable objects.

There's certainly other ways to achieve the same, but this was they way I chose to do it.

Do you have a certain preference for an alternative method?

aggFTW:
Completely agreed on the 4 PRs part. Again, getting used to my tools =)

.travis.yml
aggFTW: Do we have to chmod the file before we can call it in Travis? If that's the fix, let me tackle that in a different PR.

remotespark/__init__.py
```diff
@@ -1,2 +1 @@
-from .RemoteSparkMagics import *
```
aggFTW: Can you explain what this line was doing? I ran into problems from nosetests because of it.

.travis.yml
alope107: That sounds like that would fix it.

remotespark/__init__.py
alope107: This was just to make sure that everything gets imported and is recognized by IPython. I'm not sure if it's still necessary for newer versions of IPython.

aggFTW: Ok, then I'll remove it for now.

.travis.yml
aggFTW: I'll try that in a separate PR.

alope107:
Looks good to me. Closes #3. Closes #11.

aggFTW:
Thanks!

aggFTW:
Closes #4

PR #12, opened 2015-09-26
Create SQL context for both python and scala sessions

aggFTW:
SQL is not a language in Spark. It's a feature tied to python/scala. Adding sql context for both languages.


alope107:
Looks good to me. We should explore giving users the option to use different types of SQL contexts (HiveQL etc). It might end up being more than we want to put in this package, but it's something we should at least consider.

aggFTW:
Agreed. I'm glad you opened the discussion.
Thanks for reviewing!

PR #14, opened 2015-09-29
make language required for livy endpoint add

aggFTW:
language is now tied to client/session

### DIFF
--- a/remotespark/RemoteSparkMagics.py
+++ b/remotespark/RemoteSparkMagics.py
@@ -28,7 +28,7 @@ class RemoteSparkMagics(Magics):
         self.client_factory = LivyClientFactory()
 
     @magic_arguments()
-    @argument("-l", "--language", help='The language to execute: "scala", "pyspark", "sql", "pysql". Default is "scala".')
+    @argument("-l", "--language", help='The language to execute: "default" for session or "sql". If not provided, "default".')
     @argument("-m", "--mode", help='The mode to execute the magic in: "normal" or "debug". Default is "normal".')
     @argument("-c", "--client", help="The name of the Livy client to use. "
               "Add a session by using %sparkconfig. "
@@ -54,19 +54,17 @@ class RemoteSparkMagics(Magics):
         self.logger.debug("args: " + str(args))
         self.logger.debug("command: " + command)
 
-        # Select language
-        if not args.language:
-            args.language = Constants.lang_scala
-        args.language = args.language.lower()
-        if args.language not in Constants.lang_supported:
-            raise ValueError("Language '{}' not supported.".format(args.language))
-
         # Select client
         if not args.client:
             client_to_use = self.client_manager.get_any_client()
         else:
             args.client = args.client.lower()
             client_to_use = self.client_manager.get_client(args.client)
+
+        # Select language
+        if not args.language:
+            args.language = client_to_use.language
+        args.language = args.language.lower()
 
         # Execute
         print(self._send_command(client_to_use, command, args.language))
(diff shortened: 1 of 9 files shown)


remotespark/RemoteSparkMagics.py
```diff
@@ -161,14 +160,12 @@ def _get_client_keys(self):
...
     def _send_command(self, client, command, language):
-        if language == Constants.lang_scala:
-            return client.execute_scala(command)
-        elif language == Constants.lang_pyspark:
-            return client.execute_pyspark(command)
-        elif language == Constants.lang_sql:
-            return client.execute_scala_sql(command)
-        elif language == Constants.lang_pysql:
-            return client.execute_pyspark_sql(command)
+        if language == "sql":
```
alope107: Why does this no longer use the constants? The language that gets here is going to be scala or python if the user does not specify it in the sparkmagic option, and thus will always raise a language not supported exception.

remotespark/livyclientlib/livyclient.py
```diff
@@ -8,26 +8,19 @@ class LivyClient(object):
...
+        self._session.delete()
 
-    def execute_pyspark_sql(self, command):
-        self._pyspark_session.create_sql_context()
-        return self.execute_pyspark("sqlContext.sql('" + command + "').collect()")
-
-    def close_sessions(self):
-        self._spark_session.delete()
-        self._pyspark_session.delete()
+    def language(self):
```
alope107: Needs a property decorator, currently causes exceptions elsewhere.

alope107:
It looks like PR introduced confusion as to what a language is.  Sessions and clients have a concept of language enumerated in the Constants: scala or python.  However, _send_command is expecting a language of sql or default. This leads to the magic crashing when I don't specify a language when using %%sparkmagic.  It's also strange that the user specifies language multiple times: first when creating the endpoint, and again each time that they run the %%sparkmagic. This semantic overloading is confusing and leading to bugs. We should find a different term for whether SQL is being used or not, and a better way for the user to specify it.

remotespark/RemoteSparkMagics.py
aggFTW: The languages from constants are for the session. Now, when doing code submission, you are using a session that is tied to a language ("default") or the sqlcontext in that language ("sql"). 

You won't get a language not supported exception because that verification is done at session creation time, not code submission.

Did I explain myself?

remotespark/livyclientlib/livyclient.py
aggFTW: That's what happens with late night coding :) I'll sneak that in the next change.

remotespark/RemoteSparkMagics.py
alope107: It's not consistently done that way though. See lines 64-67. If you do not specify a language when using %%sparkmagic, it will take the language of the client/session, which will either be scala or python. This causes the error.  Even if this was done consistently, it's very confusing to have two different meanings for "language" meant to be used at two different places.

aggFTW:
Now sessions are tied to a language; I think that's the right approach to align with Livy and to have a clean design. 

I hear you on the fact that the user might get confused on how he has to specify sql usage on a session. We'll work on that when you combine magics and use widgets.

As for the crash, that's a bug. Reason I didn't catch it was that we have no unit tests for the magics. I have been holding off on adding them until your api redesign comes in. Sending the bug fix for now :)

Agreed @alope107?

alope107:
Heh, I just was about to open a PR for the bugfix myself. We should communicate better next time when we're working on something like that.

I think it's important for both development and for usage that we find different terminology to use than "language" in both cases. What I implemented was a boolean flag --sql that the user would add to the code submission if they wanted to use SQL, otherwise it would go to the language specified by the endpoint.  I think that clears it up nicely in both the code and in usage.

We definitely need magic unit tests, but I agree that it makes sense to hold off on those.

aggFTW:
PR is out, but I can close it if you implemented the different terminology and all. Your choice :)

remotespark/RemoteSparkMagics.py
aggFTW: Yes, it's confusing, which is why I introduced that bug.

PR #18, opened 2015-09-30
Fixing language usage

alope107:
A slightly different approach on how to fix the bugs around language. This sets aside different terminology so that we're not confused as to which language refers to which.


aggFTW:
I like this approach better, thank you!

As for the file change, are you also going to change the travis command for unit tests?

alope107:
Probably should've been done in a different PR, but I just tacked on that change.

Issue #6, opened 2015-09-23
Revise API


alope107:
Consolidate magics and commands.  Clean up UX.

alope107:
I propose that we merge the current two magics, `%sparkconf` and `%%sparkmagic` into a single magic `%%spark`.  This magic supports all the same functionality of the other current magics through a very similar argument passing style. 
If the user omits a required argument, they are prompted for it. In the notebook, this prompt takes the form of IPython widgets (a drop-down selection for language and client, text field for connection string etc). Entering `%%spark` without any arguments thus gives a simple GUI that allows the user to perform any remote spark task without needing to learn the command syntax.
On the command line, we can decide whether it makes more sense to have textual prompts to input the missing arguments or to display an error message.
These proposed changes are designed to make it as easy as possible for a user to interact with spark without having to remember any magic-specific syntax.
@aggFTW What do you think of this idea?

aggFTW:
I like it! I was also thinking of specifying a language for a specific endpoint. So, for adding an endpoint, you would provide url, username, password and language. That will allow us to handle languages differently, which is needed for our integration with automatic visualization.

I'll start the change to take a language for an endpoint now. We can then merge our changes.

Thanks @alope107!

PR #19, opened 2015-09-30
Merged magics

alope107:
Merged the two magics into a single magic named %%spark. Factored out logic into separate methods for easier testing and so that they may be called by an alternate interface, like the planned widgets. Added tests for command parsing. Will need additional testing for actual functionality, and example notebook will need to be redone.

### DIFF
--- a/remotespark/RemoteSparkMagics.py
+++ b/remotespark/RemoteSparkMagics.py
@@ -55,7 +55,7 @@ class RemoteSparkMagics(Magics):
                Delete all Livy endpoints. No arguments required.
                e.g. `%%spark cleanup`
         """
-        usage = "Please look at usage of sparkconf by executing `%spark?`."
+        usage = "Please look at usage of %spark by executing `%spark?`."
         user_input = line
         args = parse_argstring(self.spark, user_input)
 
@@ -99,6 +99,9 @@ class RemoteSparkMagics(Magics):
         
         # Print info after any valid subcommand
         if len(subcommand) > 0:
+            if len(cell) > 0:
+                print("Warning: Cell body not executed because subcommmand found.")
+                print(usage)
             self._print_info()
 
     def run_cell(self, client_name, sql, cell):
--- a/remotespark/RemoteSparkMagics.py
+++ b/remotespark/RemoteSparkMagics.py
@@ -29,94 +29,29 @@ class RemoteSparkMagics(Magics):
 
     @magic_arguments()
     @argument("-s", "--sql", type=bool, default=False, help='Whether to use SQL.')
-    @argument("-m", "--mode", help='The mode to execute the magic in: "normal" or "debug". Default is "normal".')
     @argument("-c", "--client", help="The name of the Livy client to use. "
               "Add a session by using %sparkconfig. "
               "If only one client has been created, there's no need to specify a client.")
     @argument("command", type=str, default=[""], nargs="*", help="Commands to execute.")
     @line_cell_magic
-    def sparkmagic(self, line, cell=""):
+    def spark(self, line, cell=""):
         """Magic to do remote execution of Spark code.
            Arguments should be in line while code should be in cell."""
-        user_input = line
-        args = parse_argstring(self.sparkmagic, user_input)
-        
-        # Change mode
-        previous_mode = Log.mode
-        if args.mode:
-            Log.mode = args.mode
-
-        # Consolidate commands
-        command = cell
-        
-        self.logger.debug("line: " + line)
-        self.logger.debug("cell: " + cell)
-        self.logger.debug("args: " + str(args))
-        self.logger.debug("command: " + command)
-
-        # Select client
-        if not args.client:
-            client_to_use = self.client_manager.get_any_client()
-        else:
-            args.client = args.client.lower()
-            client_to_use = self.client_manager.get_client(args.client)
-
-        # Execute
-        print(self._send_command(client_to_use, command, args.sql))
-
-        # Revert mode
-        Log.mode = previous_mode
-
-    @magic_arguments()
-    @argument("command", type=str, default=[""], nargs="*", help="Command to execute.")
-    @line_magic
-    def sparkconf(self, line, cell=None):
-        """Magic to configure remote spark usage.
-           
-           Usage
-           -----
-
-               %sparkconf [subcommand] [arg]
-           
-           If invoked with no subcommand, subcommand info will be assumed.
-
-           Subcommands
-           -----------
-
-           info
-               Display the mode and available Livy endpoints.
-           mode
-               Set the mode to be used. Possible arguments are: "normal" or "debug".
-               e.g. `%sparkconf mode debug`
-           add
-               Add a Livy endpoint. First argument is the friendly name of the endpoint, second argument
-               is the language, and third argument is the connection string.
-               e.g. `%sparkconf add test python url=https://sparkcluster.example.net/livy;username=admin;password=MyPassword`
-           delete
-               Delete a Livy endpoint. Argument is the friendly name of the endpoint to be deleted.
-               e.g. `%sparkconf delete defaultlivy`
-           cleanup
-               Delete all Livy endpoints. No arguments required.
-               e.g. `%sparkconf cleanup`
-        """
         usage = "Please look at usage of sparkconf by executing `%sparkconf?`."
-        args = parse_argstring(self.sparkconf, line)
+        user_input = line
+        args = parse_argstring(self.spark, user_input)
 
-        # Select subcommand
-        if args.command[0] == "":
-            subcommand = "info"
-        else:
-            subcommand = args.command[0].lower()
+
+        subcommand = args.command[0].lower()
 
         # info
         if subcommand == "info":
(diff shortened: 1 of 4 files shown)


remotespark/RemoteSparkMagics.py
```diff
@@ -124,29 +59,53 @@ def sparkconf(self, line, cell=None):
...
         elif subcommand == "delete":
             if len(args.command) != 2:
                 raise ValueError("Subcommand 'delete' requires an argument. {}".format(usage))
-
             name = args.command[1].lower()
-            self.client_manager.delete_client(name)
-
-            self._print_info()
+            self.delete_endpoint(name)  
         # delete
```
aggFTW: Comment is wrong :) This was an error of mine!

tests/run_tests.cmd
```diff
@@ -1 +1 @@
-nosetests -v
\ No newline at end of file
+ipython -c "import nose; nose.run()"
```
aggFTW: What's the advantage?

aggFTW: Or is it that it's needed for get_ipython() to be available?

tests/test_remotesparkmagics.py
```diff
@@ -0,0 +1,94 @@
...
+def _teardown():
+	pass
+
+@with_setup(_setup, _teardown)
+def test_info_command():
+	mock_method = MagicMock()
+	magic._print_info = mock_method
+	command = "info"
+
+	ip.run_line_magic("spark", command)
```
aggFTW: What would happen if I were to run the spark magic with a subcommand but providing code in the cell.
Do you think that users could possibly get confused if code doesn't run?

tests/test_remotesparkmagics.py
```diff
@@ -0,0 +1,94 @@
...
+	magic.add_endpoint = mock_method
+	command = "add"
+	name = "name"
+	language = "python"
+	connection_string = "url=http://location:port;username=name;password=word"
+	line = " ".join([command, name, language, connection_string])
+
+	ip.run_line_magic("spark", line)
+
+	mock_method.assert_called_with(name, language, connection_string)
```
aggFTW: The unit test here should assert that the client_manager's add method gets called instead.

tests/test_remotesparkmagics.py
```diff
@@ -0,0 +1,94 @@
...
+def test_delete_endpoint_command():
+	mock_method = MagicMock()
+	magic.delete_endpoint = mock_method
+	command = "delete"
+	name = "name"
+	line = " ".join([command, name])
+
+	ip.run_line_magic("spark", line)
+
+	mock_method.assert_called_with(name)
```
aggFTW: The unit test here should assert that the client_manager's delete method gets called instead.

tests/run_tests.cmd
alope107: It's needed for get_ipython.  Nose runs in vanilla python by default.

tests/test_remotesparkmagics.py
```diff
@@ -0,0 +1,94 @@
...
+	mock_method = MagicMock()
+	magic.log_mode = mock_method
+	command = "mode"
+	mode = "debug"
+
+	line = " ".join([command, mode])
+
+	ip.run_line_magic("spark", line)
+
+	mock_method.assert_called_with(mode)
```
aggFTW: The unit test here should assert that the Log's mode is changed.

tests/test_remotesparkmagics.py
```diff
@@ -0,0 +1,94 @@
...
+
+@with_setup(_setup, _teardown)
+def test_cleanup_command():
+	mock_method = MagicMock()
+	magic.cleanup = mock_method
+	command = "cleanup"
+
+	ip.run_line_magic("spark", command)
+
+	mock_method.assert_called_with()
```
aggFTW: The unit test here should assert that the client_manager's clean up method gets called instead.

tests/test_remotesparkmagics.py
alope107: That's a good point I hadn't considered.  We have a couple of options:

* throw an error
* run the subcommand but not the code and display a warning
* run both the subcommand and the code

I'm personally leaning towards displaying a warning.

remotespark/RemoteSparkMagics.py
```diff
@@ -29,94 +29,29 @@ def __init__(self, shell, data=None, mode="normal"):
 
     @magic_arguments()
     @argument("-s", "--sql", type=bool, default=False, help='Whether to use SQL.')
-    @argument("-m", "--mode", help='The mode to execute the magic in: "normal" or "debug". Default is "normal".')
     @argument("-c", "--client", help="The name of the Livy client to use. "
               "Add a session by using %sparkconfig. "
```
aggFTW: No longer true.

tests/test_remotesparkmagics.py
aggFTW: I think so too. Let's do that. We can always change it in the future based on feedback.

tests/test_remotesparkmagics.py
alope107: The tests I've written so far are just testing the parsing of the commands.  I agree that we need to add tests for the actual functionality.

tests/test_remotesparkmagics.py
aggFTW: The metapoint of all of these unit test comments is that we should pass in a client_manager to the magic's constructor and then verify that its methods are being called.

tests/test_remotesparkmagics.py
aggFTW: Ah, that´s what that line meant =)
Are you adding them with this pull request?

aggFTW: Actually, if you add functional tests, parsing could change and you wouldn't care... agreed? Is there value in the parsing tests?

tests/test_remotesparkmagics.py
alope107: We should definitely have additional tests with a mocked client_manager, but it shouldn't be necessary to alter the magic's constructor to do it.

tests/test_remotesparkmagics.py
```diff
@@ -0,0 +1,94 @@
...
+
+@raises(ValueError)
+@with_setup(_setup, _teardown)
+def test_bad_command():
+	command = "bad_command"
+
+	ip.run_line_magic("spark", command)
+
+@with_setup(_setup, _teardown)
+def test_run_cell():
```
aggFTW: The functional test for run cell should test that a client is selected and its execute method called with the given command when:
* client is specified
* client is not specified
* sql is true
* sql is false

tests/test_remotesparkmagics.py
aggFTW: I'm showing my C# ways here by suggesting that it's passed via the constructor. The alternative would be to change the client_manager after construction, yes?

I still think it's cleaner to pass a client_manager or a mock of it depending on the situation with the constructor, letting the magic handle its lifetime, and not altering the state of the magics out of band.

What do you think?

tests/test_remotesparkmagics.py
alope107: It adds a nicer level of granularity.  If we have separate parsing and functional tests, we can more easily identify where the problem is when one fails.  It also allows us to test the methods' functionality directly without all the magic invocation, which will become increasingly important when we have multiple ways to interact with the magic (command options vs. widgets).

aggFTW: Got it. I was under the impression that there was only going to be one way of interacting with the magics, not two.

tests/test_remotesparkmagics.py
alope107: I don't think it makes a huge difference, but I think it's more pythonic to change it after construction.
In general I try to avoid complicating method signatures just to make testing a bit more convenient.

remotespark/RemoteSparkMagics.py
aggFTW: Also, doc string in line 2, were we outline which magics are offered, needs updating.

tests/test_remotesparkmagics.py
aggFTW: Yes, I agree it's more pythonic. Your call. I don't have strong feelings on this case.

alope107:
Added the discussed functional tests and fixed docstring.

tests/test_remotesparkmagics.py
aggFTW: You didn't implement this in the latest commit right? Just making sure I didn't miss it =)

aggFTW:
Closes #6.

tests/test_remotesparkmagics.py
alope107: Implemented now :)

alope107:
I think this is ready to merge now. Please take a final pass @aggFTW

aggFTW:
Git noob here. I apparently did commit your changes but could not close your pull request automatically. Maybe you can show me next time? :) Thanks!

aggFTW:
Do you want to undo the commit and push yourself?

PR #20, opened 2015-10-01
Added ability to return pandas dataframes from SparkSQL for Python and Scala

aggFTW:
With unit tests. Also, made unit tests run faster.


aggFTW:
Closes #9.  Closes #15.

.gitignore
```diff
@@ -58,3 +58,13 @@ target/
...
+
+.idea/misc.xml
+
+.idea/modules.xml
+
+.idea/vcs.xml
+
+.idea/sparkmagic.iml
+
+.idea/workspace.xml
```
alope107: Minor nitpick: these can all be combined to .idea/*

remotespark/RemoteSparkMagics.py
```diff
@@ -62,11 +62,16 @@ def sparkmagic(self, line, cell=""):
             client_to_use = self.client_manager.get_client(args.client)
 
         # Execute
-        print(self._send_command(client_to_use, command, args.sql))
+        result = self._send_command(client_to_use, command, args.sql)
 
         # Revert mode
         Log.mode = previous_mode
 
+        if result is str:
```
alope107: I don't think we need this conditional. A returned string should be displayed by default.

remotespark/livyclientlib/livyclient.py
```diff
@@ -10,17 +10,18 @@ class LivyClient(object):
 
     def __init__(self, session):
         self._session = session
+        self._session.create_sql_context()
```
alope107: How expensive is this operation? It may be worthwhile to lazily delay this until an operation requiring the SQL context is requested.

remotespark/livyclientlib/pandasscalalivyclient.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pandas as pd
+import json
+import re
+
+from .log import Log
+from .livyclient import LivyClient
+
+
+class PandasScalaLivyClient(LivyClient):
```
alope107: It looks like there's a lot of duplicated code between the two pandas clients.  Can we merge them into a single class (the same way we only have a single LivyClient) or at least extract some of their shared functionality into a common place?

.gitignore
aggFTW: Thanks!

remotespark/RemoteSparkMagics.py
aggFTW: Thanks!

remotespark/livyclientlib/livyclient.py
aggFTW: It is kind of expensive and thought about it for a while. Here's my thought process: if we don't do this and the user runs code that creates a sqlContext, next time that he runs a -s True (sql query), he won't find the tables he defined on the context because we would have overridden it. By virtue of instantiating it here, we would provide the guarantee that as long as he doesn't redefine sc or sqlContext, both contexts are available and should work.

remotespark/livyclientlib/pandasscalalivyclient.py
aggFTW: I thought about it too and refrained from doing it just yet. I wanted to hold off until we have a good idea of how Altair integration will look like.

aggFTW:
With this last commit, I think we are good for now, unless you have more comments on the previous points. Please take a look @alope107. Thank you!

remotespark/livyclientlib/livyclient.py
alope107: Fair enough. To reduce the chance of overwriting, we might want to give the default sqlContext a bit more obscure of a name.

remotespark/livyclientlib/pandasscalalivyclient.py
alope107: Fair enough, it shouldn't be much work to refactor this anyway.

alope107:
I have the mentioned minor nitpicks, but I don't think they need to be included in this PR. Looks good to me.

remotespark/livyclientlib/livyclient.py
aggFTW: An obscure name would be awkward to use. I think this is fine for now.

PR #23, opened 2015-10-10
Implement wait_for_state timeout

aggFTW:


alope107:
You may not have meant to push the Untitled notebook in the test directory. You may also want to change the password to your livy server now :)
Other than that, looks good to me.

PR #29, opened 2015-10-17
Added ability to periodically serialize client manager state to disk and read from it

aggFTW:
This will allow people to restart their kernels and have their Livy sessions/endpoints just as they left them.

The client manager now has the capability, but it's turned off. Next PR to come will enable it.

To note, the PR is HUGE. I apologize. If I had known all the changes that are required, I would have partitioned it in smaller chunks.
I didn't know what was going to be required: I had to refactor a lot of relationships between the http_client, the liv_session, the livy_client, and the client_manager, and the need surfaced as I was coding.

I can sit with you and walk you through the commits or you can look at the final diff. Every commit made was left on a clean state where all unit tests were passing.

To mitigate risk, I got test coverage to be >85% and tried with real notebook. Everything looks great.

I intend the next PR to be a separation of the magics class into two classes: magics and a controller.
After that, I'd like to enable serialization for the client manager in a second PR.

Thanks!


aggFTW:
Will help towards #21

aggFTW:
I'll merge to continue working. I can address any comments and can send PRs to address them as needed.

PR #31, opened 2015-10-26
Wrapper Kernels for Python/Scala

aggFTW:
Automatically create a livy session based on env variables. Simplified syntax in notebooks.


aggFTW:
It would be good to include a script that would install this two kernels with the library. What do you think @alope107?

The rationale behind these two kernels is to reduce friction of having to create sessions and having to type %spark in every cell if the user only wants to execute Spark code. This is something we have heard time and again from our users.

This approach gives total flexibility to the user regarding what approach they want to use: kernels or magics.

aggFTW:
Merging for demo now. Can address comments later!

PR #32, opened 2015-10-27
Add Raw/Altair viewers to magics

aggFTW:
This allows the user to visualize their results with Altair or just with as a table (as a pandas df)


aggFTW:
Merging for demo for now. Can address comments later!

PR #42, opened 2015-11-17
Use logging standard library module

msftristew:
This pull request enables support for Python's standard library `logging` module for magics. This greatly increases the configurability of the logging system without adding much additional complexity.  The following files have been changed:

* `log.py` has been completely rewritten to be a light wrapper around the standard library logging module

* A new `FileHandler` class has been added to serve as a "default" log handler

* Configuration files will need a different schema and I've edited `default_config.json` to reflect that. I have removed the `"log_level"` and `"log_to_disk"` settings, which are now both entirely subsumed by the new `"logging_config"` setting. The `logging_config` setting is a JSON object which can be passed to the `logging.config.dictConfig()` function as described [here](https://docs.python.org/2/library/logging.config.html).  This allows users to easily drop-in their own logging behavior.

* Finally, there are some small changes to `constants.py` and `test_logger.py` to accomodate this change.

### DIFF
--- a/remotespark/default_config.json
+++ b/remotespark/default_config.json
@@ -15,8 +15,27 @@
   "kernel_scala_password": "",
   "kernel_scala_url": "",
 
-  "log_level": "debug",
-  "log_to_disk": true,
+  "logging_config": {
+    "version": 1,
+    "formatters": {
+      "magicsFormatter": { 
+        "format": "%(asctime)s	%(levelname)s	%(message)s",
+	"datefmt": ""
+      }
+    },
+    "handlers": {
+      "magicsHandler": { 
+        "class": "remotespark.livyclientlib.log.MagicsFileHandler",
+	"formatter": "magicsFormatter"
+      }
+    },
+    "loggers": {
+      "magicsLogger": { 
+        "handlers": ["magicsHandler"],
+	"level": "DEBUG"
+      }
+    }
+  },
 
   "execute_timeout_seconds": 3600,
   "status_sleep_seconds": 2,
--- a/remotespark/default_config.json
+++ b/remotespark/default_config.json
@@ -20,19 +20,19 @@
     "formatters": {
       "magicsFormatter": { 
         "format": "%(asctime)s	%(levelname)s	%(message)s",
-	"datefmt": ""
+        "datefmt": ""
       }
     },
     "handlers": {
       "magicsHandler": { 
-        "class": "remotespark.livyclientlib.log.MagicsFileHandler",
-	"formatter": "magicsFormatter"
+        "class": "remotespark.livyclientlib.filehandler.MagicsFileHandler",
+        "formatter": "magicsFormatter"
       }
     },
     "loggers": {
       "magicsLogger": { 
         "handlers": ["magicsHandler"],
-	"level": "DEBUG"
+        "level": "DEBUG"
       }
     }
   },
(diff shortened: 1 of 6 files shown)


remotespark/default_config.json
```diff
@@ -15,8 +15,27 @@
   "kernel_scala_password": "",
   "kernel_scala_url": "",
 
-  "log_level": "debug",
-  "log_to_disk": true,
+  "logging_config": {
+    "version": 1,
+    "formatters": {
+      "magicsFormatter": { 
+        "format": "%(asctime)s	%(levelname)s	%(message)s",
```
aggFTW: Why different space quantity in between parts of the format?

msftristew: You mean the indentation? That's fixed in the small fix I just pushed.

remotespark/livyclientlib/filehandler.py
```diff
@@ -0,0 +1,17 @@
+import logging
+from .utils import join_paths, get_magics_home_path, get_instance_id, ensure_path_exists
+
+class MagicsFileHandler(logging.FileHandler):
+    """The default logging handler used by the magics; this behavior can be overriden by modifying the config file"""
+    def __init__(self, **kwargs):
+        # Simply invokes the behavior of the superclass, but sets the filename keyword argument if it's not already set.
+        if 'filename' in kwargs:
```
aggFTW: Can a user specify their own path from the config file?

msftristew: Yes, if you modify the file handler to pass in the "filename" argument.

remotespark/livyclientlib/log.py
```diff
@@ -2,63 +2,26 @@
...
-        if self.level in accepted_modes:
-            message = "{}\t{}\t{}\t{}\n".format(str(datetime.utcnow()), level, self._caller_name, message)
+        self.logger.error(self._transform_log_message(message))
 
-            if get_configuration(Constants.log_to_disk, True):
-                self._write_to_disk(message)
-            else:
-                print(message)
+    def _transform_log_message(self, message):
+        return '{}\t{}'.format(self._caller_name, message)
```
aggFTW: It'd be good to have a unit test that tests that a message includes the caller name every time you call the debug or error method.

msftristew:
Looks like the Travis build is failing because "logging_config" is missing in the config file it's using. Is there something I can do to fix that?

remotespark/default_config.json
msftristew: They aren't spaces, they're actually tab characters.

remotespark/livyclientlib/log.py
msftristew: Unit testing is tough here because there's a tight coupling between the contents of the configuration file and the behavior of the logger. If I modify the contents of the configuration file, how will that interfere with the other tests? Keep in mind that the get_configuration function caches the file contents.

aggFTW: Look at line here:
https://github.com/jupyter-incubator/sparkmagic/blob/89704f35caab700744006b2232f378362b20e532/tests/test_livysession.py#L75

and line 81 afterwards.

msftristew:
Travis is complaining about the tabs in my JSON document. That's funny, it worked in my testing. I'll fix that.

aggFTW:
Everything looks perfect. Merging!

PR #47, opened 2015-12-03
Gracefully handle kernel cannot create livy session in wrapper kernels.

aggFTW:
Kernel will display a fatal failure message and will refuse to run code afterwards.

### DIFF
--- a/remotespark/livyclientlib/livysession.py
+++ b/remotespark/livyclientlib/livysession.py
@@ -68,13 +68,21 @@ class LivySession(object):
 
         self.logger.debug("Starting '{}' sql and hive session.".format(self.language))
 
-        self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
-        self.execute(self._get_sql_context_creation_command())
-        self.logger.debug("Started '{}' sql session.".format(self.language))
+        try:
+            self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
+            self.execute(self._get_sql_context_creation_command())
+            self.logger.debug("Started '{}' sql session.".format(self.language))
+        except LivyClientTimeoutError:
+            raise LivyClientTimeoutError("Failed to create the SQL context in time. Timed out after {} seconds."
+                                         .format(self._create_sql_context_timeout_seconds))
 
-        self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
-        self.execute(self._get_hive_context_creation_command())
-        self.logger.debug("Started '{}' hive session.".format(self.language))
+        try:
+            self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
+            self.execute(self._get_hive_context_creation_command())
+            self.logger.debug("Started '{}' hive session.".format(self.language))
+        except LivyClientTimeoutError:
+            raise LivyClientTimeoutError("Failed to create the Hive context in time. Timed out after {} seconds."
+                                         .format(self._create_sql_context_timeout_seconds))
 
         self._state.sql_context_created = True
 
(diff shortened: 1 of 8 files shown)


remotespark/livyclientlib/livysession.py
```diff
@@ -68,13 +68,21 @@ def create_sql_context(self):
...
 
-        self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
-        self.execute(self._get_sql_context_creation_command())
-        self.logger.debug("Started '{}' sql session.".format(self.language))
-
-        self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
-        self.execute(self._get_hive_context_creation_command())
-        self.logger.debug("Started '{}' hive session.".format(self.language))
+        try:
+            self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
```
msftristew: Will this method `wait_for_status` ever be called with a session status that is not `idle`? If so then you can consider changing the method to `wait_for_idle` and just doing away with the first argument

remotespark/livyclientlib/livysession.py
```diff
@@ -68,13 +68,21 @@ def create_sql_context(self):
...
+            self.execute(self._get_sql_context_creation_command())
+            self.logger.debug("Started '{}' sql session.".format(self.language))
+        except LivyClientTimeoutError:
+            raise LivyClientTimeoutError("Failed to create the SQL context in time. Timed out after {} seconds."
+                                         .format(self._create_sql_context_timeout_seconds))
+
+        try:
+            self.wait_for_status(Constants.idle_session_status, self._create_sql_context_timeout_seconds)
```
msftristew: Copy-paste here. This and the above try-except block should probably be refactored into a separate method

remotespark/sparkkernelbase.py
```diff
@@ -81,19 +60,74 @@ def do_execute(self, code, silent, store_history=True, user_expressions=None, al
...
         return super(SparkKernelBase, self).do_execute(code, silent, store_history, user_expressions, allow_stdin)
 
-    def do_shutdown_ipykernel(self, restart):
+    def _do_shutdown_ipykernel(self, restart):
         return super(SparkKernelBase, self).do_shutdown(restart)
+
+    def _abort_with_fatal_error(self, message):
+        self._fatal_error = message
+
+        error = get_configuration(Constants.fatal_error_suggestion, self.fatal_error_suggestion).format(message)
```
msftristew: From a design perspective I'd say it's more sensible for default values to be handled by the get_configuration function (or, maybe better, a Configuration class).  Having default values for configurations be spread out through the entire codebase seems messy.

remotespark/sparkkernelbase.py
```diff
@@ -28,46 +35,18 @@ def __init__(self, **kwargs):
...
-        self.already_ran_once = True
-
     def do_execute(self, code, silent, store_history=True, user_expressions=None, allow_stdin=False):
+        if self._fatal_error is not None:
+            self._abort_with_fatal_error(self._fatal_error)
+
         if not self.already_ran_once:
-            (username, password, url) = self.get_configuration()
-            self.initialize_magics(username, password, url)
+            (username, password, url) = self._get_configuration()
```
msftristew: Idea: the _get_configuration method is only called once, here, and its output is passed directly to the _initialize_magics method. So maybe _initialize_magics should take no arguments and call _get_configuration itself.

msftristew:
Made some comments.

remotespark/livyclientlib/livysession.py
aggFTW: I don't know. We can make that change later.

remotespark/livyclientlib/livysession.py
aggFTW: Done.

remotespark/sparkkernelbase.py
aggFTW: We can do this later in one pass through the code.

remotespark/sparkkernelbase.py
aggFTW: This was done for unit testing purposes.

PR #62, opened 2015-12-08
Privatepreview0 1

aggFTW:
This should be set as soon as we load magics, so that even session creations that go into error are cleaned up by do_shutdown


msftristew:
+1

PR #68, opened 2015-12-16
"Smart" rendering of objects to standard out, standard error, and conversion to dataframes

msftristew:


aggFTW:
Looked through files quickly. Seems OK but I want to do a deeper dive.

In the meantime, can you please describe the behavior this implements:

1. What are magics supposed to send back as output (e.g. what is O[0] when executing I[0])?
2. What is displayed to the user for each case--what formatting and how many times is the result displayed?
3. Will the write method from shell respect the user setting up their own visualization function for a given datatype? A user might want to have a widget display all of his pandas dataframes.

Thanks!

msftristew:
1. When the magics are a SQL query (%sql or %hive), they send back a dataframe as output.  In that case, the magics don't have any side effect aside from returning the dataframe.  Otherwise (when you're just running some code that isn't a SQL query), the magics don't return anything, but send the strings to stdout or stderr as appropriate.  
2. I'll try to get some screenshots.
3. Yes, this change has no effect on the way that dataframes are visualized.  People can continue to plug in their own dataframe visualizer if they want (and indeed we will be doing so for autoviz with the wrapper kernels)

msftristew:
screen shot 2015-12-16 at 11 52 39 am

msftristew:
The output from the exception seems a little minimal, it should at least have all the exception info.  I'll look into this.

msftristew:
Okay, it turns out that Livy stores only some info in the "evalue" field of the response, and puts more detailed information into the "traceback" field:

    {"id":8,"state":"available","output":
        {"status":"error","execution_count":8,"ename":"ValueError","evalue":"hey",
            "traceback":["Traceback (most recent call last):\n","ValueError: hey\n"]}}

aggFTW:
Thanks for the detailed explanations. Let's add the traceback if we can and I'll continue doing the review.

remotespark/RemoteSparkMagics.py
```diff
@@ -49,11 +46,6 @@ def __init__(self, shell, data=None, test=False):
         except KeyError:
             self.logger.error("Could not read env vars for serialization.")
 
-        if use_auto_viz:
```
aggFTW: use_auto_viz is a configuration that is not needed anymore. Let's remove all traces of it.

tests/test_livysession.py
```diff
@@ -334,7 +334,9 @@ def test_execute(self):
         http_client.post.assert_called_with("/sessions/0/statements", [201], {"code": command})
         http_client.get.assert_called_with("/sessions/0/statements", [200])
         assert_equals(2, http_client.get.call_count)
-        assert_equals(self.pi_result, result)
+        assert isinstance(result, Result)
+        assert isinstance(result, SuccessResult)
```
aggFTW: We need to check for the case isinstance(result, ErrorResult)

remotespark/livyclientlib/pandaslivyclientbase.py
```diff
@@ -31,7 +31,8 @@ def _execute_dataframe_helper(self, context_name, command):
             return records_text
 
     def get_columns(self, context_name, command):
-        return self.execute(self.make_context_columns(context_name, command))
+        return str(self.execute(self.make_context_columns(context_name,
```
aggFTW: If the result of execute is an ErrorResult, wouldn't we have a bug?

remotespark/livyclientlib/pandaspysparklivyclient.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, session, max_take_rows):
 
     def get_records(self, context_name, command, max_take_rows):
         command = '{}.sql("""{}""").toJSON().take({})'.format(context_name, command, max_take_rows)
-        return self.execute(command)
+        return str(self.execute(command))
```
aggFTW: If the result of execute is an ErrorResult, wouldn't we have a bug?

remotespark/livyclientlib/pandasscalalivyclient.py
```diff
@@ -15,7 +16,7 @@ def __init__(self, session, max_take_rows):
 
     def get_records(self, context_name, command, max_take_rows):
         command = '{}.sql("""{}""").toJSON.take({}).foreach(println)'.format(context_name, command, max_take_rows)
-        return self.execute(command)
+        return str(self.execute(command))
```
aggFTW: If the result of execute is an ErrorResult, wouldn't we have a bug?

remotespark/livyclientlib/pandaspysparklivyclient.py
```diff
@@ -24,9 +24,10 @@ def get_columns_dataframe(self, columns_text):
         records = list()
```
aggFTW: We should rename this method to get_columns_dataframe_result

remotespark/livyclientlib/pandaspysparklivyclient.py
```diff
@@ -24,9 +24,10 @@ def get_columns_dataframe(self, columns_text):
...
         columns = eval(columns_text)
 
-        return pd.DataFrame.from_records(records, columns=columns)
+        return DataFrameResult(pd.DataFrame.from_records(records,
+                                                         columns=columns))
 
     def get_data_dataframe(self, records_text):
```
aggFTW: We should rename this method to get_data_dataframe_result

remotespark/livyclientlib/result.py
```diff
@@ -0,0 +1,53 @@
...
+if there is a need. 
+
+Every Result has a "render" method which accepts an IPython kernel as its 
+input and returns the data associated with the Result that the kernel should
+display. The render method may also have the side effect of writing to stdout
+or stderr.
+
+Every result can also be converted to a string with the str function."""
+
+from .utils import write_to_iopub_socket
```
aggFTW: This is not being used in the class?

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -36,15 +38,15 @@ def run_cell(self, client_name, context, cell):
...
-            res = client_to_use.execute_hive(cell)
+            result = client_to_use.execute_hive(cell)
         elif context == Constants.context_name_spark:
-            res = client_to_use.execute(cell)
+            result = client_to_use.execute(cell)
         else:
             raise ValueError("Context '{}' specified is not known.")
 
-        return res
+        return result.render(get_ipython())
```
aggFTW: Do unit tests start throwing when multiple unit tests do a get_ipython call?
If so, you should wrap the call to get_ipython() in its own method so that it's mockable.

remotespark/livyclientlib/utils.py
```diff
@@ -108,3 +108,7 @@ def get_instance_id():
         raise ValueError("Tried to return empty instance ID.")
 
     return instance_id
+
+def write_to_iopub_socket(ipykernel, name, content):
```
aggFTW: Who is using this?

remotespark/livyclientlib/pandaslivyclientbase.py
msftristew: Yes, we need to propagate the exception here.

tests/test_pandaspysparklivyclient.py
```diff
@@ -44,15 +44,20 @@ def test_execute_sql_pandas_pyspark_livy():
...
     execute_m.assert_called_with('sqlContext.sql("""{}""").toJSON().take({})'.format(command, 10))
-
-    # Verify result is desired pandas df
-    assert_frame_equal(desired_result, result)
+    
+    assert isinstance(result, Result)
+    assert isinstance(result, DataFrameResult)
+    mock_shell = MagicMock()
+    df = result.render(mock_shell)
+    assert not mock_shell.write.called
```
aggFTW: Assertions in 58 and 59 should ideally be done in their own unit test and not here. They should be part of a test_results class. Granularity of unit tests will allow us to quickly debug where the problem is.

tests/test_pandaspysparklivyclient.py
```diff
@@ -68,16 +73,20 @@ def test_execute_sql_pandas_pyspark_livy_no_results():
...
-
-    # Verify result is desired pandas dataframe
-    assert_frame_equal(desired_result, result)
-
+    
+    assert isinstance(result, Result)
+    assert isinstance(result, DataFrameResult)
+    mock_shell = MagicMock()
+    df = result.render(mock_shell)
+    assert not mock_shell.write.called
```
aggFTW: Ditto.

tests/test_pandasscalalivyclient.py
```diff
@@ -44,15 +45,20 @@ def test_execute_sql_pandas_scala_livy():
...
 
     execute_m.assert_called_with('sqlContext.sql("""{}""").toJSON.take({}).foreach(println)'.format(command, 10))
 
-    # Verify result is desired pandas df
-    assert_frame_equal(desired_result, result)
+    assert isinstance(result, Result)
+    assert isinstance(result, DataFrameResult)
+    mock_shell = MagicMock()
+    df = result.render(mock_shell)
+    assert not mock_shell.write.called
```
aggFTW: Ditto

tests/test_pandasscalalivyclient.py
```diff
@@ -68,15 +74,20 @@ def test_execute_sql_pandas_scala_livy_no_results():
...
     execute_m.assert_called_with('sqlContext.sql("""{}""").columns'.format(command))
-
-    # Verify result is desired pandas dataframe
-    assert_frame_equal(desired_result, result)
+    
+    assert isinstance(result, Result)
+    assert isinstance(result, DataFrameResult)
+    mock_shell = MagicMock()
+    df = result.render(mock_shell)
+    assert not mock_shell.write.called
```
aggFTW: Ditto

remotespark/livyclientlib/result.py
msftristew: You're right, this is something I forgot to take out before pushing. I'll remove it.

aggFTW:
Could you please post an update on how the change with the traceback looks?

msftristew:
screen shot 2015-12-16 at 4 08 43 pm

aggFTW:
What about adding a \n between hey and Traceback? Personal choice, your call.

msftristew:
For now I'm going to keep it the way it is, particularly because the results from the Scala shell are constructed a bit different, and this looks OK for both Python and Scala.  We might want to keep this in mind in the future.

aggFTW:
Resolves #60.

remotespark/RemoteSparkMagics.py
aggFTW: This is a setting in the default_config.json file as well as in Constants. Can you please remove it from there too?

remotespark/livyclientlib/pandaslivyclientbase.py
```diff
@@ -28,10 +30,11 @@ def _execute_dataframe_helper(self, context_name, command):
                 return self.get_data_dataframe(records_text)
         except (ValueError, SyntaxError):
             self.logger.error("Could not convert sql results to pandas DF.")
-            return records_text
+            raise DataFrameParseException(records_text)
 
     def get_columns(self, context_name, command):
-        return self.execute(self.make_context_columns(context_name, command))
+        return self.execute(self.make_context_columns(context_name,
```
aggFTW: Will the success/failure here cause a ValueError or SyntaxError (except statement in line 31 above)?

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -64,3 +67,11 @@ def add_endpoint(self, name, language, connection_string, skip_if_exists):
 
     def get_client_keys(self):
         return self.client_manager.get_endpoints_list()
+
+    def get_client_by_name(self, client_name = None):
```
aggFTW: Why would the get_client_by_name method make the name optional?

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -4,19 +4,22 @@
...
 from .constants import Constants
-
+from .dataframeparseexception import DataFrameParseException
 
 class SparkController(object):
 
     def __init__(self, serialize_path=None):
         self.logger = Log("SparkController")
         self.client_factory = LivyClientFactory()
+        self.ipython = get_ipython()
```
aggFTW: I thought we'd said that all ipython calls would be made in the magics class?

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -24,27 +27,27 @@ def __init__(self, serialize_path=None):
...
 
-    def run_cell(self, client_name, context, cell):
-        context = context.lower()
-
-        # Select client
-        if client_name is None:
-            client_to_use = self.client_manager.get_any_client()
+    def run_cell(self, cell, client_name = None):
+        client_to_use = self.get_client_by_name(client_name)
+        (success, out) = client_to_use.execute(cell)
```
aggFTW: I thought `(success, out)` should be returned to magics and the if/else below should be executed in the magics class?

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -24,27 +27,27 @@ def __init__(self, serialize_path=None):
...
-        elif context == Constants.context_name_spark:
-            res = client_to_use.execute(cell)
-        else:
-            raise ValueError("Context '{}' specified is not known.")
+    def run_cell_sql(self, cell, client_name = None):
+        client_to_use = self.get_client_by_name(client_name)
+        try:
+            return client_to_use.execute_sql(cell)
+        except DataFrameParseException as e:
+            self.ipython.write_err(e.out)
```
aggFTW: viewer should decide what to do with the result/exception.

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -24,27 +27,27 @@ def __init__(self, serialize_path=None):
...
 
-        return res
+    def run_cell_hive(self, cell, client_name = None):
+        client_to_use = self.get_client_by_name(client_name)
+        try:
+            return client_to_use.execute_hive(cell)
+        except DataFrameParseException as e:
+            self.ipython.write_err(e.out)
```
aggFTW: Ditto.

remotespark/RemoteSparkMagics.py
msftristew: Yep

remotespark/livyclientlib/pandaspysparklivyclient.py
msftristew: I forgot to incorporate this rename into my last commit. I can go ahead and do that as well

remotespark/livyclientlib/sparkcontroller.py
msftristew: Maybe I got mixed up, let me have a look.

remotespark/livyclientlib/pandaslivyclientbase.py
msftristew: You're right, the error handling here isn't right. What the helper methods (get_columns_dataframe, get_data_dataframe) should do is throw a DataFrameParseException themselves if they can't parse correctly.

msftristew:
Okay, this PR should be finalized. There's a remaining task to tighten up the behavior on error conditions for which I will make an issue now.

aggFTW:
Cool, merge :)

Once that task on error conditions is implemented, let's write down the way data flows through the classes for any new dev on-boarding the project.

PR #70, opened 2015-12-17
Add application name to session creation payload

aggFTW:
Checking in. I can address any comments later. @msftristew


msftristew:
Looks fine to me!

Issue #64, opened 2015-12-11
Rethink configuration


msftristew:
This is part of a greater architectural issue around configurations. There should be a configuration module which:
1. Automatically loads data from a configuration file,
2. loads the configuration with an appropriate set of defaults if things are missing from the configuration file, and
3. allows a developer to substitute a different configuration module if necessary (i.e. for tests).

aggFTW:
Goals:

1. break configuration settings into different files per project
2. have defaults in centralized place
3. have get_config method return defaults from the centralized place
4. be able to easily unit test without config files
5. allow config changes at runtime for unit testing purposes

msftristew:
PR #77

PR #77, opened 2015-12-18
Rethink configuration

msftristew:
This change improves the extensibility and functionality of the configuration module. The following changes have been made:

* Begin reorganizing the repository by creating a new `remotespark.utils` subdirectory, under which some common utilities are stored.

* Change the API of the configuration module. The module now supports the following methods:

        # loads a set of configuration overrides from the config file on disk
        configuration.load() 
        # initializes the set of configuration overrides, loading iff the module has not already been initialized
        configuration.initialize()
        # given a dict which is a set of configuration overrides, override those and only those configurations
        # this method subsumes the old hacky _t_config_hook function
        configuration.override(object)

* The configuration module now supports a function for each configuration. e.g. there is a function `configuration.default_chart_type()`, a function `configuration.serialize()`, a function `configuration.status_sleep_seconds()`, etc.  These all come with a default value if they were not overridden by user settings.

* Some slight cleanup regarding the schema of configuration files (`default_config.json`)

* Appropriate changes to the rest of the codebase and to the Constants class to accomodate these changes

These tasks remain:

* Configuration is not split between magics/kernels/Livy client lib.  That refactoring can be done in a later pull request.

* Some further changes to the directory structure of the package are probably still needed.  This is simply an incremental step in the correct direction.


msftristew:
Looks like a client manager serialization test is failing.  I'll have a look.

aggFTW:
I don't like this part at all:
`The configuration module now supports a function for each configuration`

That means that we'd have to add a new function per configuration. Can we instead read the defaults from the default_config.json file?

So, `configuration.initialize()` would read `default_config.json`.
Then, `configuration.load(path_to_file)` would read the configuration file and override anything previously loaded by `default_config`.
Then, you could do a `configuration.override(dictionary)` to override specific values.
And last, you could do `configuration.get(config_name)`.

msftristew:
One potential issue with that is that you now absolutely need to have a `default_config.json` file in place with the correct schema, or else nothing will work.  If the user deletes, loses track of, or accidentally makes a bad edit to their default_config file they'll be blocked.  I don't know if this is a real concern in practice.

tests/test_clientmanager.py
```diff
@@ -28,7 +29,7 @@ def test_deserialize_on_creation():
 
 
 def test_serialize_periodically():
-    _t_config_hook({"serialize_period_seconds": 0.1})
+    conf.override({"serialize_period_seconds": 0.1, "serialize_periodically": True})
```
aggFTW: Suggestion: refer to the method's __name__ property here, so that if we change the name, a refactor tool will change the name here too.

remotespark/utils/configuration.py
```diff
@@ -0,0 +1,155 @@
...
+def initialize():
+    """Checks if the configuration is initialized. If so, initializes the
+    global configuration object by reading from the configuration
+    file, overwriting the current set of overrides if there is one"""
+    global _overrides
+    if _overrides is None:
+        load()
+
+
+def load():
```
aggFTW: Please add unit test for load

remotespark/utils/configuration.py
```diff
@@ -0,0 +1,155 @@
...
+        line = "".join(config_text).strip()
+        if line == "":
+            overrides = {}
+        else:
+            overrides = json.loads(line)
+    override(overrides)
+
+
+def override(obj):
+    """Given a dictionary representing the overrided defaults for this
```
aggFTW: Please add unit test for override

remotespark/utils/configuration.py
```diff
@@ -0,0 +1,155 @@
...
+    configuration, initialize the global configuration."""
+    global _overrides
+    _overrides = obj
+
+
+def _override(f):
+    """A decorator which first initializes the overrided configurations,
```
aggFTW: Please add unit test for the decorator

aggFTW:
Let's try this approach and see how we like it. Just address the unit test comments and it's good to go.

Thanks!

aggFTW:
Closes #64

Issue #66, opened 2015-12-12
wait for state doesn't return immediately if state is final


aggFTW:
When state goes into a final state, like error, wait for state should immediately return.

aggFTW:
It should throw if state is final and not what is expected. If state is error, explanation should go back to the user.

msftristew:
PR #81

PR #81, opened 2015-12-19
Fail eagerly to wait for the idle status if we have reached a "final" status

msftristew:
* Rename `wait_for_status()` to `wait_for_idle()` and adjust parameter list appropriately.  If we ever need to wait for a status besides idle (seems unlikely?), then this can be refactored back out again.

* Throw a `LivyUnexpectedError` if the Livy session goes into error or dead.

### DIFF
--- a/remotespark/livyclientlib/livyunexpectederror.py
+++ b/remotespark/livyclientlib/livyunexpectederror.py
@@ -1,0 +1,3 @@
+class LivyUnexpectedError(Exception):
+    """An exception that will be shown if some unexpected error happens on the Livy side."""
+    pass
--- a/remotespark/livyclientlib/livyclient.py
+++ b/remotespark/livyclientlib/livyclient.py
@@ -21,7 +21,7 @@ class LivyClient(object):
         return self._session.get_state().to_dict()
 
     def execute(self, commands):
-        self._session.wait_for_status("idle", self._execute_timeout_seconds)
+        self._session.wait_for_idle(self._execute_timeout_seconds)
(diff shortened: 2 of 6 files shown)


msftristew:
1. Output will look similar to when a `LivyClientTimeoutError` is thrown.

2. I am unsure a) if we can get useful error info from Livy when we get this kind of error and b) how we might capture that useful error info and display it to the user.  This merits further discussion.  However, the current error message is "good enough" IMO.

remotespark/livyclientlib/livyunexpectederror.py
```diff
@@ -0,0 +1,3 @@
+class LivyUnexpectedError(Exception):
+    """An exception that will be shown if some unexpected error happens on the Livy side."""
```
aggFTW: Can we rename to LivyUnexpectedFinalStatusErrors? This seems too broad.

aggFTW:
If exception is in error, session will have an extra field in the json payload that comes back. Please include that for errored sessions.

remotespark/livyclientlib/livyunexpectederror.py
msftristew: My idea was that this exception could be used for other uses besides just an unexpected final status but this is OK too.

aggFTW:
Please include Closes #66 when addressing a bug. That will close it automatically on merge.

msftristew:
What is the name of the extra field?

aggFTW:
Lgtm! Merging

PR #84, opened 2015-12-19
Allow user to specify Livy session configs

### LINKED ISSUE #72: Allow user to specify memory/cores/etc for every session

aggFTW:
Closes #72

Request Body from POST request in /sessions as documented by Livy:
https://github.com/cloudera/livy/tree/6fe1e80cfc72327c28107e0de20c818c1f13e027

### DIFF
--- a/remotespark/remotesparkmagics.py
+++ b/remotespark/remotesparkmagics.py
@@ -7,6 +7,7 @@ Provides the %spark magic."""
 from __future__ import print_function
 from IPython.core.magic import Magics, magics_class, line_cell_magic, needs_local_scope
 from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+import json
 
 import remotespark.utils.configuration as conf
 from remotespark.utils.constants import Constants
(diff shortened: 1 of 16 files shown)


msftristew:
Looks good to me, +1

remotespark/remotesparkmagics.py
```diff
@@ -99,18 +110,29 @@ def spark(self, line, cell="", local_ns={}):
         # info
         if subcommand == "info":
             self._print_info()
+        # config
+        elif subcommand == "config":
+            # Would normally do " ".join(args.command[1:]) but parse_argstring removes quotes...
+            rest_of_line = user_input[7:]
+            self.properties = json.loads(rest_of_line)
```
msftristew: So it looks like the design we have here is that `self.properties` is the "source of truth" for configurations, unlike other configurations where the functions that come from the configuration.py module are the "source of truth".  I don't necessarily think this is an issue right now which is why I +1'd but it's something to keep in mind for the future.

PR #87, opened 2015-12-21
Separated load magics from start session

aggFTW:
This will make it easier to add configs

### DIFF
--- a/remotespark/sparkkernelbase.py
+++ b/remotespark/sparkkernelbase.py
@@ -9,35 +9,40 @@ from remotespark.utils.utils import get_connection_string
 
 
 class SparkKernelBase(IPythonKernel):
-    # Required by Jupyter - Override
-    implementation = None
-    implementation_version = None
-    language = None
-    language_version = None
-    language_info = None
-    banner = None
+    def __init__(self, implementation, implementation_version, language, language_version, language_info,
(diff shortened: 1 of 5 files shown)


remotespark/sparkkernelbase.py
```diff
@@ -9,35 +9,40 @@
...
+
         self.logger = Log(self.client_name)
-        self.already_ran_once = False
+        self.session_started = False
         self._fatal_error = None
 
         # Disable warnings for test env in HDI
         requests.packages.urllib3.disable_warnings()
 
+        if "testing" not in kwargs.keys():
```
msftristew: Try `if not kwargs.get("testing", False)`.  This will work even if I pass in `testing=False` as a keyword argument

msftristew:
I'm confused about the nature of the change you made.  In what way did you "separate load magics from start session"?  Can you point me to the block of code which does this?

aggFTW:
_start_session and _initialize_magics used to be the same.

remotespark/sparkkernelbase.py
aggFTW: Already made the change but will add in next PR.

msftristew:
I see.  Looks good, I sign off

PR #91, opened 2015-12-22
Improve error message when using config from wrapper kernel and fix %spark info to show actual info

aggFTW:


aggFTW:
Very very small PR. Merging to keep working. I can address comments later. This came from using magics against an actual Livy endpoint and realizing there were some usability issues.

PR #92, opened 2015-12-22
Allow to override a single configuration

msftristew:
This paves the way for being able to update arbitrary configuration values at runtime, including those passed in by the user.


aggFTW:
lgtm! Merging

PR #102, opened 2016-01-07
Fix typo which causes Python kernel not to recognize empty dataframe results

msftristew:
Now the Python kernel should be able to properly handle empty dataframes, which it wasn't able to do before thanks to a typo.


msftristew:
Closing.

PR #104, opened 2016-01-08
Syntax highlighting for Python kernel

msftristew:
Add syntax highlighting for the Python kernel.  This one's very easy.  We can probably use the same general solution for Scala but I'm not sure how to do it right now.


msftristew:
When I try the same solution with Scala, nothing works but no 404 errors are triggered Jupyter side either.  Going to keep investigating.

PR #105, opened 2016-01-08
Scala syntax highlighting

msftristew:
Enable syntax highlighting for Scala. Making this work required that I change the `name` field of the `language_info` for the Spark kernel. I don't believe this has any other repercussions, but I'll wait for Alejandro's +1 before merging this in.


aggFTW:
This seems fine. +1

PR #107, opened 2016-01-09
Fix value and label computation for pie graph

msftristew:
Now values and labels for pie graphs should be computed correctly.

### DIFF
--- a/remotespark/datawidgets/plotlygraphs/piegraph.py
+++ b/remotespark/datawidgets/plotlygraphs/piegraph.py
@@ -9,7 +9,7 @@ class PieGraph(object):
     @staticmethod
     def render(df, encoding, output):
         series = df.groupby([encoding.x]).size()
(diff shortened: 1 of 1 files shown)


msftristew:
Going to merge after tests pass because this is a pretty small and innocuous change

remotespark/datawidgets/plotlygraphs/piegraph.py
```diff
@@ -8,7 +8,8 @@
 class PieGraph(object):
     @staticmethod
     def render(df, encoding, output):
-        data = [Pie(values=PieGraph._get_x_values(df, encoding))]
+        series = df.groupby([encoding.x]).size()
```
aggFTW: Could we put this in its own method and unit test it?

msftristew: Good idea

PR #115, opened 2016-01-13
Fix performance issues for data/pie graphs

### LINKED ISSUE #96: Investigate perf issues with auto_viz
Can all graph types handle ~2500 result rows?

### LINKED ISSUE #100: Incorrect visualizations on some sample data
Ran the following code:

    hvac = sc.textFile('wasb:///HdiSamples/HdiSamples/SensorSampleData/hvac/HVAC.csv')
    from pyspark.sql import Row
    Doc = Row("TargetTemp", "ActualTemp", "System", "SystemAge", "BuildingID")
    def parseDocument(line):
        values = [str(x) for x in line.split(',')]
        return Doc(values[2], values[3], values[4], values[5], values[6])
    documents = hvac.filter(lambda s: "Date" not in s).map(parseDocument)
    df = sqlContext.createDataFrame(documents)
    df.registerTempTable('data')

and then

    %select * from data limit 100

The visualizations, at least for the pie graphs, are wrong.  Screenshot:

screen shot 2016-01-06 at 6 18 47 pm

Clearly there is no building where the desired target temperature is 1.

aggFTW:
Closes #100
Closes #96 

I would've liked to add unit tests for the behaviors of the render methods of DataGraph and PieGraph, but due to how the output widget is written, code would increase in complexity. As it is, code is simple and regressions shouldn't be that likely.


remotespark/datawidgets/ipythondisplay.py
```diff
@@ -0,0 +1,11 @@
+from IPython.core.display import display, HTML
+
+
+class IpythonDisplay(object):
```
msftristew: This shouldn't be a class.  `display_to_ipython` and `html_to_ipython` should just be top-level functions in this module.

msftristew: This should be a class if and only if IpythonDisplay is an instantiable/mockable object, but it doesn't look like it is.

aggFTW: It's not at the moment, but it's the type of methods that could very well be. My intention was to mock it when called from `with o:`... I will reconsider adding that unit test tomorrow, which means it will get mocked.

msftristew:
When the IpythonDisplay issue gets resolved (i.e. either turn it into a class which can be mocked or into a module with some public top-level functions) then I sign off.

remotespark/datawidgets/ipythondisplay.py
```diff
@@ -0,0 +1,11 @@
...
+    @staticmethod
+    def display_to_ipython(to_display):
```
msftristew: Should these still be static methods with the new design in mind?

aggFTW: Why not?

msftristew:
signoff

Issue #53, opened 2015-12-04
Syntax highlighting


aggFTW:
For Scala and R support, we may want to look at http://pygments.org/languages/

For SQL support, look at @alope107's solution in https://github.com/alope107/spark-sql-magic/commit/34b4d864cdaaa0cc56481b3045a66bf679141f78

msftristew:
Scala support is in CodeMirror through this file, clike.js: https://github.com/codemirror/CodeMirror/blob/master/mode/clike/clike.js

However, when I set the mode to `clike` for Scala, I don't get correct highlighting/text editing behavior (none of the correct keywords are highlighted, and default indentation is incorrect unless I put a semicolon at the end of all my lines).  Going to keep investigating.

msftristew:
Python taken care of through #104

msftristew:
In my testing, Scala isn't working even if I set the codemirror mode to "clike" and the MIME type to `text/x-scala`.  For some reason CodeMirror isn't "getting" that what I type in is supposed to be Scala code and is instead applying some very simple, stupid syntax highlighting to it (i.e. only highlighting equals signs and integer literals).  Can keep looking at this later.

aggFTW:
Hey @msftristew, any updates for this?

msftristew:
Scala and Python syntax highlighting are both working great.  I haven't tested out SQL but it should be easy (@alope107's solution should be fine here)

msftristew:
There's an issue.  See https://github.com/ipython/ipython/issues/7548

It looks like only cell magics trigger magics syntax highlighting. I don't think that syntax-highlighted line magics are supported.  This means that our SQL and Hive commands need to be %%sql and %%hive in order for syntax highlighting to kick in -- %sql and %hive won't work.

aggFTW:
Let's go with that for now and we'll investigate line magics in parallel.

msftristew:
Arguably all of the "magics" in our Spark kernel are cell magics, not line magics. So it breaks with convention less if we move to a model where `%%sql`, `%%hive`, `%%config`, etc. are the "normal" way to call things.

msftristew:
Still need to do this :(

msftristew:
So, I have a branch going (kerneljs) which is supposed to fix SQL syntax highlighting.  As it stands, on my Windows machine, SQL syntax highlighting works for the first cell but fails with an error message in the console for the rest.

[...]

PR #116, opened 2016-01-13
SQL syntax highlighting

msftristew:
We need to add kernel.js files for the kernels. They have the same content.

### DIFF
--- a/remotespark/pysparkkernel/kernel.js
+++ b/remotespark/pysparkkernel/kernel.js
@@ -1,4 +1,1 @@
 IPython.CodeCell.config_defaults.highlight_modes['magic_text/x-sql'] = {'reg':[/^%%sql/, /^%%hive/]};
-
-
-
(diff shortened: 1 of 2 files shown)


msftristew:
Closes #53

remotespark/pysparkkernel/kernel.js
```diff
@@ -0,0 +1 @@
+IPython.CodeCell.config_defaults.highlight_modes['magic_text/x-sql'] = {'reg':[/^%%sql/, /^%%hive/]};
```
rgbkrk: Have you figured out a way to use the hive dialect here?

aggFTW: You mean syntax highlight HQL? We just use the SQL lexer.
Sorry if my answer isn't clear, but I'm not sure I understood. Use it for what?

rgbkrk: The default here is the ANSI SQL dialect, [there's also a hive mime mode that supports additional keywords with the Hive SQL dialect](https://github.com/codemirror/CodeMirror/blob/c4387d6073b15ccf0f32773eb71a54f3b694f2f0/mode/sql/sql.js#L394-L402).

aggFTW: Cool! Did you want to change it to Hive? We'll get to that later (no one's asked for it yet), but nice to know it exists!

Issue #51, opened 2015-12-04
Print progress while session is being created


aggFTW:
Give the user some kind of feedback so that user knows that code is not frozen.

msftristew:
Here's a module for rich progress bars: https://pypi.python.org/pypi/ipy-progressbar/1.0.2

Since Livy doesn't give us rich info surrounding session creation progress, this is probably overkill.  A few short sentences that say something along the lines of "Contacting Livy to start session... Waiting for Livy session to return to idle state..." is probably fine.

PR #118, opened 2016-01-13
Print progress while session is being created

MohamedElKamhawy:
I'm showing some messages while creating:
- Livy Session
- Creating SQL Context
- Creating Hive Context
Then, after successfully creating them.


remotespark/livyclientlib/livysession.py
```diff
@@ -3,7 +3,7 @@
 
 import textwrap
 from time import sleep, time
-
+from IPython import get_ipython
```
aggFTW: Remove any mention to IPython from within Livy classes. They should be agnostic to the view.

remotespark/livyclientlib/livysession.py
```diff
@@ -154,7 +158,7 @@ def wait_for_idle(self, seconds_to_wait):
         Parameters:
             seconds_to_wait : number of seconds to wait before giving up.
         """
-
+        #get_ipython().write('Testing\n')
```
msftristew: No committing commented out code :)

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -70,7 +71,10 @@ def add_session(self, name, connection_string, skip_if_exists, properties):
 
         session = self.client_factory.create_session(connection_string, properties, "-1", False)
         session.start()
-        livy_client = self.client_factory.build_client(session)
+
+        msg_printer = MessagePrinter.MessagePrinter()
```
msftristew: This doesn't work, this would tie livyclientlib to IPython.  You need to make this object an argument to the method so you can create it elsewhere and mock it.

remotespark/livyclientlib/livyclientfactory.py
```diff
@@ -18,14 +18,14 @@ def __init__(self):
...
         self.max_results = conf.max_results_sql()
 
-    def build_client(self, session):
+    def build_client(self, session, msg_printer):
         assert session is not None
         kind = session.kind
 
         if kind == Constants.session_kind_pyspark:
-            return PandasPysparkLivyClient(session, self.max_results)
+            return PandasPysparkLivyClient(session, msg_printer, self.max_results)
```
aggFTW: Why are there no modification to `PandasPysparkLivyClient` and `PandasScalaLivyClient`?

remotespark/sparkkernelbase.py
```diff
@@ -54,16 +56,17 @@ def do_execute(self, code, silent, store_history=True, user_expressions=None, al
             self._abort_with_fatal_error(self._fatal_error)
 
         subcommand, flags, code_to_run = self._parse_user_command(code)
+        msg_printer = MessagePrinter.MessagePrinter()
```
msftristew: I don't understand why you're creating one MessagePrinter in SparkKernelBase and another in SparkController. Why wouldn't you be using the same one all the time?

remotespark/livyclientlib/sparkcontroller.py
aggFTW: +1

msftristew:
Make sure that you run unit tests before pushing anything.

remotespark/livyclientlib/livysession.py
msftristew: Looks like this was left in by accident -- it's not necessary

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Removed

remotespark/sparkkernelbase.py
aggFTW: Furthermore, please import like:

`from remotespark.messageprinter import MessagePrinter`
so that instantiation looks like:
`msg_printer = MessagePrinter()`

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Yup, removed.

remotespark/MessagePrinter.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2015  aggftw@gmail.com
```
aggFTW: Files should only have lowercase characters. DO NOT check this in like it is. It's a pain to change it later.

aggFTW: Also, I've created a very similar class... Let's work on merging our classes. Let me check it in and show you how to merge it.

msftristew:
In the `sparkkernelbase` class, there are some methods `_send_error()` and `_ipython_send_error()`.  Maybe we could move that functionality into the `MessagePrinter` so that that's the only facility we use for writing output and errors to IPython.

examples/remotespark.ipynb
```diff
@@ -3,9 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
```
aggFTW: Please remove all changes to this example notebook.

remotespark/sparkkernelbase.py
```diff
@@ -134,9 +137,10 @@ def _register_auto_viz(self):
                            log_if_error="Failed to register auto viz for notebook.")
         self.logger.debug("Registered auto viz.")
 
-    def _start_session(self):
+    def _start_session(self, msg_printer):
```
msftristew: `msg_printer` shouldn't be an argument to all these methods -- it should be an instance variable.  

i.e. in the constructor you say `self._msg_printer = MessagePrinter()` and then reference that instance variable in the methods which need it.

remotespark/livyclientlib/livyclientfactory.py
MohamedElKamhawy: Why should we make a modification there? I didn't find a constructor for the base class.

remotespark/MessagePrinter.py
MohamedElKamhawy: I changed it to lower case name.
Ok, let's check them.

aggFTW:
Couple of comments overall:

* Existing unit tests are not passing. Please try to never send a PR where unit tests are not passing.
* Where's the unit tests for the new functionality? Reasons to add unit tests:
	You want to write unit tests because they are useful to you and to the team. They ensure quality and speed everyone up. I've been coached to write unit tests mostly for three reasons, and I've found that writing unit tests is extremely useful (to the tune that I don't mind spending ~50% of my time writing them):

	1. To test key scenarios.
	If you are writing code that is supposed to do something, make sure that it does just that! Especially when handling edge/tricky cases.

	2. To prevent regressions.
	If there's one tricky piece of code that could be easily broken or if there's some non-obvious guarantees your code needs for it to work, you might want to add tests so that other people (or yourself) don't break that behavior in the future when modifying or adding code.

	3. To speed up development.
	I'm not too good about this myself, but I've found that when I write unit tests in advance, I usually end up with better design/abstractions. When I do them after I code everything up, I sometimes find that I need to do some refactoring to easily test things, and this refactoring pays off in that my code is maintainable: I build extension points and I can refactor classes/methods knowing that any bug I introduce in other parts will be caught by existing unit tests.

	Writing unit tests will force you to write code in a way that abstracts dependencies out. This is done via a variety of tools/patterns, but I've found that in Python, the mock package gets 95% of the work done. We can talk about testable code more if you want to.
* The class to interact with IPython could be restructured a little bit and use through the code base like Ricky suggested. Let's work on this before you send the next iteration. I'll drop by.

MohamedElKamhawy:
That's great comment, Alex. :)

I do know the importance of the unit tests, but we have to add this to a programming guide to the new people landing on the team. This will help in avoiding this case.

remotespark/livyclientlib/livysession.py
```diff
@@ -68,16 +68,20 @@ def start(self):
...
 
-    def create_sql_context(self):
+    def create_sql_context(self, msg_printer):
         """Create a sqlContext object on the session. Object will be accessible via variable 'sqlContext'."""
         if self.started_sql_context:
             return
 
         self.logger.debug("Starting '{}' sql and hive session.".format(self.kind))
 
+        msg_printer.print_message('Creating SQL Context...')
```
aggFTW: Can you please add the name of the context?

remotespark/livyclientlib/livysession.py
```diff
@@ -68,16 +68,20 @@ def start(self):
...
         self._create_context(Constants.context_name_sql)
+
+        msg_printer.print_message('Creating Hive Context...')
```
aggFTW: Can you please add the name of the context? No ellipsis.

remotespark/livyclientlib/livysession.py
```diff
@@ -68,16 +68,20 @@ def start(self):
...
         self._create_context(Constants.context_name_hive)
 
+        msg_printer.print_message('Contexts have been created successfully!')
```
aggFTW: This is not needed.

remotespark/livyclientlib/livysession.py
aggFTW: Also, what about adding some output to `wait_for_idle`? I was thinking maybe adding some dots every time the method is run and it's not in idle, so it'd look something like:

```
Creating SqlContext as 'sqlContext'.
.....
Creating HiveContext as 'hiveContext'.
.........
```

and for execute methods it'd look something like:
```
.....................
res0: DataFrame[column1, column2]
```

remotespark/livyclientlib/livysession.py
msftristew: +1

remotespark/livyclientlib/livysession.py
msftristew: If we implemented this, I would definitely want the dots to be dismissed after we're done waiting for idle, since they're just noise.

It looks like we're only calling `wait_for_idle` to wait before creating the SQL context; we never actually call that method before cell executions.  We could add another call to this method before cell executions if you wanted to implement this.

I'm not sure if I like the idea all that much myself.  Usually dots that programs pop up suggest the program is busy doing some useful work, when the opposite is true here.  It might be confusing for users if they sporadically get 1 or 2 dots, and then get no more dots for another 10 minutes before their result is displayed. 

What is it that you would want to communicate with output text like this?  I think it might be better to identify those scenarios, recognize when they happen, and pop up with disposable text messages that capture what's happening for the user (e.g. "This query is taking longer than usual", etc.)

MohamedElKamhawy: I do like that, I think we should open a new enhancement to track this work, it'll be nice from the user perspective to see a meaningful waiting statement.

MohamedElKamhawy: Also, I prefer the dots to appear then being dismissed like any waiting message, that'll be very nice.

aggFTW: Mohamed, can you open a bug please to add dismissable dots during statement execution?

msftristew:
In remotesparkmagics.py, please have the class call `self.ipython_display.write()` instead of `self.shell.write()`

msftristew:
You should probably add some asserts to the `test_livysession` file to verify that `ipython_display.writeln()` is being called when the `start()` method is called on LivySessions

msftristew:
Looks good otherwise, let's just wait for Alejandro to sign off

Untitled.ipynb
```diff
@@ -0,0 +1,52 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
```
aggFTW: Please remove this file.

examples/remotespark.ipynb
aggFTW: Again, please revert changes.

remotespark/livyclientlib/livysession.py
```diff
@@ -75,7 +75,10 @@ def create_sql_context(self):
 
         self.logger.debug("Starting '{}' sql and hive session.".format(self.kind))
 
+        self.ipython_display.writeln('Creating SqlContext as \'sqlContext\'')
```
aggFTW: Nit: you could just do double quotes so that you don't have to escape single quotes:
`self.ipython_display.writeln("Creating SqlContext as 'sqlContext'")`

remotespark/livyclientlib/livysession.py
```diff
@@ -75,7 +75,10 @@ def create_sql_context(self):
```
aggFTW: In line 68 of this file, you should also add the SparkContext as:
`self.ipython_display.writeln("Creating SparkContext as 'sc'")`

remotespark/sparkkernelbase.py
```diff
@@ -38,6 +39,7 @@ def __init__(self, implementation, implementation_version, language, language_ve
         self.logger = Log(self.client_name)
         self.session_started = False
         self._fatal_error = None
+        self._ipython_display = IpythonDisplay()
```
aggFTW: Can you please remove the `_` at the beginning of `_ipython_display`?

tests/test_livyclientfactory.py
```diff
@@ -13,8 +14,9 @@ def test_build_session_with_defaults():
     connection_string = get_connection_string("url", "user", "pass")
     kind = Constants.session_kind_pyspark
     properties = {"kind": kind}
+    ipython_display = IpythonDisplay()
```
aggFTW: Can you please make `ipython_display` a `MagicMock()` and verify that its methods are called correctly? Remove the `import` after this.

tests/test_livysession.py
```diff
@@ -56,7 +57,9 @@ def _create_session(self, kind=Constants.session_kind_spark, session_id="-1", sq
         if http_client is None:
             http_client = MagicMock()
 
-        return LivySession(http_client, session_id, sql_created, {"kind": kind})
+        ipython_display = MagicMock()
+
+        return LivySession(ipython_display, http_client, session_id, sql_created, {"kind": kind})
```
aggFTW: As Ricky pointed out, please verify that `ipython_display` methods are called correctly throughout this class.

remotespark/sparkkernelbase.py
MohamedElKamhawy: This is private variable, we should add underscore at the beginning to follow the same concept in the code, right?

tests/test_sparkkernelbase.py
```diff
@@ -284,7 +282,6 @@ def test_execute_throws_if_fatal_error_happened():
         # Assertions
         assert kernel._fatal_error == fatal_error
         assert execute_cell_mock.call_count == 0
-        assert send_error_mock.call_count == 1
```
aggFTW: This assertion is still needed. It's just that you would now verify with the ipython_display mock.
You might need to add `ipython_display=None` to SparkKernelBase constructor.

tests/test_sparkkernelbase.py
```diff
@@ -309,7 +306,6 @@ def test_execute_throws_if_fatal_error_happens_for_execution():
         # Assertions
         assert kernel._fatal_error == message
         assert execute_cell_mock.call_count == 1
-        send_error_mock.assert_called_once_with(stream_content)
```
aggFTW: This assertion is still needed. It's just that you would now verify with the ipython_display mock.
You might need to add `ipython_display=None` to SparkKernelBase constructor.

remotespark/sparkkernelbase.py
aggFTW: You are right. Can you please make logger and session_started consistent?

MohamedElKamhawy: Sure.

tests/test_livyclientfactory.py
msftristew: +1

msftristew:
I sign off

msftristew:
once tests pass that is :)

msftristew:
Lots of failures -- please have a second look @MohamedElKamhawy

aggFTW:
For me to sign off, please make unit tests pass and revert changes to `examples/remotespark.ipynb`.

msftristew:
Do we even need `examples/remotespark.ipynb`? Is it still up to date?

aggFTW:
We need it, but it's not up to date. Look at #4

msftristew:
@aggFTW look good to you?

aggFTW:
Looks good. Please revert changes to example if possible. If it takes you more than 5 minutes, just merge.

aggFTW:
Closes #51

PR #119, opened 2016-01-13
Now, Spark kernel will not throw errors if missing login configuration

msftristew:
You'll get a nice error message if an error is thrown while getting the configurations in the `__init__` method of the spark kernel base

### DIFF
--- a/tests/test_sparkkernelbase.py
+++ b/tests/test_sparkkernelbase.py
@@ -53,7 +53,7 @@ def test_get_config():
     pwd = "p"
     url = "url"
 
-    config = {"kernel_python_credentials": {user_ev: usr, pass_ev: pwd, url_ev: url}}
+    config = {conf.kernel_python_credentials.__name__: {user_ev: usr, pass_ev: pwd, url_ev: url}}
     conf.override_all(config)
 
     u, p, r = kernel._get_configuration()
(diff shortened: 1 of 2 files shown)


tests/test_sparkkernelbase.py
```diff
@@ -65,16 +65,21 @@ def test_get_config():
...
-    try:
-        kernel._get_configuration()
+    kernel._get_configuration()
+    assert kernel._fatal_error
+    conf.load()
 
-        # Above should have thrown because env var not set
-        assert False
-    except ValueError:
-        assert send_error_mock.call_count == 1
```
aggFTW: The intent of this unit test was to test that user got an error sent to him (displayed) immediately when configuration is not there. Does that work?

msftristew: That's not really possible in the way that we were doing it before.

Actually, what the way it was implemented before *should have* worked, except if you throw an error in the `__init__` method of a kernel, Jupyter assumes that the kernel has crashed.  So displaying an error to the user before they've even entered any input is not generally possible, I think.

The idea behind this change is that now the `__init__` method will mark the kernel as being in an error state when it starts up so that the error can be thrown when the user tries to run a cell.  This unit test is testing that.

aggFTW:
@MohamedElKamhawy Please take note of this PR. It will impact the refactoring you are doing for write error methods and such.

PR #120, opened 2016-01-14
Handle string aggregations correctly

aggFTW:
Widget does not crash if you try to aggregate a string column of Y by avg.


msftristew:
I assume the error message goes away when you select another option?  If so, I sign off

PR #121, opened 2016-01-14
Create SQL context after adding client to our client list

### LINKED ISSUE #110: Shutdown pyspark kernel doesn't gaurantee that session has been deleted
Repro Steps: 
- Open Pyspark kernel 
- Do any operation, 1+1 for example 
- shutdown it before getting the answer (while creating the SQL context & Hive one)
- SSH the cluster, you'll find the session still existing

For now, to delete it, you have to do that manually from ssh the cluster.

msftristew:
Closes #110 

Add a new "start" method to Livy clients. Add to our client manager's list before calling the start method.  This reduces the odds of leaking sessions.

### DIFF
--- a/remotespark/livyclientlib/sparkcontroller.py
+++ b/remotespark/livyclientlib/sparkcontroller.py
@@ -72,6 +72,7 @@ class SparkController(object):
         session.start()
         livy_client = self.client_factory.build_client(session)
         self.client_manager.add_client(name, livy_client)
+        livy_client.start()
 
     def get_client_keys(self):
         return self.client_manager.get_sessions_list()
--- a/remotespark/livyclientlib/livyclient.py
+++ b/remotespark/livyclientlib/livyclient.py
@@ -10,15 +10,14 @@ class LivyClient(object):
 
     def __init__(self, session):
         self.logger = Log("LivyClient")
-
-        execute_timeout_seconds = conf.execute_timeout_seconds()
-
         self._session = session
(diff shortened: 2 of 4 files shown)


remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -72,6 +72,7 @@ def add_session(self, name, connection_string, skip_if_exists, properties):
         session.start()
         livy_client = self.client_factory.build_client(session)
         self.client_manager.add_client(name, livy_client)
+        livy_client.start()
```
aggFTW: Should this be in the factory or here?

msftristew: I don't know how it would work if it went in the factory.  The job of a factory is to create other objects, not engage some behaviors on them.

aggFTW: I don't have strong feelings on this. It just might be useful to have the factory call the start method because then you are not calling it everywhere.

msftristew: I don't know what the interface would look like.  We would need a new method for factories, `start_client(client)`, since the factory can't start the client for us in the `build_client` method.  Are you talking about some other design?

aggFTW: Why couldn't the factory start the client for us in the `build_client` method? That's what I meant.

PR #125, opened 2016-01-16
Provide %logs command in magics and wrapper kernel

### LINKED ISSUE #123: Provide %logs for session and in wrapper kernels
This should query `/sessions/ID/log?from=0` and return output nicely formatted.

aggFTW:
Closes #123


aggFTW:
Merging so that I can start working on the python `%local` directive. I can address any comment whatsoever later.

PR #127, opened 2016-01-19
Refactor SparkKernelBase

aggFTW:
Break it up into:
* UserCommandParser
* CodeTransformers

and tie them to SparkKernelBase.

Also restructured unit tests so that they reflect new contracts between classes.

### DIFF
--- a/remotespark/utils/constants.py
+++ b/remotespark/utils/constants.py
@@ -2,7 +2,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 
-class Constants:
+class Constants(object):
     config_json = "config.json"
 
     session_kind_spark = "spark"
@@ -34,3 +34,6 @@ class Constants:
                                busy_session_status, error_session_status, dead_session_status]
     final_status = [dead_session_status, error_session_status]
 
+    delete_session_action = "delete"
+    start_session_action = "start"
+    do_nothing_action = "nothing"
--- a/remotespark/wrapperkernel/codetransformers.py
+++ b/remotespark/wrapperkernel/codetransformers.py
@@ -59,10 +59,15 @@ class SparkTransformer(UserCodeTransformerBase):
         return code_to_run, error_to_show, begin_action, end_action, deletes_session
 
 
-class SqlTransformer(UserCodeTransformerBase):
+class ContextOutputVariableTransformer(UserCodeTransformerBase):
+    def __init__(self, subcommand):
+        super(ContextOutputVariableTransformer, self).__init__(subcommand)
+
+        self.context_name = None
+
     def get_code_to_execute(self, session_started, connection_string, force, output_var, command):
         error_to_show = None
-        code_to_run = "%%spark -c sql\n{}".format(command)
+        code_to_run = "%%spark -c {}\n{}".format(self.context_name, command)
         begin_action = Constants.start_session_action
         end_action = Constants.do_nothing_action
         deletes_session = False
@@ -70,15 +75,18 @@ class SqlTransformer(UserCodeTransformerBase):
         return code_to_run, error_to_show, begin_action, end_action, deletes_session
 
 
-class HiveTransformer(UserCodeTransformerBase):
-    def get_code_to_execute(self, session_started, connection_string, force, output_var, command):
-        error_to_show = None
-        code_to_run = "%%spark -c hive\n{}".format(command)
-        begin_action = Constants.start_session_action
-        end_action = Constants.do_nothing_action
-        deletes_session = False
+class SqlTransformer(ContextOutputVariableTransformer):
+    def __init__(self, subcommand):
+        super(SqlTransformer, self).__init__(subcommand)
 
-        return code_to_run, error_to_show, begin_action, end_action, deletes_session
+        self.context_name = Constants.context_name_sql
+
+
+class HiveTransformer(ContextOutputVariableTransformer):
+    def __init__(self, subcommand):
+        super(HiveTransformer, self).__init__(subcommand)
+
+        self.context_name = Constants.context_name_hive
 
 
 class InfoTransformer(UserCodeTransformerBase):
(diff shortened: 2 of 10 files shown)


aggFTW:
Will merge as soon as this is done so that I can start working on `%local` magic for wrapper kernels.

I can address any feedback later on.

msftristew:
Thanks, you made a lot of changes that I wanted to implement myself.  I'll have a look at the changes.

remotespark/wrapperkernel/codetransformers.py
```diff
@@ -0,0 +1,159 @@
...
+
+class NotSupportedTransformer(UserCodeTransformerBase):
+    def get_code_to_execute(self, session_started, connection_string, force, output_var, command):
+        error_to_show = "Magic '{}' not supported.".format(self.name)
+        code_to_run = ""
+        begin_action = Constants.do_nothing_action
+        end_action = Constants.do_nothing_action
+        deletes_session = False
+
+        return code_to_run, error_to_show, begin_action, end_action, deletes_session
```
msftristew: I'm a little bit ambivalent about returning 5 results from this function (I'm definitely not going to remember the order, what if they get mixed up?)

Maybe returning an instance of a class or a namedtuple would be better here.

remotespark/utils/constants.py
```diff
@@ -2,7 +2,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 
-class Constants:
+class Constants(object):
     config_json = "config.json"
```
msftristew: This should have been corrected a long time ago, but this shouldn't be a class either.  The class definition should be factored out.  We can take care of that in a later CR.

remotespark/wrapperkernel/codetransformers.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def get_code_to_execute(self, session_started, connection_string, force, output_var, command):
+        raise NotImplementedError()
+
+
+class NotSupportedTransformer(UserCodeTransformerBase):
+    def get_code_to_execute(self, session_started, connection_string, force, output_var, command):
+        error_to_show = "Magic '{}' not supported.".format(self.name)
+        code_to_run = ""
+        begin_action = Constants.do_nothing_action
```
msftristew: I don't know if I'm crazy about using string-based "indirection" for actions -- it might be simpler to just use functions.  I could go either way on that.

remotespark/wrapperkernel/codetransformers.py
```diff
@@ -0,0 +1,159 @@
...
+        end_action = Constants.do_nothing_action
+        deletes_session = False
+
+        return code_to_run, error_to_show, begin_action, end_action, deletes_session
+
+
+class ConfigTransformer(UserCodeTransformerBase):
+    def get_code_to_execute(self, session_started, connection_string, force, output_var, command):
```
msftristew: As with the above comment, it might be fruitful in the future to refactor this so that we have a class which represents a "command" which has fields like "connection_string", "force", "command", etc.  This is going to be more scalable and simple than passing around a bunch of tokens.

We're OK for now but this should really be refactored in the near future because this only gets more complicated as we add more functionality.

remotespark/wrapperkernel/codetransformers.py
```diff
@@ -0,0 +1,159 @@
...
+        deletes_session = False
+
+        if not force:
+            error_to_show = "The session you are trying to delete could be this kernel's session. In order " \
+                            "to delete this session, please provide the '-f' flag at the beginning of the " \
+                            "delete magic:\n\te.g. `%delete -f id`\n\nAll previously run commands in the " \
+                            "session will be lost."
+            code_to_run = ""
+        else:
+            deletes_session = True
```
msftristew: Why is "deleting the session" not part of the `end_action`?

remotespark/wrapperkernel/sparkkernelbase.py
```diff
@@ -62,72 +55,51 @@ def do_execute(self, code, silent, store_history=True, user_expressions=None, al
...
-            return self._run_without_session(code_to_run, silent, store_history, user_expressions, allow_stdin)
-        elif subcommand == self.delete_command:
-            if self.force_flag not in flags:
-                self._show_user_error("The session you are trying to delete could be this kernel's session. In order "
-                                      "to delete this session, please provide the '-f' flag at the beginning of the "
-                                      "delete magic:\n\te.g. `%delete -f id`\n\nAll previously run commands in the "
-                                      "session will be lost.")
-                code_to_run = ""
+        # Parse command
+        subcommand, force, output_var, command = self.user_command_parser.parse_user_command(code)
```
msftristew: See comment above: this needs to be refactored into a struct.

remotespark/wrapperkernel/sparkkernelbase.py
```diff
@@ -135,6 +107,27 @@ def do_shutdown(self, restart):
 
         return self._do_shutdown_ipykernel(restart)
 
+    @staticmethod
+    def _get_code_transformer(subcommand):
```
msftristew: From a structural standpoint I don't know how much sense it makes for this function to be in this file and not in `codetransformers.py`.

remotespark/wrapperkernel/codetransformers.py
aggFTW: Agreed, let's tackle this in a later PR. I'll create a bug.

remotespark/wrapperkernel/codetransformers.py
aggFTW: I don't see this changing much, but sure.

remotespark/wrapperkernel/codetransformers.py
aggFTW: The command deletes the session.

remotespark/wrapperkernel/sparkkernelbase.py
aggFTW: We could change that.

PR #129, opened 2016-01-19
Fix multiline commands in parser

aggFTW:
Small bug I introduced.


msftristew:
Thanks for adding a new unit test for this

PR #131, opened 2016-01-19
Print instatneously using ipython_display

MohamedElKamhawy:


remotespark/livyclientlib/livysession.py
```diff
@@ -76,10 +77,10 @@ def create_sql_context(self):
 
         self.logger.debug("Starting '{}' sql and hive session.".format(self.kind))
 
-        self.ipython_display.writeln("Creating SqlContext as 'sqlContext'")
+        self.ipython_display.writeln_and_flush("Creating SqlContext as 'sqlContext'")
         self._create_context(Constants.context_name_sql)
 
-        self.ipython_display.writeln("Creating HiveContext as 'hiveContext'")
+        self.ipython_display.writeln_and_flush("Creating HiveContext as 'hiveContext'")
```
aggFTW: I'm against this rename. Leave it as writeln

remotespark/utils/ipythondisplay.py
```diff
@@ -18,5 +19,9 @@ def write(self, msg):
     def writeln(self, msg):
         self.write("{}\n".format(msg))
```
aggFTW: Flush should happen in write

remotespark/utils/ipythondisplay.py
```diff
@@ -18,5 +19,9 @@ def write(self, msg):
...
 
+    def writeln_and_flush(self, msg):
```
aggFTW: Do not introduce a new function. Just use writeln

aggFTW:
Did just test this? Is there a way we could unit test this?

aggFTW:
What about other methods, like `send_error`?

msftristew:
@aggFTW, I suggested that there be a new method for writing a message and flushing the output stream.  It is not generally necessary to flush the output stream after every write and that can have performance implications.

aggFTW:
Class `ipythondisplay` is intended to be used as a way for us to show something to the user instantaneously. It's our view. It should always show immediately to the user.

msftristew:
You can only unit test this by mocking `sys.stdout` -- possible, but nothing else in the class is mockable either.

msftristew:
Obviously this is the only circumstance in which Jupyter's output buffering has ever been an issue for us.  I don't see why we would flush the output buffer after every write, but if we're confident there will never be a situation where it would ever trouble us from a performance standpoint, then I guess it's fine.

Note that while I haven't tested it, flushing `sys.stderr` will *probably* also work for errors.

MohamedElKamhawy:
@aggFTW  does send_error have the same problem of flushing?

MohamedElKamhawy:
@msftristew  Yea, I found it here: http://stackoverflow.com/questions/10019456/usage-of-sys-stdout-flush-method

aggFTW:
I'm not worried about performance for this class. It's not the intended usage.

What did you find @MohamedElKamhawy?

MohamedElKamhawy:
@aggFTW  We can use sys.sterr.flush as well

aggFTW:
Please unit test that flush gets called on every `write` and `writeln` or some useful unit test. We don't want regressions. This also applies to the error scenarios. You need to find out if we have a problem there @MohamedElKamhawy.

msftristew:
Mohamed, are you going to add a unit test for this?

MohamedElKamhawy:
@msftristew  Yes, I'm trying that but have some errors.

tests/test_ipythondisplay.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_stdout_flush():
+    ipython_shell = MagicMock()
+    ipython_display = IpythonDisplay()
+    ipython_display._ipython_shell = ipython_shell
+    sys.stdout = MagicMock()
+
+    ipython_display.write('Testing Stdout Flush')
+    assert sys.stdout.flush.call_count == 1
+
```
aggFTW: Nit: double new line between methods per PEP specification.

remotespark/livyclientlib/livysession.py
```diff
@@ -4,6 +4,7 @@
 import textwrap
 from time import sleep, time
 import remotespark.utils.configuration as conf
+
```
aggFTW: Nit: Remove line.

msftristew:
LGTM

PR #134, opened 2016-01-21
Include asynchronous module definition for Javascript files

msftristew:
Now include an asynchronous module definition (AMD) for kernel.js files.  This should remove instances of undefined behavior on the client side.


msftristew:
I can confirm this corrected the warning in the console.

PR #136, opened 2016-01-21
Default Livy URL is localhost:8998

### LINKED ISSUE #108: Should the default Livy URL be localhost:8998?
See title. I wonder if this change would increase the odds that someone can do a clean install of the wrapper kernels and have everything "just work" without having to mess with configurations at all.

aggFTW:
Closes #108


msftristew:
gr8

PR #137, opened 2016-01-21
Coerce df data types that we are getting back from Livy via SQL queries

### LINKED ISSUE #55: Make pandas dtypes correct
This includes running [nunique](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.nunique.html) on the string columns and then making them categorical values instead of just strings.

aggFTW:
This is now done on the livy client and not on datawidgets

Closes #55 
Not coercing into `categorical` because of bug in unit testing framework, outlined in https://github.com/pydata/pandas/issues/12108

### DIFF
--- a/remotespark/utils/utils.py
+++ b/remotespark/utils/utils.py
@@ -117,11 +117,3 @@ def coerce_pandas_df_to_numeric_datetime(df):
                 coerced = True
             except:
                 pass
-
-        # TODO: To be uncommented when https://github.com/pydata/pandas/issues/12108 is fixed
-        # if not coerced and df[column_name].nunique() < SOME_NUMBER:
-        #     try:
-        #         df[column_name] = df[column_name].astype('category')
-        #         coerced = True
-        #     except:
-        #         pass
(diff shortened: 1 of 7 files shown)


msftristew:
Your issue has been closed as not a bug, are you going to make any further changes to this PR in light of that?

aggFTW:
I do think it's a bug, but I'll probably make some more changes based on his other comments.

aggFTW:
I won't be making any more changes for now.

msftristew:
OK will review in a bit

remotespark/utils/utils.py
```diff
@@ -97,3 +99,21 @@ def get_instance_id():
...
+
+def coerce_pandas_df_to_numeric_datetime(df):
+    for column_name in df.columns:
+        coerced = False
+
+        if not coerced and df[column_name].dtype == np.dtype("object"):
+            try:
+                df[column_name] = pd.to_datetime(df[column_name], errors="raise")
+                coerced = True
+            except:
```
msftristew: Catching all exceptions is a bad idea.  Identify the actual exception that Pandas throws when a coercion error happens and catch only those.  If the exception Pandas throws is unpredictable or unknown, then at least `except Exception:` -- but that should be a last-resort thing

aggFTW: Agreed :) Catching `ValueError` now

PR #140, opened 2016-01-21
Graceful error handling for wrapper kernels

### LINKED ISSUE #138: Unexpected errors are not displayed to the user
We should catch these errors and display them.  This involves catching any exceptions our code might throw unexpectedly.

msftristew:
Improve error handling for wrapper kernels, especially for unexpected errors.  Also ensure that cells complete their execution from Jupyter's UI if an error is encountered.

Closes #138

### DIFF
--- a/remotespark/wrapperkernel/sparkkernelbase.py
+++ b/remotespark/wrapperkernel/sparkkernelbase.py
@@ -58,7 +58,7 @@ class SparkKernelBase(IPythonKernel):
         try:
             return self._do_execute(code, silent, store_history, user_expressions, allow_stdin)
         except Exception as e:
-            self._throw_internal_error(e)
+            self._show_internal_error(e)
             return self._complete_cell()
 
     def do_shutdown(self, restart):
@@ -217,7 +217,7 @@ ip.display_formatter.ipython_display_formatter.for_type_by_name('pandas.core.fra
         self._logger.error(message)
         self._ipython_display.send_error(message)
 
-    def _throw_internal_error(self, e):
+    def _show_internal_error(self, e):
         self._logger.error("ENCOUNTERED AN INTERNAL ERROR: {}".format(e))
         self._ipython_display.send_error("An internal error was encountered. "
                                          "Please file an issue at https://github.com/jupyter-incubator/sparkmagic")
@@ -231,7 +231,7 @@ ip.display_formatter.ipython_display_formatter.for_type_by_name('pandas.core.fra
     def _abort_with_fatal_error(self, message):
         """Queues up a fatal error and throws it immediately."""
         self._queue_fatal_error(message)
-        self._repeat_fatal_error()
+        return self._repeat_fatal_error()
 
     def _repeat_fatal_error(self):
         """Throws an error that has already been queued."""
--- a/remotespark/wrapperkernel/sparkkernelbase.py
+++ b/remotespark/wrapperkernel/sparkkernelbase.py
@@ -211,7 +211,7 @@ ip.display_formatter.ipython_display_formatter.for_type_by_name('pandas.core.fra
         """A method that runs a cell with no effect. Call this and return the value it
         returns when there's some sort of error preventing the user's cell from executing; this
         will register the cell from the Jupyter UI as being completed."""
-        return self._execute_cell("None", False, False, None, False)
+        return self._execute_cell("None", False, True, None, False)
 
     def _show_user_error(self, message):
         self._logger.error(message)
--- a/remotespark/wrapperkernel/sparkkernelbase.py
+++ b/remotespark/wrapperkernel/sparkkernelbase.py
@@ -54,59 +54,64 @@ class SparkKernelBase(IPythonKernel):
 
     def do_execute(self, code, silent, store_history=True, user_expressions=None, allow_stdin=False):
         if self._fatal_error is not None:
-            self._repeat_fatal_error()
-
-        # Parse command
-        subcommand, force, output_var, command = self.user_command_parser.parse_user_command(code)
-
-        # Get transformer
-        transformer = self._get_code_transformer(subcommand)
-
-        # Get instructions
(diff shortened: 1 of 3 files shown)


remotespark/wrapperkernel/sparkkernelbase.py
```diff
@@ -224,4 +238,4 @@ def _repeat_fatal_error(self):
         error = conf.fatal_error_suggestion().format(self._fatal_error)
         self._logger.error(error)
         self._ipython_display.send_error(error)
-        raise ValueError(self._fatal_error)
+        return self._complete_cell()
```
aggFTW: Shouldn't line 234 be `return self._repeat_fatal_error()`?

remotespark/wrapperkernel/sparkkernelbase.py
```diff
@@ -204,10 +207,21 @@ def _execute_cell_for_user(self, code, silent, store_history=True, user_expressi
...
+        return self._execute_cell("", silent, store_history, user_expressions, allow_stdin)
+
     def _show_user_error(self, message):
         self._logger.error(message)
         self._ipython_display.send_error(message)
 
+    def _throw_internal_error(self, e):
+        self._logger.error("ENCOUNTERED AN INTERNAL ERROR: {}".format(e))
+        self._ipython_display.send_error("An internal error was encountered. "
+                                         "Please file an issue at https://github.com/jupyter-incubator/sparkmagic.")
```
aggFTW: Idea: What about displaying this as HTML so that the link works?

msftristew: The link works already :)

remotespark/wrapperkernel/sparkkernelbase.py
msftristew: ? It is.

aggFTW: `_abort_with_fatal_error` returns `None` now because it has no `return` statement.

remotespark/wrapperkernel/sparkkernelbase.py
msftristew: Wait, need to make a small change so that this works

remotespark/wrapperkernel/sparkkernelbase.py
msftristew: Right, I see.

remotespark/wrapperkernel/sparkkernelbase.py
```diff
@@ -54,59 +54,64 @@ def __init__(self, implementation, implementation_version, language, language_ve
 
     def do_execute(self, code, silent, store_history=True, user_expressions=None, allow_stdin=False):
         if self._fatal_error is not None:
-            self._repeat_fatal_error()
+            return self._repeat_fatal_error()
+        try:
+            return self._do_execute(code, silent, store_history, user_expressions, allow_stdin)
+        except Exception as e:
```
aggFTW: As it is, `SyntaxError` from user will result in the suggestion to file an issue on this repo. We should treat SyntaxError differently.

remotespark/wrapperkernel/sparkkernelbase.py
aggFTW: Why were unit tests not catching this scenario?

msftristew: To my knowledge, there are no unit tests that check the return value of `do_execute` -- all of them just that calling the function has the right behavior.

aggFTW:
Can we just test the following scenarios for the right behavior?

* Something that results in fatal error
* Valid statement after fatal error
* Something that throws (raise error on purpose)
* No Livy endpoint for kernel
* SyntaxError from user
* Spark error coming back

Any other scenario?

remotespark/wrapperkernel/sparkkernelbase.py
aggFTW: Should we add unit tests?

remotespark/wrapperkernel/sparkkernelbase.py
msftristew: I am unsure if that's true.  Can you give me an example of something that might trigger the error.

aggFTW: Execute:

`%delete -f 9`

remotespark/wrapperkernel/sparkkernelbase.py
msftristew: It's a good idea but I don't personally know what this method returns, since we're offshoring all the work in this method to the IPython kernel.  Might require some investigation to see exactly what we should expect in order to add unit tests.

aggFTW: No need to do that. We can just mock what the method that calls IPython returns.

_____________________________

remotespark/wrapperkernel/sparkkernelbase.py
msftristew: This issue is caused by a problem in _do_execute.  I've corrected it

msftristew: BTW, `%delete -f 9` is not a syntax error

msftristew:
I don't know how to trigger a SyntaxError (`%delete -f 9` doesn't do it).  What do "No Livy endpoint for kernel" and "Spark error coming back" mean?

screen shot 2016-01-21 at 2 47 44 pm

screen shot 2016-01-21 at 2 52 18 pm

Here are cases 1-3 from that list.

There's a problem in that the In number doesn't get incremented correctly.

aggFTW:
I'm concerned about the number not increasing. What is special about that case?

msftristew:
Wait, now the issue is still there?  Hold on.

aggFTW:
Could it be the history flag?

msftristew:
Yep. IPython is very finicky :)

PR #142, opened 2016-01-22
Improve -f flag parsing for delete and cleanup magics

aggFTW:
Now all of these are valid parsing statements:

%delete -f 9
%delete 9
%delete 9 -f
%cleanup -f
%cleanup


msftristew:
Exciting!

PR #152, opened 2016-01-27
Moving some files around

aggFTW:


msftristew:
Are any further changes necessary to be made to e.g. the kernelspecs for things to work out of the box?  Don't want any more surprises for installation.

aggFTW:
Yes, there are more changes, but I don't want to fix them right now. I'm going to break them again very shortly. From this PR to the time that I'm done, we should freeze publicpreview0.1 and install from there.

Thoughts?

msftristew:
I am OK with freezing publicpreview0.1 from now on but nothing should be committed to master unless it builds and runs correctly (i.e. I should be able to go back to any point in time, check out master, run the necessary `pip install` and `jupyter kernelspec install` commands, and run the magics).  If this is going to break build then I'd say doing the work on a feature branch would be more appropriate.  Do you disagree?

aggFTW:
No, I agree. Let me create a feature branch and resubmit PR.

PR #154, opened 2016-01-28
Rewrite kernels

aggFTW:
Sorry for the massive PR... Github was down and I didn't stop writing code :smile: 

You could walk through Commits, pull me in to explain code in person, look at codecoverage (it's pretty good), test against a live cluster by installing them (I did test like that), or just look at everything together.

Still TODO:
* Turn kernel magics into line or cell magics, but not both.
* Implement a %help
* %info needs to also print session configs and identify the session number that kernel is currently running under


msftristew:
Wow.

remotespark/kernels/wrapperkernel/usercodeparser.py
```diff
@@ -0,0 +1,120 @@
...
+            first_line_next = first_line[l+1:]
+        else:
+            separator = ""
+            first_line_next = first_line[l:]
+
+        return "{}{}{}".format(first_line_next, separator, rest)
+
+    @staticmethod
+    def _is_comment(language, line, in_multi_line, start_used):
+        """Returns (bool, bool, str) where:
```
msftristew: The API surrounding parsing comments here is pretty messy, and (though I haven't tested it) I'm fairly certain that it doesn't work in 100% of cases (example: in order to implement a proper multiline comment parser for Python, what you actually need is a string parser, and this code recognizes no escape sequences).  

In order to make everyone's lives easier what you instead want is the function:

    def _chomp_comment(language, code):
         """Given a block of code in the source language, return all the code after an initial comment if there is one."""
        ...

Then the implementation of `_chomp_comment` can simply iterate through all the characters in the code block and return a new string after all the comments/whitespace at the beginning.  This is more flexible and allows you to actually capture all the corner cases.

remotespark/kernels/wrapperkernel/sparkkernelbase.py
```diff
@@ -68,80 +59,24 @@ def do_shutdown(self, restart):
...
-            pass
-        else:
-            raise ValueError("End action {} not supported.".format(end_action))
-
-        if deletes_session:
-            self._session_started = False
-
         return res
 
-    @staticmethod
```
msftristew: It's great how much code we're able to get rid of here.

remotespark/livyclientlib/clientmanager.py
```diff
@@ -44,7 +44,7 @@ def get_sessions_list(self):
         return list(self._livy_clients.keys())
 
     def get_sessions_info(self):
-        return ["Name: {}\t{}".format(k, str(self._livy_clients[k])) for k in self._livy_clients.keys()]
+        return ["Name: {}\t{}".format(k, str(self._livy_clients[k])) for k in list(self._livy_clients.keys())]
```
msftristew: Why this change?

remotespark/kernels/wrapperkernel/usercodeparser.py
aggFTW: Can you please suggest unit test cases that might break it?

msftristew:
I won't be able to wrap my mind around this entire code review at once but if we could get together and talk through the major changes, that would be good.

remotespark/livyclientlib/clientmanager.py
aggFTW: Python 3 dict.keys() returns a keys object and not a list.

aggFTW:
I understand :) 
I'll stop by.

remotespark/kernels/wrapperkernel/usercodeparser.py
msftristew: `"""abc\"""`

I think your code would recognize this as a full comment that begins and ends on the same line, but actually it's not (you should continue lexing in the next line).

remotespark/livyclientlib/clientmanager.py
msftristew: Does that have an effect on correctness?

msftristew:
Everything I see here looks more or less OK at first glance though (this seems more like a big reorganization, but which looks like it makes everything much more comprehensible).

MohamedElKamhawy:
@msftristew  +1
I'm doing my best to recognize what's the difference between the old and the new changes.

@aggFTW  We can view this in a meeting room to get an idea of the changes.

remotespark/kernels/wrapperkernel/usercodeparser.py
msftristew: Are we making it a policy that code cannot come directly after a multiline comment?  So this is never expected to work?

    /* yo */ %%cleanup

msftristew: On the same token, what if a user puts in this?

/* yo */ %%cleanup /* yo */

Your code would interpret this as a single multi-line comment, which is both unexpected and incorrect.

aggFTW: I think you are right. There's too many edge cases I did not consider. I believe we need a parser for this. I'll do some reading and come back to the thread.

aggFTW:
Met with you. We agreed to re-do our approach to parsing. That will come in a later PR.
Thanks!

PR #160, opened 2016-02-04
Manage widget for IPython kernel - first effort

aggFTW:
Play around with it. Screenshots below.


remotespark/controllerwidget/abstractmenuwidget.py
```diff
@@ -0,0 +1,40 @@
...
+
+        if ipywidget_factory is None:
+            ipywidget_factory = IpyWidgetFactory()
+        self.ipywidget_factory = ipywidget_factory
+
+        if ipython_display is None:
+            ipython_display = IpythonDisplay()
+        self.ipython_display = ipython_display
+
+        self.children = []
```
msftristew: This is an attribute that needs to be overridden by a subclass, right?  I'm wondering if you should not set this here so it will be an error if it's accessed and not implemented by the subclass.

msftristew:
Design looks very sensible to me. Just get those tests passing :)

remotespark/controllerwidget/abstractmenuwidget.py
aggFTW: Which attribute?

aggFTW:
First view:


Add endpoint:


Manage Endpoint:


Create Session:


Manage Session:

msftristew:
Not going to public preview branch right?

aggFTW:
No, this would go to master only. From here on, if we need to bug fix anything in public preview, we'd have to apply changes both in master and publicpreview

aggFTW:
@msftristew is this ready for a merge?

msftristew:
Go for it.

PR #161, opened 2016-02-05
Configure prints info

aggFTW:


aggFTW:
Nitin was puzzled that after running `%%configure` he didn't know if the changes had gone in. So now, %%configure will always print info for the session afterwards.

MohamedElKamhawy:
Sign-Off :)

msftristew:
Good idea

PR #162, opened 2016-02-08
Start session timeouts send fatal error and stop further executions

aggFTW:


msftristew:
Doesn't have to happen in this PR, but we should really get the ball rolling on removing this "private magics" (the ones that begin with `_do_not_call`)

remotespark/kernels/kernelmagics.py
```diff
@@ -30,6 +30,8 @@ def __init__(self, shell, data=None):
         self.url = None
         self.connection_string = None
```
MohamedElKamhawy: Remove extra line.

remotespark/kernels/kernelmagics.py
```diff
@@ -190,7 +195,7 @@ def cleanup(self, line, cell="", local_ns=None):
             self.ipython_display.send_error("When you clean up the endpoint, all sessions will be lost, including the "
```
MohamedElKamhawy: This is not related to this PR. It's better to set a string variable with this sentence then print it.

remotespark/kernels/kernelmagics.py
```diff
@@ -206,23 +211,38 @@ def delete(self, line, cell="", local_ns=None):
                 self.ipython_display.send_error("Cannot delete this kernel's session ({}). Specify a different session,"
```
MohamedElKamhawy: The same applied here.

remotespark/kernels/kernelmagics.py
```diff
@@ -206,23 +211,38 @@ def delete(self, line, cell="", local_ns=None):
...
+
+        if not self.session_started:
             skip = False
             properties = conf.get_session_properties(self.language)
 
-            self.spark_controller.add_session(self.session_name, self.connection_string, skip, properties)
+            try:
+                self.session_started = True
+                self.spark_controller.add_session(self.session_name, self.connection_string, skip, properties)
+            except Exception as e:
```
MohamedElKamhawy: Leave an extra line before this line for readability.

tests/test_kernel_magics.py
```diff
@@ -5,6 +5,7 @@
 import remotespark.utils.configuration as conf
 from remotespark.kernels.kernelmagics import KernelMagics
 from remotespark.utils.constants import Constants
+from remotespark.livyclientlib.livyclienttimeouterror import LivyClientTimeoutError
```
MohamedElKamhawy: Re-arrange alphabetically. They should be:
 from remotespark.kernels.kernelmagics import KernelMagics
 from remotespark.livyclientlib.livyclienttimeouterror import LivyClientTimeoutError
 from remotespark.utils.constants import Constants

msftristew:
##APPROVED

tests/test_kernel_magics.py
```diff
@@ -53,19 +54,42 @@ def test_start_session():
     line = ""
     assert not magic.session_started
 
-    magic._do_not_call_start_session(line)
+    ret = magic._do_not_call_start_session(line)
```
MohamedElKamhawy: Change the variable name to be more meaningful.

remotespark/kernels/kernelmagics.py
aggFTW: ?

remotespark/kernels/kernelmagics.py
aggFTW: That's not PEP

tests/test_kernel_magics.py
aggFTW: Meaningful enough. It's a unit test.

MohamedElKamhawy: What "ret" stands for?

aggFTW: It's short for whatever is returned. Another very popular option is `ret_v`. Usually used in places where it's not really that meaningful, like a unit test.

remotespark/kernels/kernelmagics.py
MohamedElKamhawy: What's "PEP"? Generally, this will help the code reader to identify whether this is a different statement separated from the try or not.

remotespark/kernels/kernelmagics.py
MohamedElKamhawy: To use a string and doing whatever you want to do with it, it's better to save the string first in a new variable then use it.

remotespark/kernels/kernelmagics.py
aggFTW: Did you search for it? That's what indentation is for.

remotespark/kernels/kernelmagics.py
aggFTW: Maybe. I wouldn't say that as a blank statement.

tests/test_kernel_magics.py
MohamedElKamhawy: Usually I prefer to have a meaningful variable name even if it's in unit test, this will help any new person to this code section to get the intent of the test easily.

remotespark/kernels/kernelmagics.py
MohamedElKamhawy: Ok, that works.

PR #168, opened 2016-02-19
Remove Hive queries from magics; replace SqlContext with HiveContext everywhere

### LINKED ISSUE #166: Create hiveContext only -- sqlContext is not needed
"In addition to the basic SQLContext, you can also create a HiveContext, which provides a superset of the functionality provided by the basic SQLContext. Additional features include the ability to write queries using the more complete HiveQL parser, access to Hive UDFs, and the ability to read data from Hive tables. To use a HiveContext, you do not need to have an existing Hive setup, and all of the data sources available to a SQLContext are still available. HiveContext is only packaged separately to avoid including all of Hive’s dependencies in the default Spark build. If these dependencies are not a problem for your application then using HiveContext is recommended for the 1.3 release of Spark. "

msftristew:
1. Remove `%%hive`, Hive execution context, all references to Hive in documentatoin

2. Rename `hiveContext` to `sqlContext`. 

Closes #166


msftristew:
Idea: not all Spark users use Hive so it might be neat if we had a configuration option `"hive-enabled"` which you could set.  If you choose not to use Hive, then it defaults to a normal `SqlContext`.

msftristew:
Need to merge from master

msftristew:
Done

aggFTW:
This looks fine. The idea to not use hive is neat and worth thinking about. Let's open an issue to track it.

PR #170, opened 2016-02-23
Add -q feature to silence SQL queries selectively

### LINKED ISSUE #141: Silence SQL queries
Now that we support `-o` for collecting results as a dataframe, it would be nice to turn off automatic visualizations on a per-query level so that e.g. if I want to collect the results of a SQL query and visualize it in `%%local` mode, I can turn off visualizations for that query to cut down on visual noise.

msftristew:
We do this by returning `None` from the magic (instead of the dataframe) whenever the `-q` argument is passed.

Closes #141

### DIFF
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -161,7 +161,7 @@ class KernelMagics(SparkMagicBase):
     @needs_local_scope
     @argument("-o", "--output", type=str, default=None, help="If present, query will be stored in variable of this "
                                                              "name.")
-    @argument("-q", "--quiet", type=bool, default=False, const=True, nargs="?", help="Do not display visualization")
+    @argument("-q", "--quiet", type=bool, default=False, const=True, nargs="?", help="Return None instead of the dataframe.")
     def sql(self, line, cell="", local_ns=None):
         if self._do_not_call_start_session(""):
             args = parse_argstring(self.sql, line)
@@ -175,7 +175,7 @@ class KernelMagics(SparkMagicBase):
     @needs_local_scope
     @argument("-o", "--output", type=str, default=None, help="If present, query will be stored in variable of this "
                                                              "name.")
-    @argument("-q", "--quiet", type=bool, default=False, const=True, nargs="?", help="Do not display visualization")
+    @argument("-q", "--quiet", type=bool, default=False, const=True, nargs="?", help="Return None instead of the dataframe.")
     def hive(self, line, cell="", local_ns=None):
         if self._do_not_call_start_session(""):
             args = parse_argstring(self.hive, line)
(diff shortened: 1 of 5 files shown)


MohamedElKamhawy:
@msftristew  What do you mean by "silence SQL queries selectively"?

msftristew:
This turns off autoviz when you use the `-q` option with a SQL query.

MohamedElKamhawy:
Signed-Off

remotespark/kernels/kernelmagics.py
```diff
@@ -174,11 +175,12 @@ def sql(self, line, cell="", local_ns=None):
     @needs_local_scope
     @argument("-o", "--output", type=str, default=None, help="If present, query will be stored in variable of this "
                                                              "name.")
+    @argument("-q", "--quiet", type=bool, default=False, const=True, nargs="?", help="Do not display visualization")
```
aggFTW: This doesn't display visualizations, it only returns a pandas df.

remotespark/magics/remotesparkmagics.py
```diff
@@ -43,6 +43,7 @@ def manage_spark(self, line, cell="", local_ns=None):
                                       "If only one session has been created, there's no need to specify one.")
     @argument("-o", "--output", type=str, default=None, help="If present, output when using SQL or Hive "
                                                              "query will be stored in variable of this name.")
+    @argument("-q", "--quiet", type=bool, default=False, nargs="?", const=True, help="Do not display visualizations")
```
aggFTW: Ditto

remotespark/magics/sparkmagicsbase.py
```diff
@@ -43,12 +43,15 @@ def __init__(self, shell, data=None):
...
         self.logger.debug("Initialized spark magics.")
 
-    def execute_against_context_that_returns_df(self, method, cell, session, output_var):
+    def execute_against_context_that_returns_df(self, method, cell, session, output_var, quiet):
         try:
             df = method(cell, session)
             if output_var is not None:
                 self.shell.user_ns[output_var] = df
-            return df
+            if quiet:
```
aggFTW: Should we throw an assertion error when quiet is True and output_var is None? That option doesn't make any sense to me.

msftristew: No. You can run SQL queries which have a side effect and which return no value (e.g. CREATE TABLE)

aggFTW: Very true!

remotespark/kernels/kernelmagics.py
msftristew: Good point. Should I edit the "help" to be something like "return None instead of a pandas df"?

aggFTW:
Rubberstamp

remotespark/kernels/kernelmagics.py
aggFTW: Yes please

PR #174, opened 2016-02-26
Added capability to configure sampling from SQL queries + refactoring

### LINKED ISSUE #59: Allow user to specify how many rows/what method to use when doing a sql query
A user should be able to specify how many rows to get back in a case by case basis.

Also, the user might want to sometimes do a take, sometimes do a sample, and sometimes do something else when getting results back in a case by case basis.

msftristew:
Closes #59 

* Refactored LivyClient, PysparkLivyClient, ScalaLivyClient, etc

* Add SQLQuery class (a POD class which represents a SQL query from the user, with sampling options)

* Add feature to allow you to select how you sample SQL results from the cluster with `-m`, `-n`, `-r` options


remotespark/kernels/kernelmagics.py
```diff
@@ -70,10 +73,18 @@ def help(self, line, cell="", local_ns=None):
...
-    <td>Executes a SQL query against the sqlContext. If the -o parameter is passed, the result of the query will be
-    available in the %%local Python context as a <a href="http://pandas.pydata.org/">Pandas</a> dataframe. If
-    the -q parameter is passed, the magic will return None instead of the dataframe.</td>
+    <td>%%sql -o tables -q<br/>SHOW TABLES</td>
+    <td>Executes a SQL query against the sqlContext.
+    Parameters:
+      <ul>
+        <li>-o VAR_NAME: The result of the query will be available in the %%local Python context as a
+          <a href="http://pandas.pydata.org/">Pandas</a> dataframe.</li>
+        <li>-q: The magic will return None instead of the dataframe (no visualization).</li>
```
aggFTW: Can we expect users to understand that every cell has a return value? Can we explain that here?

remotespark/livyclientlib/sqlquery.py
```diff
@@ -0,0 +1,40 @@
...
+                 samplefraction=None, only_columns=False):
+        if samplemethod is None:
+            samplemethod = conf.default_samplemethod()
+        if maxrows is None:
+            maxrows = conf.default_maxrows()
+        if samplefraction is None:
+            samplefraction = conf.default_samplefraction()
+
+        assert samplemethod == 'take' or samplemethod == 'sample'
+        assert isinstance(maxrows, int)
```
aggFTW: Can maxrows be negative?

remotespark/livyclientlib/sqlquery.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Used only for unit testing
+    def __eq__(self, other):
+        return self.query == other.query and \
+            self.samplemethod == other.samplemethod and \
+            self.maxrows == other.maxrows and \
+            self.samplefraction == other.samplefraction and \
+            self.only_columns == other.only_columns
+
+    def __ne__(self, other):
```
aggFTW: I would've thought this is the default implementation? No?

remotespark/kernels/kernelmagics.py
MohamedElKamhawy: +1

aggFTW:
As mentioned f2f, I'll wait to do the review until you send the new iteration to move all the language handling to the SQLquery class, at which point we don't even need different clients.

remotespark/livyclientlib/sqlquery.py
msftristew: Yes, if maxrows is negative that means that the number of rows pulled from the server is not limited.

remotespark/livyclientlib/sqlquery.py
msftristew: Nope (Python is surprising sometimes).

remotespark/kernels/kernelmagics.py
msftristew: This is a Jupyter kernel so I am assuming a baseline familiarity with the idea that Jupyter cells have a return value.  I don't think it's our responsibility to educate on Jupyter features from the `%%help` popup.

msftristew: (The (no visualizations) part of the explanation is supposed to be a clue about what the end-effect of the option is for the user)

remotespark/livyclientlib/sqlquery.py
aggFTW: All sentient values should be in the documentation and %%help

msftristew: Think you mean sentinel :)

msftristew: I better not have written any code with sentient values

msftristew:
Refactor completed

remotespark/livyclientlib/sqlquery.py
aggFTW: the singularity is coming

tests/test_livyclient.py
```diff
@@ -122,10 +131,58 @@ def test_get_logs_returns_session_logs():
...
 
     res, logs_r = client.get_logs()
 
     assert not res
     assert logs_r == err
+
+
+@with_setup(_setup, _teardown)
+def test_execute_sql():
+    sqlquery = SQLQuery("HERE IS THE QUERY", "take", 100, 0.2)
```
aggFTW: Should we mock SQLQuery on these class's unit tests? I'm worried the implementation of the `to_command` method could impact the unit tests here. Should we decouple?

msftristew: My thought on this was that the implementation of the `to_command` method is (1) very simple and (2) extremely unlikely to change.  As long as `to_command` is a pure function (a reasonable assumption), these unit tests will not fail.

aggFTW: Fair enough.

aggFTW:
Rubberstamp

PR #175, opened 2016-02-29
Adding a skeleton for storing Spark Magic events to MDS

MohamedElKamhawy:


MohamedElKamhawy:
It's my bad, I just committed without adding the pull request.

msftristew:
Talked f2f

remotespark/utils/configuration.py
```diff
@@ -138,6 +138,13 @@ def logging_config():
         }
     }
 
+@_override
+def events_config():
+    return {
+        "local": 1,
```
msftristew: Unnecessary

remotespark/utils/jupytereventsfilehandler.py
```diff
@@ -0,0 +1,22 @@
...
+from remotespark.utils.filesystemreaderwriter import FileSystemReaderWriter
+
+
+class EventsFileHandler(logging.FileHandler):
+    """The default logging handler used by the magics; this behavior can be overridden by modifying the config file"""
+    def __init__(self):
+        # Simply invokes the behavior of the superclass, but sets the filename keyword argument if it's not already set.
+
+        magics_home_path = get_magics_home_path()
+        events_folder_name = "events"
```
msftristew: No need to create a new file, just use the logger

remotespark/utils/jupytereventsfilehandler.py
```diff
@@ -0,0 +1,22 @@
...
+        events_file_name = "events{}.log".format(get_instance_id())
+        directory = FileSystemReaderWriter(join_paths(magics_home_path, events_folder_name))
+        directory.ensure_path_exists()
+        events_file = join_paths(directory.path, events_file_name)
+
+    def emit(self, **kwargs):
```
msftristew: API needs to be more granular

remotespark/utils/jupyterevents.py
```diff
@@ -0,0 +1,28 @@
...
+            # Create the local file for the events
+        else:
+            # Load the hdinsighteventshandler class
+            self.local = 0
+            self._import(conf.events_config()["handler"])
+
+    def _import(self, name):
+        components = name.split('.')
+        mod = __import__(components[0])
+        for comp in components[1:]:
```
msftristew: ?

remotespark/utils/jupyterevents.py
```diff
@@ -0,0 +1,28 @@
+# Read the config file that will return the filehandler
+# file handler will be one of two local file handler or hdinsighteventshandler
+
+import remotespark.utils.configuration as conf
+
+
+class JupyterEvents:
```
msftristew: This name isn't really accurate

remotespark/utils/jupytereventsfilehandler.py
```diff
@@ -0,0 +1,22 @@
+import logging
+
+from remotespark.utils.utils import join_paths, get_magics_home_path, get_instance_id
+from remotespark.utils.filesystemreaderwriter import FileSystemReaderWriter
+
+
+class EventsFileHandler(logging.FileHandler):
```
msftristew: Shouldn't use logging infrastructure

msftristew:
In general I would recommend against submitting incomplete PR's that are not ready to be merged.

remotespark/utils/configuration.py
aggFTW: What does `local` mean @MohamedElKamhawy?

remotespark/utils/jupyterevents.py
```diff
@@ -0,0 +1,28 @@
+# Read the config file that will return the filehandler
+# file handler will be one of two local file handler or hdinsighteventshandler
```
aggFTW: No mention of hdinsight in this repo please. Just say that the file handler will be one of N implementations, default one being X.

remotespark/utils/jupyterevents.py
```diff
@@ -0,0 +1,28 @@
...
+    def _import(self, name):
+        components = name.split('.')
+        mod = __import__(components[0])
+        for comp in components[1:]:
+            mod = getattr(mod, comp)
+        return mod
+
+    def fire_event(self, **kwargs):
+        # Store the events
+        # Based on the config file which wil control which
```
aggFTW: Where's the rest of the file?

remotespark/utils/jupyterevents.py
aggFTW: I also don't get what's going on here.

remotespark/utils/jupyterevents.py
```diff
@@ -0,0 +1,28 @@
...
+
+import remotespark.utils.configuration as conf
+
+
+class JupyterEvents:
+    def __init__(self):
+
+        if conf.events_config()["local"] == 1:
+            # Create a local events storage
+            self.local = 1
```
aggFTW: How will this be used?

remotespark/utils/jupyterevents.py
aggFTW: What is this class's responsibility?

remotespark/utils/jupytereventsfilehandler.py
aggFTW: +1

MohamedElKamhawy:
@aggFTW , @msftristew : I added the main skeleton (not the whole code yet), please have a look at it. Also, I'm still playing with livysession, this is not the final version.

Give an attention for the sparkevents & eventshandler files, please.

remotespark/livyclientlib/livysessionstate.py
```diff
@@ -39,5 +44,5 @@ def connection_string(self):
         return self._connection_string
 
     def to_dict(self):
-        return {"id": self.session_id, "kind": self.kind, "sqlcontext": self.sql_context_created,
+        return {"GUID": self.session_guid, "id": self.session_id, "kind": self.kind, "sqlcontext": self.sql_context_created,
```
msftristew: why is GUID in all caps when the others aren't?

remotespark/utils/configuration.py
```diff
@@ -144,6 +144,12 @@ def logging_config():
         }
     }
 
+@_override
+def events_config():
+    return {
+        "handler": "remotespark.utils.jupytereventfilehandler"
```
msftristew: * Don't put the string `Jupyter` in any file names (remember, these events have nothing to do with Jupyter)

* Remember that the default handler is not a "file" handler

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,32 @@
+import remotespark.utils.configuration as conf
+
+
+class JupyterEvents:
```
msftristew: Change the name from JupyterEvents

MohamedElKamhawy: There is a missing change!
I'll pull a new request.

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self):
+        self.logger = Log("EventsHandler")
+
+    def store_event(self, **kwargs):
+        """
+        Storing the Event details using the logger.
+        """
+        event_line = ""
+        for key, value in kwargs.iteritems():
+            event_line += "{} = {}\n".format(key, value)
```
msftristew: The key-value pairs should all be on the same line.

MohamedElKamhawy:
I have a problem with Git so reverted everything and wrote it back.
Will make sure that everything is right then will have a new PR.

msftristew:
Mohamed, also be sure to run unit tests before pushing.

remotespark/livyclientlib/livysession.py
```diff
@@ -50,8 +51,10 @@ def __init__(self, http_client, properties, ipython_display,
         self._status_sleep_seconds = status_sleep_seconds
         self._statement_sleep_seconds = statement_sleep_seconds
         self._create_sql_context_timeout_seconds = create_sql_context_timeout_seconds
+        session_guid = uuid.uuid4()
 
-        self._state = LivySessionState(session_id, self._http_client.connection_string,
+
+        self._state = LivySessionState(session_guid, session_id, self._http_client.connection_string,
```
msftristew: Is it important that the GUID be in the session state?

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,15 @@
+from remotespark.utils.log import Log
+
+
+class EventsHandler:
+    def __init__(self):
+        self.logger = Log("EventsHandler")
+
+    def store_event(self, **kwargs):
```
msftristew: Why `**kwargs` rather than `**args`?

Mohamed, remember that you have to find a way to order the dimensions (so keyword arguments don't work)

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,15 @@
...
+        """
+        Storing the Event details using the logger.
+        """
+        event_line = ""
+        for key, value in kwargs.iteritems():
+            event_line += "{}: {} ".format(key, value)
```
msftristew: Try: `";".join("{}: {}".format(key, value) for key, value in kwargs.iteritems()`

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,32 @@
...
+        self.handler = events_handler()
+
+    def _import_handler(self, name):
+        """
+        Importing the handler based on the passed parameter.
+        """
+        components = name.split('.')
+        mod = __import__(components[0])
+        for comp in components[1:]:
+            mod = getattr(mod, comp)
```
msftristew: What does this do?

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,32 @@
...
+
+class SparkEvents:
+    def __init__(self):
+        """
+        Create an instance from the handler mentioned in the config file.
+        """
+        events_handler = self._import_handler(conf.events_config()["handler"])
+        self.handler = events_handler()
+
+    def _import_handler(self, name):
```
msftristew: How is this API supposed to be used?

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        Importing the handler based on the passed parameter.
+        """
+        components = name.split('.')
+        mod = __import__(components[0])
+        for comp in components[1:]:
+            mod = getattr(mod, comp)
+        return mod
```
msftristew: This doesn't match up with the design we discussed in our meeting.  What is returned from this method?  I thought this class was supposed to create a handler and use it in the `emit_*` methods.

MohamedElKamhawy: We decided to leave this decision for me, remember? And since it's going be a very tiny number of code lines. I found it's better to encapsulate it in a method.

msftristew: NVM, it looks like I misunderstood. This is OK.

remotespark/utils/sparkevents.py
msftristew: No reason for this to be its own method -- just put this in `__init__`

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,32 @@
+import remotespark.utils.configuration as conf
+
+
+class SparkEvents:
+    def __init__(self):
+        """
+        Create an instance from the handler mentioned in the config file.
+        """
+        events_handler = self._import_handler(conf.events_config()["handler"])
```
msftristew: There is no reason for `events_config` to be a dictionary with a `"handler"` element -- just make it a string

msftristew: Maybe call it `events_handler_class` instead of `events_config`

remotespark/utils/eventshandler.py
MohamedElKamhawy: I see that order doesn't matter in case of [file] EventsHandler.

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,32 @@
...
+        self.handler = events_handler()
+
+    def _import_handler(self, name):
+        """
+        Importing the handler based on the passed parameter.
+        """
+        components = name.split('.')
+        mod = __import__(components[0])
```
msftristew: I'm reading online that `__import__` is an internal implementation detail.  Can you use `importlib` instead?

remotespark/utils/eventshandler.py
msftristew: The point is that it will matter for other implementations of `EventsHandler` so you need to be able to have an interface that is appropriately generic that it works for everything.

remotespark/utils/sparkevents.py
MohamedElKamhawy: imports python class dynamically.

remotespark/utils/sparkevents.py
MohamedElKamhawy: Have a loot at this: http://stackoverflow.com/questions/547829/how-to-dynamically-load-a-python-class

msftristew: Yes, but it's still an internal implementation detail that shouldn't be used by us:

> Import a module. Because this function is meant for use by the Python interpreter and not for general use it is better to use importlib.import_module() to programmatically import a module.

msftristew: `importlib` also has an easier API so you don't have to do the thing with the for loop and the `getattr`

MohamedElKamhawy: OK

remotespark/utils/eventshandler.py
MohamedElKamhawy: Do you mean "*args" or "**args"?

msftristew: Yes, `*args`, sorry.

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Yes, it should be a unique for every session created. It'll help us to identify all of the related events to specific session instead of using session Id which can be similar with other ones.

MohamedElKamhawy: Actually, we can store the value without adding it to session state.

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,22 @@
...
+        Create an instance from the handler mentioned in the config file.
+        """
+        events_handler = importlib.import_module(conf.events_handler_class())
+        self.handler = events_handler()
+
+    def emit_start_session_event(self, time_stamp, event_name, session_guid, language):
+        """
+        Emitting Start Session Event
+        """
+        raise NotImplemented
```
aggFTW: This isn't valid Python code.

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def emit_end_session_event(self,time_stamp, event_name, session_guid, language, session_id):
```
aggFTW: Please watch your spaces in-between parameters

remotespark/livyclientlib/livysession.py
```diff
@@ -50,9 +51,9 @@ def __init__(self, http_client, properties, ipython_display,
         self._status_sleep_seconds = status_sleep_seconds
         self._statement_sleep_seconds = statement_sleep_seconds
         self._create_sql_context_timeout_seconds = create_sql_context_timeout_seconds
+        self.session_guid = uuid.uuid4()
```
aggFTW: We should talk about GUIDs for events. There's a bunch of operations that will need GUIDs. We shouldn't pollute all of our methods...

I believe the GUID should be handled by `SparkEvents` behind the covers. Let's talk about it tomorrow.

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,22 @@
...
+
+class SparkEvents:
+    def __init__(self):
+        """
+        Create an instance from the handler mentioned in the config file.
+        """
+        events_handler = importlib.import_module(conf.events_handler_class())
+        self.handler = events_handler()
+
+    def emit_start_session_event(self, time_stamp, event_name, session_guid, language):
```
aggFTW: What's the dimensions for each of these events?

I believe you don't need to pass in the `event_name` or `time_stamp`. Both of these fields can be handled by this class.

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,13 @@
...
+class EventsHandler:
+    def __init__(self):
+        self.logger = Log("EventsHandler")
+
+    def store_event(self, *args):
+        """
+        Storing the Event details using the logger.
+        """
+        event_line = ",".join("{}".format(arg) for arg in args)
+        self.logger.debug(event_line)
```
aggFTW: Call `info` method instead.

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,13 @@
...
+
+class EventsHandler:
+    def __init__(self):
+        self.logger = Log("EventsHandler")
+
+    def store_event(self, *args):
+        """
+        Storing the Event details using the logger.
+        """
+        event_line = ",".join("{}".format(arg) for arg in args)
```
aggFTW: Should this be a key-value string?

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,13 @@
+from remotespark.utils.log import Log
+
+
+class EventsHandler:
+    def __init__(self):
+        self.logger = Log("EventsHandler")
+
+    def store_event(self, *args):
```
aggFTW: If this is a `handler`, should we name the method `handle_event`?

remotespark/utils/sparkevents.py
MohamedElKamhawy: I was thinking about the same thing, this make more sense.

remotespark/utils/sparkevents.py
MohamedElKamhawy: What do you mean?

remotespark/utils/eventshandler.py
MohamedElKamhawy: More meaningful. will change it.

MohamedElKamhawy: I see your point but the main job for this method is just storing the event info, what do you think?

remotespark/utils/eventshandler.py
MohamedElKamhawy: If we want to unify the API across the different handlers the list will be more fit. on the other side, key-value will give more sense to the logs information.  So, I go for the current state, what do you think?

remotespark/utils/eventshandler.py
aggFTW: I prefer handle. @msftristew ?

remotespark/utils/eventshandler.py
aggFTW: What does 
>unify the API across the different handlers

mean?

remotespark/utils/sparkevents.py
aggFTW: Run it on a python terminal. That class doesn't exist and you are not instantiating an object.

remotespark/utils/eventshandler.py
MohamedElKamhawy: When I add the new handler "hdinsighteventshandler" the method there will be the same and it will help to insert them in the MDS table with the same arrangement. 

Let's say we have dim1, dim2..etc. arranged in this way then, we put them in a key-value, we will need to iterate over them to insert them in the right way. 

So, if they are already sorted in a list? so we will insert them easily without need to iterate over the dictionary to put the value for each attribute in the right place.

remotespark/utils/eventshandler.py
msftristew: Agreed with @aggFTW 

If the name is `EventsHandler`, it should handle events.  It should store events if its name is `EventStorage`

remotespark/utils/eventshandler.py
msftristew: Log doesn't have an `info` method -- remember when you wrote that class @aggFTW ? :)

remotespark/utils/eventshandler.py
MohamedElKamhawy: Okay.

remotespark/utils/sparkevents.py
msftristew: Please try to test code before pushing it

remotespark/utils/eventshandler.py
aggFTW: :(

remotespark/livyclientlib/livysession.py
aggFTW: We agreed to make this a decorator.

remotespark/utils/eventshandler.py
aggFTW: You could have sorted key,value pairs. Just don't use a dictionary. Or don't sort them and iterate over them; what's wrong with that?

I'm not focusing on implementation, I'm focusing on building a robust API that any system could use.

remotespark/utils/sparkevents.py
MohamedElKamhawy: I think this is the reason for the error you are getting: http://stackoverflow.com/questions/15514593/importerror-no-module-named-when-trying-to-run-python-script

aggFTW: ?

remotespark/utils/eventshandler.py
msftristew: A list of key-value pairs would satisfy Alejandro's request, @MohamedElKamhawy

remotespark/utils/sparkevents.py
msftristew: Mohamed, Alejandro is talking about how this should be

    raise NotImplementedError()

remotespark/utils/decorators.py
```diff
@@ -0,0 +1,17 @@
...
+    def class_decorator(cls):
+        def get_attr(self, attr=attr_name):
+            return getattr(self, "_" + attr)
+
+        def set_attr(self, value, attr=attr_name):
+            setattr(self, "_" + attr, value)
+
+        prop = property(get_attr, set_attr)
+        setattr(cls, attr_name, prop)
+        setattr(cls, "_" + attr_name, utils.generate_uuid())  # Default value for that attribute
```
msftristew: This adds the same GUID to all sessions (as a class variable).  Two alternatives:

* Edit the class's `__init__` method directly in this decorator

* Eschew the decorator and just use a base class for all things with GUID's in them (I prefer this)

MohamedElKamhawy:
@msftristew , @aggFTW  I have added a new PR, hopefully it's the last before adding the unit tests.

MohamedElKamhawy:
If there is no more comments, I fixed the last comments and I'll push per merging.

remotespark/livyclientlib/sparkcontroller.py
```diff
@@ -73,12 +74,17 @@ def add_session(self, name, connection_string, skip_if_exists, properties):
             return
 
         session = self._create_livy_session(connection_string, properties, self.ipython_display)
+
+        spark_events = SparkEvents()
```
msftristew: Why not create one `SparkEvents` object in the constructor and use it for all events on the `SparkController`?

msftristew:
Mohamed, please do not merge anything until you have it thoroughly tested.

MohamedElKamhawy:
@msftristew  Sure.

remotespark/livyclientlib/livysession.py
```diff
@@ -12,7 +13,8 @@
 from .livyreliablehttpclient import LivyReliableHttpClient
 
 
-class LivySession(object):
+
```
aggFTW: Please revert all whitespace changes on this file. It's no longer PEP compliant.

remotespark/livyclientlib/sparkcontroller.py
aggFTW: What about emitting the events right before we make the call to Livy and right when we get the call that says it's in running state? This includes hiveContext creation, which should be a different operation.

remotespark/utils/constants.py
```diff
@@ -9,6 +9,10 @@
 SESSION_KIND_SPARKR = "sparkr"
 SESSION_KINDS_SUPPORTED = [SESSION_KIND_SPARK, SESSION_KIND_PYSPARK, SESSION_KIND_SPARKR]
 
+SESSION_CREATION_START_EVENT = "notebookSessionCreationStart"
+SESSION_CREATION_END_EVENT = "notebookSessionCreationEnd"
+EVENTS = {SESSION_CREATION_START_EVENT, SESSION_CREATION_END_EVENT}
```
aggFTW: What is this `EVENTS` for?

remotespark/utils/eventshandler.py
```diff
@@ -0,0 +1,13 @@
+from remotespark.utils.log import Log
+
+
+class EventsHandler:
+    def __init__(self):
+        self.logger = Log("EventsHandler")
+
+    def handle_event(self, args):
```
aggFTW: I'd rename `args` to reflect it's a key, value collection

remotespark/livyclientlib/livysession.py
```diff
@@ -12,7 +13,8 @@
...
+class LivySession(GuidMixin, object):
```
aggFTW: Don't you have to call the parent's constructor `__init__` method for the mixin to work?

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,48 @@
...
+import remotespark.utils.configuration as conf
+import remotespark.utils.constants as constants
+
+
+class SparkEvents:
+    def __init__(self):
+        """
+        Create an instance from the handler mentioned in the config file.
+        """
+        module, class_name = conf.events_handler_class().rsplit('.', 1)
```
aggFTW: What if there's no `.` in the config?

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,48 @@
...
+        events_handler_module = importlib.import_module(module)
+        events_handler = getattr(events_handler_module, class_name)
+
+        self.handler = events_handler()
+
+
```
aggFTW: Make whitespace in the file PEP-compliant.

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,48 @@
...
+    def emit_session_creation_start_event(self, session_guid, language):
+        """
+        Emitting Start Session Event
+        """
+        assert language in constants.SESSION_KINDS_SUPPORTED
+
+        event_name = constants.SESSION_CREATION_START_EVENT
+        time_stamp = datetime.now()
```
aggFTW: Getting the current time should be in a method by itself. What is the timezone used by `datetime.now()`?

remotespark/utils/sparkevents.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        assert language in constants.SESSION_KINDS_SUPPORTED
+        assert session_id >= 0
+
+        args = []
+
+        event_name = constants.SESSION_CREATION_END_EVENT
+        time_stamp = datetime.now()
+
+        args = [("TimeStamp", time_stamp), ("EventName", event_name), ("SessionGuid", session_guid),
```
aggFTW: What about the final state?

What metrics will we be able to collect from these events?

tests/test_sparkcontroller.py
```diff
@@ -41,6 +42,8 @@ def test_add_session():
     connection_string = "url=http://location:port;username=name;password=word"
     client = MagicMock()
     session = MagicMock()
+    session.kind = SESSION_KIND_SPARK
```
aggFTW: Why did you add this? Your unit tests should be emitting events to memory and should not be using the real implementation.

remotespark/utils/constants.py
MohamedElKamhawy: For newly added events, this should be reflected here instead of adding it in the SparkEvents file.

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: No, no need for that. I tried it out myself and works fine.

remotespark/livyclientlib/sparkcontroller.py
MohamedElKamhawy: Yes, I agree with you @aggFTW  to be before the call to Livy but for session creation end I think this is the right place for it .

remotespark/utils/sparkevents.py
MohamedElKamhawy: It returns Local date & time. I'll use the utc instead.

remotespark/utils/sparkevents.py
MohamedElKamhawy: The intention for those two events is to get the Average Session Creation Time.

remotespark/utils/sparkevents.py
MohamedElKamhawy: Anyone should use his own handler should add the whole path till the class used as the handler. Else, this will not work (Compilation Error) as we are using importlib_module method which expects module.

remotespark/utils/constants.py
aggFTW: Sorry, didn't get it. Who is going to use this variable?

remotespark/livyclientlib/sparkcontroller.py
aggFTW: That's not true. We can get both before the call to Livy.

remotespark/utils/constants.py
MohamedElKamhawy: This will contain all of our added events to save them in one place.

msftristew: Why?

remotespark/livyclientlib/livysession.py
aggFTW: From what I read, you DO need to call it, no?
http://stackoverflow.com/a/1385836

remotespark/utils/constants.py
MohamedElKamhawy: You suggested that Ricky, remember?

remotespark/utils/sparkevents.py
aggFTW: And what else? That's one of N metrics we want for session creation. Will the schema support all the metrics? For a list of metrics, refer to the email I shared with a list of metrics from Asad.

tests/test_sparkevents.py
```diff
@@ -0,0 +1,43 @@
...
+def test_emit_session_creation_start_event():
+
+    language = constants.SESSION_KIND_SPARK
+    event_name = constants.SESSION_CREATION_START_EVENT
+
+    kwargs_list = [("TimeStamp", spark_events.get_utc_date_time()), ("EventName", event_name), ("SessionGuid", guid),
+                ("SparkLanguage", language)]
+
+    spark_events.emit_session_creation_start_event(guid, language)
+    spark_events.get_utc_date_time.assrt_called_once_with()
```
aggFTW: Typo in assert.

remotespark/livyclientlib/livysession.py
msftristew: Mohamed, the `__init__` method does not get called automatically.

hdiuser@hn0-jupyte:~$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13)
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> class A(object):
...     def __init__(self):
...             print 'yo'
...
>>> class B(object):
...     def __init__(self):
...             print 'wat'
...
>>> b = B()
wat
>>>

remotespark/utils/constants.py
msftristew: No.

remotespark/livyclientlib/livysession.py
aggFTW: How did you try it and had it work Mohamed? I'd be very interested in seeing how you got it to work without calling it explicitly.

remotespark/utils/constants.py
msftristew: The question is, why are you creating a set with all the events in it?

remotespark/utils/sparkevents.py
MohamedElKamhawy: Yes, I believe that. We have discussed that when I sent the design document. We have 11 dimensions, that can be used to satisfy the different metrics mentioned in the email you have sent. For now, let's start with Average Session creation Time as Sangeetha wanted then will add the new metrics gradually as needed.

aggFTW: Talked f2f. You'll go over the document and let us know if we need to log any more information with these events for other metrics so that we don't have breaking changes in the information we are logging for each event.

Please let us know the outcome of your investigation before you merge.

MohamedElKamhawy:
@msftristew , @aggFTW  Can you have a look? It seems that it didn't notify you yesterday with the new commit.

msftristew:
GitHub never notifies us about new commits, please notify us if you have any updates we need to be aware of

MohamedElKamhawy:
@msftristew  Yes, please check the new commit from yesterday.

MohamedElKamhawy:
I have added a new commit, it'll be great if you can give me the whole comments you see to speed the work at my side.

aggFTW:
I cannot give comments on new changes introduced by commits before the commit is made :)

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Here is the code, I tried:

class GuidMixin(object):
    def __init__(self):
        self.guid = generate_uuid()

class Subclass(GuidMixin):

    def printx(self):
        print self.guid

if __name__ == "__main__":
    a = Subclass()
    b = Subclass()

    print a.guid
    a.printx()

    print b.guid
    b.printx()

aggFTW:
LGTM!

msftristew:
Just rename `GuidMixin` to something else (it's not really a mixin) and then LGTM

msftristew:
Is this tested E2E? You've run Jupyter and verified the events are properly logged?

PR #176, opened 2016-02-29
Refactor constants

msftristew:
* Turn constants into a module rather than a class (no reason for it to be a class since instantiating it is useless)

* Use the normal Python formatting for constant variables, i.e. ALL_CAPS_WITH_UNDERCORES


msftristew:
livyclientfactory is still in this branch for some reason?

Deleting it

aggFTW:
Merge as soon as Travis passes.

PR #186, opened 2016-03-09
Autoviz works on browser refresh

### LINKED ISSUE #183: Autoviz does not work if you refresh page
Steps to repro:

1. Create a visualization

2. Refresh page (visualization disappears when page reloads -- this is expected behavior)

3. Try to create another visualization -- you'll get a Plotly error

This would be resolved by putting the `init_notebook_mode` Javascript somewhere else besides inside the widget

aggFTW:
Closes #183

With this solution:
* Browser refresh works
* File size never increases because we are writing to widget, not notebook


remotespark/datawidgets/plotlygraphs/graphrenderer.py
```diff
@@ -13,15 +13,10 @@
...
-    already_initialized = False
-
-    def __init__(self, testing=False):
-        if not testing and not GraphRenderer.already_initialized:
-            init_notebook_mode()
-            GraphRenderer.already_initialized = True
-
     @staticmethod
     def render(df, encoding, output):
+        with output:
```
msftristew: For my own edification, can you describe why this works but the previous solution didn't?

aggFTW: The previous solution wrote to file while this one writes to ouput (which is a widget). Hence, notebook file size does not increase.

msftristew:
As long as you've verified that notebook size doesn't blow up with this solution, then feel free to merge.

I wish there were a solution that only required us to call the `init_notebook_mode()` function when necessary (so we don't have to run 5MB of Javascript on every single autoviz load), but I don't think it's possible.

PR #188, opened 2016-03-11
Converting Guid & Datetime to strings

MohamedElKamhawy:
Datetime & Guid objects are not JSON serializable so I converted them to strings before logging them.

### DIFF
--- a/remotespark/utils/sparkevents.py
+++ b/remotespark/utils/sparkevents.py
@@ -22,10 +22,10 @@ class SparkEvents:
         assert language in constants.SESSION_KINDS_SUPPORTED
 
         event_name = constants.SESSION_CREATION_START_EVENT
-        time_stamp = SparkEvents.get_utc_date_time()
+        time_stamp = "{}".format(SparkEvents.get_utc_date_time())
 
-        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp), (constants.SESSION_GUID, session_guid),
-                       (constants.LIVY_KIND, language)]
+        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp),
+                       (constants.SESSION_GUID, "{}".format(session_guid)), (constants.LIVY_KIND, language)]
 
         self.handler.handle_event(kwargs_list)
 
@@ -37,10 +37,11 @@ class SparkEvents:
         assert session_id >= 0
 
         event_name = constants.SESSION_CREATION_END_EVENT
-        time_stamp = SparkEvents.get_utc_date_time()
+        time_stamp = "{}".format(SparkEvents.get_utc_date_time())
 
-        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp), (constants.SESSION_GUID, session_guid),
-                       (constants.LIVY_KIND, language), (constants.SESSION_ID, session_id), (constants.STATUS, status)]
+        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp),
+                       (constants.SESSION_GUID, "{}".format(session_guid)), (constants.LIVY_KIND, language),
+                       (constants.SESSION_ID, session_id), (constants.STATUS, status)]
 
         self.handler.handle_event(kwargs_list)
 
--- a/remotespark/utils/sparkevents.py
+++ b/remotespark/utils/sparkevents.py
@@ -22,10 +22,10 @@ class SparkEvents:
         assert language in constants.SESSION_KINDS_SUPPORTED
 
         event_name = constants.SESSION_CREATION_START_EVENT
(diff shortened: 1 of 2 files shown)


MohamedElKamhawy:
I'll check the unit tests for this, it seems they are going to fail.

remotespark/utils/sparkevents.py
```diff
@@ -22,10 +22,10 @@ def emit_session_creation_start_event(self, session_guid, language):
         assert language in constants.SESSION_KINDS_SUPPORTED
 
         event_name = constants.SESSION_CREATION_START_EVENT
-        time_stamp = SparkEvents.get_utc_date_time()
+        time_stamp = "{}".format(SparkEvents.get_utc_date_time())
```
msftristew: 1. `get_utc_date_time` should begin with `_` if it's private

2. Use `strftime` to format date strings

remotespark/utils/sparkevents.py
```diff
@@ -22,10 +22,10 @@ def emit_session_creation_start_event(self, session_guid, language):
...
 
-        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp), (constants.SESSION_GUID, session_guid),
-                       (constants.LIVY_KIND, language)]
+        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp),
+                       (constants.SESSION_GUID, "{}".format(session_guid)), (constants.LIVY_KIND, language)]
```
msftristew: What type is `session_guid`? Isn't it a string?

msftristew:
Is this necessary?  Why are you converting things to JSON?

remotespark/utils/sparkevents.py
MohamedElKamhawy: No, `utils.generate_uuid()` is returning guid.

msftristew: Use the `str()` function to convert things to strings

MohamedElKamhawy: I used it but I preferred to have the same pattern like we are using `.format` method.

msftristew: No, it's better to use the `str()` function since you're not actually doing any string formatting

tests/test_sparkevents.py
```diff
@@ -10,8 +10,8 @@ def _setup():
     spark_events = SparkEvents()
     spark_events.handler = MagicMock()
     SparkEvents.get_utc_date_time = MagicMock()
-    time_stamp = spark_events.get_utc_date_time()
-    guid = utils.generate_uuid()
+    time_stamp = "{}".format(spark_events.get_utc_date_time())
```
msftristew: `strftime`

tests/test_sparkevents.py
```diff
@@ -10,8 +10,8 @@ def _setup():
...
+    guid = "{}".format(utils.generate_uuid())
```
msftristew: `strftime`

remotespark/utils/sparkevents.py
```diff
@@ -37,10 +37,11 @@ def emit_session_creation_end_event(self, session_guid, language, session_id, st
         assert session_id >= 0
 
         event_name = constants.SESSION_CREATION_END_EVENT
-        time_stamp = SparkEvents.get_utc_date_time()
+        time_stamp = "{}".format(SparkEvents.get_utc_date_time())
 
-        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp), (constants.SESSION_GUID, session_guid),
-                       (constants.LIVY_KIND, language), (constants.SESSION_ID, session_id), (constants.STATUS, status)]
+        kwargs_list = [(constants.EVENT_NAME, event_name), (constants.TIMESTAMP, time_stamp),
+                       (constants.SESSION_GUID, "{}".format(session_guid)), (constants.LIVY_KIND, language),
```
msftristew: `str()`

remotespark/utils/sparkevents.py
```diff
@@ -37,10 +37,11 @@ def emit_session_creation_end_event(self, session_guid, language, session_id, st
         assert session_id >= 0
 
         event_name = constants.SESSION_CREATION_END_EVENT
-        time_stamp = SparkEvents.get_utc_date_time()
+        time_stamp = "{}".format(SparkEvents.get_utc_date_time())
```
msftristew: `strftime`

MohamedElKamhawy:
@msftristew  Converting them as I need to log a JSON obj to MDS table.

msftristew:
Does it make sense to convert every datetimes to a string automatically at this level of the API even if other metrics backends want to do something different with datetimes?

MohamedElKamhawy:
@msftristew  Good Point, I'll revert the whole changes and do the needed changes at my side to allow any other backend to deal in the way suitable for it.

msftristew:
Mohamed, when you have a pull request you don't want to merge, just don't merge it.  Merging pull requests like this just messes with our Git history.

PR #190, opened 2016-03-15
Convert bools to string before displaying

### LINKED ISSUE #179: Can't plot booleans
Small proof of concept in Python:

    from pyspark.sql.types import Row
    Item = Row('b', 'i')
    rdd = sc.parallelize([Item(True, 100), Item(False, 200)])
    df = sqlContext.createDataFrame(rdd)
    df.registerTempTable('df')
    
    %%sql
    SELECT * FROM df

Then choose Bar -> X = b, Y = i (a plot which should work), and no data shows up.  

I traced this to an issue with plotly.  Plotly doesn't get how to plot bools, apparently.  Proof of concept:

    %%local
    from plotly.graph_objs import Bar, Figure, Data
    from plotly.offline import iplot
    l = Bar(x=[True, False], y=[100,200])
    fig = Figure(data=Data([l]))
    iplot(fig)

How do we solve this?  Is it a matter of coercing boolean columns to strings before plotting?

msftristew:
Closes #179 

Convert booleans to strings in the dataframe before displaying.  We do this because plotly doesn't know how to plot booleans, but it's better at plotting strings.

### DIFF
--- a/remotespark/datawidgets/autovizwidget.py
+++ b/remotespark/datawidgets/autovizwidget.py
@@ -40,7 +40,7 @@ class AutoVizWidget(FlexBox):
             ipython_display = IpythonDisplay()
         self.ipython_display = ipython_display
 
-        self.df = df
+        self.df = self._convert_to_displayable_dataframe(df)
 
         self.encoding = encoding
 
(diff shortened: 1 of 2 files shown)


aggFTW:
Merge whenever

PR #194, opened 2016-03-16
%%config now a proper cell magic

### LINKED ISSUE #192: %%config API change
The `%%config` magic right now takes its argument as a line magic would:

    %%config { ... }

Since these configuration dictionaries can get pretty big and they should stretch across multiple lines, it needs to take its argument as a cell magic:

    %%config
    { 
        ....
    }

msftristew:
Closes #192 

Now, `%%config` takes its argument in the cell rather than in-line, like so:

    %%config
    { "blah": "blah" }

... so its API is now more like a proper cell magic.

### DIFF
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -127,17 +127,8 @@ class KernelMagics(SparkMagicBase):
     @magic_arguments()
     @cell_magic
     @argument("-f", "--force", type=bool, default=False, nargs="?", const=True, help="If present, user understands.")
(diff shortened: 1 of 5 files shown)


aggFTW:
This is not what Arijit was asking for though... he wanted line and cell to be merged.

msftristew:
Arijit was asking for a %%config API that allows him to spread JSON objects over multiple lines -- this is it, and it doesn't introduce any weird inconsistencies with the way other cell magics work.

msftristew:
bump

PR #198, opened 2016-03-22
Display fatal error and delete session if sql context creation errors…

MohamedElKamhawy:
Not tested yet, need to figure out a way to repro the way I found the error


msftristew:
If you remove the `if` statement and just raise a `ValueError` at that location, you'll be able to see what the output would look like from the notebook view.

aggFTW:
And you should have unit tests for this, from the magics all the way to LivySession

aggFTW:
I'll review once unit tests have been added all the way through.

MohamedElKamhawy:
I tested the code locally and added the needed unit tests.

msftristew:
1. Is the error message the user gets sensible?

2. Does it prevent the user from executing more cells when this fails?

MohamedElKamhawy:
@msftristew  I mentioned "Failed to create the SqlContext" and added the exception. 
If the user tries again to run any cell, he'll get the same message. (above)

MohamedElKamhawy:
It seems the build fails.

msftristew:
merge when tests pass

msftristew:
And when Alejandro says OK :)

tests/test_livysession.py
```diff
@@ -436,6 +437,21 @@ def test_create_sql_hive_context_happens_once(self):
...
+
+        session._get_sql_context_creation_command.return_value.execute.return_value = (True, "Success")
+        session.created_sql_context = None
+        session.create_sql_context()
+        assert session.created_sql_context
+
+        session._get_sql_context_creation_command.return_value.execute.return_value = (False, "Exception")
+        session.created_sql_context = None
+        try:
+            session.create_sql_context()
```
msftristew: You need to insert another `assert` in here to make sure that the function threw an exception

MohamedElKamhawy: I added it now in a new commit

tests/test_livysession.py
```diff
@@ -39,7 +39,8 @@ class TestLivySession:
     ready_statement_json = json.loads('{"id":0,"state":"available","output":{"status":"ok","execution_count":0,"data":{"text/plain":"Pi is roughly 3.14336"}}}')
     log_json = json.loads('{"id":6,"from":0,"total":212,"log":["hi","hi"]}')
 
-    def __init__(self):
+    def __init__(self, methodName="test_create_sql_hive_context_happens_once"):
```
msftristew: What does this do?

MohamedElKamhawy: This allows me to make sure this method is throwing an exception.

msftristew: I am asking specifically what this `methodName` addition is for.

MohamedElKamhawy: I'm inheriting unittest class and this is the way to call the super class by specifying the method that you apply your test case to.

msftristew: But the `TestLivySession` class is way bigger than just your test.

tests/test_livysession.py
msftristew: The change you made wasn't really necessary, just `assert False` at this location (like I suggested) to make sure an exception was thrown

tests/test_livysession.py
MohamedElKamhawy: Talked face to face, all tests are running and the current work is okay.

MohamedElKamhawy:
@aggFTW  Can you have a look?

remotespark/livyclientlib/livysession.py
```diff
@@ -90,11 +89,14 @@ def create_sql_context(self):
         self.ipython_display.writeln("Creating HiveContext as 'sqlContext'")
         command = self._get_sql_context_creation_command()
         try:
-            command.execute(self)
+            (success, out) = command.execute(self)
         except LivyClientTimeoutError:
             raise LivyClientTimeoutError("Failed to create the SqlContext in time. Timed out after {} seconds."
                                          .format(self._wait_for_idle_timeout_seconds))
-        self.created_sql_context = True
+        if success:
```
aggFTW: You just modified the behavior of livysession (it can now raise a ValueError). What class consumes this exception? I think it should be the magics and should set a fatal error as True. Is that already happening here?

For this to hold True, all intermediate classes between livysession and kernelmagics must not catch any ValueErrors in the methods that get called. Is that already True? Can we unit test that?

remotespark/livyclientlib/livysession.py
```diff
@@ -90,11 +89,14 @@ def create_sql_context(self):
...
+            self.created_sql_context = True
+        else:
+            raise ValueError("Failed to create the SqlContext.\nError {}".format(out))
```
aggFTW: I would put a semicolon after Error and quote the actual error:
`raise ValueError("Failed to create the SqlContext.\nError: '{}'".format(out))`

What do you think?

tests/test_livysession.py
```diff
@@ -436,6 +436,22 @@ def test_create_sql_hive_context_happens_once(self):
...
+
+        session._get_sql_context_creation_command.return_value.execute.return_value = (False, "Exception")
+        session.created_sql_context = None
+
+        try:
+            session.create_sql_context()
+            assert False
+        except ValueError:
+            assert session.created_sql_context is None
+
```
aggFTW: This assertion was valuable:

`assert ex.message == "Failed to create the SqlContext.\nError {}".format("Exception")`

Can you please add it back?

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Yes, this is true, no class catches this exception at all.
I already tested this and none catches this exception.

tests/test_livysession.py
MohamedElKamhawy: message attribute is not existing in Python 3 and I believe the current test is sufficient.

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: As a user, I prefer to see the summary first and in the second line to see the root cause for the error.

MohamedElKamhawy: But, I'm fine with either.

aggFTW: What I suggested keeps the number of lines and placement. It's just adding a semi-colon and quotes to denote that the Error is not coming from us, but from Spark.

tests/test_livysession.py
aggFTW: You can always get the ValueError text. Just do a str(e).

remotespark/livyclientlib/livysession.py
aggFTW: It is true now, but without the unit tests to all those intermediate classes, you won't know that it will never change. Once you write those tests, it would be good to say why those classes should not catch ValueErrors in the unit tests.

Please add them to prevent regressions.

remotespark/livyclientlib/livysession.py
msftristew: That's a colon, not a semicolon :)

aggFTW: Thanks Ricky :)

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Can you please explain more about what do you think about that?

MohamedElKamhawy: Start method - in LivySession Class - throw an exception if the session is not created (Id < 0), I can't find the relevant unit test for that? Can you point me to it to do the same?

MohamedElKamhawy: Any help?

msftristew: I'm advising Mohamed to add, at the very least, to add a unit test to the magics class to ensure it logs a fatal error when adding the session fails (if that unit test doesn't already exist).  He can also add a unit test to `SparkController` to verify it raises an exception if `session.start()` throws which would be very easy.  That would be enough verification.

Another thing you can do for some further assurance that this exception will not be spontaneously caught is to use a different exception than `ValueError`, like a custom exception `FailedToCreateSqlContextError`.

MohamedElKamhawy:
Addressing Feedback & Added the unit tests for the intermediate classes.

remotespark/livyclientlib/livysession.py
aggFTW: This feedback is not addressed on the latest iteration. We meant that you should at least add a unit test to the test_sparkcontroller.py file for when session.start() fails as well as a unit test to the test_kernelmagics.py file for when sparkcontroller.add_session() fails.

tests/test_livysession.py
```diff
@@ -436,6 +438,49 @@ def test_create_sql_hive_context_happens_once(self):
                in http_client.post_statement.call_args_list
         assert len(http_client.post_statement.call_args_list) == 1
 
+        session._get_sql_context_creation_command = MagicMock()
```
aggFTW: This test is massive now. Can you please break it up into different tests. The current test should only `test_create_sql_hive_context_happens_once`. This test is now testing what happens on failure.

tests/test_livysession.py
```diff
@@ -26,7 +27,7 @@ def status_code(self):
 CONN_STR = 'url=https://www.DFAS90D82309F0W9ASD0F9ZX.com;username=abcd;password=1234'
 
 
-class TestLivySession:
+class TestLivySession():
```
msftristew: `TestLivySession` should inherit from `object`

MohamedElKamhawy: This is not part of my PR, I'll add it now.

tests/test_livysession.py
MohamedElKamhawy: Agreed

MohamedElKamhawy:
Addressing the feedback.

tests/test_livysession.py
```diff
@@ -436,6 +437,72 @@ def test_create_sql_hive_context_happens_once(self):
...
+            "status_sleep_seconds": 0.01,
+            "statement_sleep_seconds": 0.01
+        })
+        session = self._create_session(kind=kind, http_client=http_client)
+        session.ipython_display = ipython_display
+        conf.load()
+
+        session._get_sql_context_creation_command = MagicMock()
+
+        session._get_sql_context_creation_command.return_value.execute.return_value = (True, "Success")
```
aggFTW: This is testing success, when the unit test name is `test_create_sql_context_failure`. Can you please break it up into different unit tests?

Also, the setup code is copy pasted from another unit test. It warrants a helper function.

tests/test_livysession.py
```diff
@@ -436,6 +437,72 @@ def test_create_sql_hive_context_happens_once(self):
...
+        session.created_sql_context = None
+
+        try:
+            session.create_sql_context()
+            assert False
+        except ValueError as ex:
+            assert str(ex) == "Failed to create the SqlContext.\nError, '{}'".format("Exception")
+            assert session.created_sql_context is None
+
+        kernelMagics = KernelMagics(shell=None)
```
msftristew: Why is this here and not in `test_kernel_magics` (the file where we're doing the rest of the testing of kernel magics)?

aggFTW: What is a kernel magics test doing in test_livysession.py?

Why is this unit test testing more than one thing again?

tests/test_livysession.py
```diff
@@ -436,6 +437,72 @@ def test_create_sql_hive_context_happens_once(self):
...
+        kernelMagics.language = constants.LANG_SCALA
+        kernelMagics.ipython_display = ipython_display
+        kernelMagics.ipython_display.send_error = MagicMock()
+
+        kernelMagics._do_not_call_start_session("Test Line")
+        assert kernelMagics.fatal_error
+        assert kernelMagics.fatal_error_message == conf.fatal_error_suggestion().format(
+            "Failed to create the SqlContext.\nError, '{}'".format("Exception"))
+
+        sparkController = SparkController(ipython_display)
```
aggFTW: What is a spark controller test doing in test_livysession.py?

Why is this unit test testing more than one thing again?

aggFTW:
This is very important:

1. Unit tests should only be testing one thing at a time. **Maybe** two if the two things are very similar... but they should be testing a **unit*. This is so that when a unit test fails, we know exactly what is failing. If the unit test is testing 4 things at once and it fails, which of the 4 things failed? I need to go debug to know that and that is wasted time.
2. Every class is in a different file. A class is some code that is isolated of its dependencies. So, if you are _unit testing_ a class, you should be testing it isolated of its dependencies, so that even if dependencies fail, your unit tests do not get affected. Unit tests are grouped into different files because every test file unit tests just one class. If you are importing dependencies, [like this](https://github.com/jupyter-incubator/sparkmagic/pull/198/files#diff-498e0d5682020c4fc81dfb940efe63feR8), it's a pretty good indication of a [code smell](https://en.wikipedia.org/wiki/Code_smell).

So, please break your unit tests into **unit** tests and put them in the files they belong at.

remotespark/livyclientlib/livysession.py
MohamedElKamhawy: Done

MohamedElKamhawy:
I have pushed a new one, and applied the different feedback I got. 
@msftristew @aggFTW  Thank you for your patience, and for helping. :)

tests/test_sparkcontroller.py
```diff
@@ -193,3 +193,24 @@ def test_get_logs():
...
+@with_setup(_setup, _teardown)
+def test_start_throw_exception():
+    name = "name"
+    properties = {"kind": "spark"}
+    endpoint = Endpoint("http://location:port", "name", "word")
+    session = MagicMock()
+
+    controller._livy_session = MagicMock(return_value=session)
+    controller._http_client = MagicMock(return_value=MagicMock())
+    session.start = dummy_create_sql_context
```
msftristew: Use `MagicMock` for this, then you can also assert the session was started

tests/test_kernel_magics.py
```diff
@@ -417,3 +417,20 @@ def test_delete_with_force_different_session():
...
+    magic.ipython_display = ipython_display
+    magic.ipython_display.send_error = MagicMock()
+
+    magic._do_not_call_start_session("Test Line")
+    assert magic.fatal_error
+    assert magic.fatal_error_message == conf.fatal_error_suggestion().format(
+        "Failed to create the SqlContext.\nError, '{}'".format("Exception"))
+
+
+def dummy_add_session(*args):
```
msftristew: Use MagicMock for this, then you can also assert the method was called

MohamedElKamhawy: Added in the new push

tests/test_sparkcontroller.py
MohamedElKamhawy: Added in the new push

msftristew:
LGTM, just resolve your conflicts

tests/test_kernel_magics.py
```diff
@@ -417,3 +417,20 @@ def test_delete_with_force_different_session():
 
     spark_controller.get_session_id_for_client.assert_called_once_with(magic.session_name)
     spark_controller.delete_session_by_id.assert_called_once_with(magic.endpoint, session_id)
+
+
+@with_setup(_setup, _teardown)
+def test_add_session_throw_exception():
```
aggFTW: A better name would be `test_start_session_displays_fatal_error_when_session_throws`

MohamedElKamhawy: isn't it too big?

tests/test_livysession.py
```diff
@@ -439,6 +440,32 @@ def test_create_sql_hive_context_happens_once(self):
...
         assert len(http_client.post_statement.call_args_list) == 1
 
+        session._get_sql_context_creation_command = MagicMock()
+
+        session._get_sql_context_creation_command.return_value.execute.return_value = (True, "Success")
+        session.created_sql_context = None
+        session.create_sql_context()
+        assert session.created_sql_context
+
+    def test_create_sql_context_throw_exception(self):
```
aggFTW: `test_create_sql_context_throws_when_command_fails`

tests/test_kernel_magics.py
msftristew: No

tests/test_sparkcontroller.py
```diff
@@ -193,3 +193,23 @@ def test_get_logs():
 def test_get_session_id_for_client():
     assert controller.get_session_id_for_client("name") is not None
     client_manager.get_session_id_for_client.assert_called_once_with("name")
+
+
+@with_setup(_setup, _teardown)
+def test_start_throw_exception():
```
aggFTW: `test_add_session_throws_when_session_start_fails`

MohamedElKamhawy: Any other comments? If no, will update per merging.

tests/test_sparkcontroller.py
```diff
@@ -193,3 +193,23 @@ def test_get_logs():
...
+    controller._livy_session = MagicMock(return_value=session)
+    controller._http_client = MagicMock(return_value=MagicMock())
+    e = ValueError("Failed to create the SqlContext.\nError, '{}'".format("Exception"))
+    session.start = MagicMock(side_effect=e)
+
+    try:
+        controller.add_session(name, endpoint, False, properties)
+        assert False
+    except ValueError as ex:
+        assert str(ex) == "Failed to create the SqlContext.\nError, '{}'".format("Exception")
```
aggFTW: `assert str(ex) == str(e)`

tests/test_kernel_magics.py
aggFTW: Too big for?

MohamedElKamhawy: As a method name, it's too big.

msftristew: No

aggFTW: The method name is too big for what?

aggFTW:
Mohamed, please don't merge until we see the conflicts resolved. So let us know when you resolve the conflicts and we'll take a last look.

tests/test_kernel_magics.py
MohamedElKamhawy: Too big from naming convention perspective :)

aggFTW: What naming convention?

I prefer clarity over anything? What about you?

MohamedElKamhawy: OK

MohamedElKamhawy:
Please, have a look to merge.

tests/test_livysession.py
```diff
@@ -1,11 +1,12 @@
 ﻿import json
-
 from mock import MagicMock, call
 from nose.tools import raises, assert_equals
 
 from remotespark.livyclientlib.livyclienttimeouterror import LivyClientTimeoutError
 from remotespark.livyclientlib.livyunexpectedstatuserror import LivyUnexpectedStatusError
 from remotespark.livyclientlib.livysession import LivySession
+from remotespark.kernels.kernelmagics import KernelMagics
+from remotespark.livyclientlib.sparkcontroller import SparkController
```
aggFTW: These imports are not needed

MohamedElKamhawy: Sure, any other comments?

tests/test_kernel_magics.py
```diff
@@ -531,6 +531,21 @@ def test_delete_with_force_different_session():
...
+    magic.ipython_display = ipython_display
+    magic.ipython_display.send_error = MagicMock()
+
+    magic._do_not_call_start_session("Test Line")
+
+    magic.spark_controller.add_session.assert_called_once_with(magic.session_name, magic.endpoint, False,
+                                                               {"kind": constants.SESSION_KIND_SPARK})
+    assert magic.fatal_error
+    assert magic.fatal_error_message == conf.fatal_error_suggestion().format(str(e))
+
```
aggFTW: Two new lines between tests.

aggFTW:
Looks good! Just the two small comments :smile:

aggFTW:
Nice! Merge after Travis passes.

MohamedElKamhawy:
closes #198

PR #199, opened 2016-03-22
Add scatter plots

msftristew:
Add scatter plots to autovisualization options.


aggFTW:
Well, that was easy :smile: 

Any performance testing done on this?

msftristew:
Yes.  Note that we're using the same plotting class that the `LineGraph` class is using, so performance of scatter plots should be about the same as performance for line graphs.

In my testing I was able to plot about 50,000 points somewhat gracefully:

image

The plot takes a few seconds to generate and it makes the browser a bit less responsive, but that's all.

Plotting 5,000 points (twice our current default maximum number of rows) is quick.

aggFTW:
Rubberstamp

PR #201, opened 2016-03-23
Display 'no results' message if empty dataframe came back from Livy server for SQL queries

### LINKED ISSUE #200: CREATE TABLE without IF NOT EXISTS fails
%%sql  
     CREATE TABLE bar (
       baz int
     )

This fails with this error:

     An error occurred while calling o56.sql.
: org.apache.spark.sql.execution.QueryExecutionException: FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.exec.DDLTask. AlreadyExistsException(message:Table bar already exists)

The SQL query actually gets executed twice (once to figure out the dataframe has no data, and another time to get the column names).  But it fails to execute the query the second time because the table was created the first time.

You can work around this issue by doing

    %%sql
    CREATE TABLE IF NOT EXISTS bar (...

... but it would be ideal if we could avoid this altogether.

I do not know if it's possible to do queries like `INSERT` in Hive, but this would also affect that by running the `INSERT` query twice.

msftristew:
Closes #200 

Only run SQL queries against the server once and display "No results." if a SQL query had no results.

image

### DIFF
--- a/remotespark/datawidgets/plotlygraphs/graphrenderer.py
+++ b/remotespark/datawidgets/plotlygraphs/graphrenderer.py
@@ -20,6 +20,7 @@ class GraphRenderer(object):
             init_notebook_mode()
         GraphRenderer._get_graph(encoding.chart_type).render(df, encoding, output)
 
+
     @staticmethod
     def display_x(chart_type):
         return GraphRenderer._get_graph(chart_type).display_x()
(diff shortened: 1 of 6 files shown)


remotespark/datawidgets/plotlygraphs/graphrenderer.py
```diff
@@ -20,6 +20,7 @@ def render(df, encoding, output):
             init_notebook_mode()
         GraphRenderer._get_graph(encoding.chart_type).render(df, encoding, output)
 
+
```
aggFTW: Nit: Remove the whitespace

aggFTW:
LFTM

PR #205, opened 2016-03-25
Add unit tests for session creation events. Also emit end event on exceptions.

aggFTW:


remotespark/livyclientlib/livysession.py
```diff
@@ -65,21 +68,26 @@ def start(self, create_sql_context=True):
...
+                self.wait_for_idle(conf.livy_session_startup_timeout_seconds())
+            except LivyClientTimeoutError:
+                raise LivyClientTimeoutError("Session {} did not start up in {} seconds."\
+                                             .format(self.id, conf.livy_session_startup_timeout_seconds()))
+
+            if create_sql_context:
+                self.create_sql_context()
+        except Exception as e:
+            self._spark_events.emit_session_creation_end_event(self.guid, self.kind, self.id, self.status,
+                                                               False, e.__class__.__name__, str(e))
```
msftristew: You need to re-raise the exception afterward.

msftristew:
lgtm

remotespark/livyclientlib/livysession.py
aggFTW: Thanks for catching that! Fixed and unit tested.

PR #206, opened 2016-03-25
Livy session emits deletion events

aggFTW:


aggFTW:
I'm merging just so that I can submit the PR for `emit_library_loaded_event`. I can address any comments later on.

tests/test_livysession.py
```diff
@@ -540,14 +539,18 @@ def test_start_emits_start_end_failed_session_when_sql_fails(self):
         kind = constants.SESSION_KIND_SPARK
         session = self._create_session(kind=kind, http_client=http_client, spark_events=spark_events)
         session.create_sql_context = MagicMock(side_effect=ValueError)
-        session.start()
+
+        try:
+            session.start()
```
msftristew: To ensure that `.start()` threw an exception, add an "`assert False`" in the next line.  Also applies for line 570, 597, 653

PR #216, opened 2016-04-01
Better error messages for magics

### LINKED ISSUE #215: Bad error messages when you forget arguments to magics
%%delete 

This throws a confusing exception rather than explaining to the user that they're missing a `-s` argument.

msftristew:
Closes #215

* Add new exceptions. There are two "types" of exceptions, "expected" exceptions (those that we are likely to find in day-to-day use: HTTP client failing, bad input data to methods, etc.) and "unexpected" or "internal" exceptions (which we hide from the user and prompt them to send an issue to the Github repository)

* Add decorators and utilities for detecting errors and displaying appropriate messages to the user

* Different logic for error handling and reporting, especially in remotesparkmagics

* Remove some input data validation from SparkEvents which can throw errors under certain error conditions.

* Lots more unit tests

### DIFF
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -287,8 +287,10 @@ class KernelMagics(SparkMagicBase):
     @cell_magic
     @handle_expected_exceptions
     def _do_not_call_delete_session(self, line, cell="", local_ns=None):
-        if self.session_started:
-            self.spark_controller.delete_session_by_name(self.session_name)
+        try:
+            if self.session_started:
+                self.spark_controller.delete_session_by_name(self.session_name)
+        finally:
             self.session_started = False
 
     @magic_arguments()
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -15,11 +15,10 @@ from IPython.core.magic_arguments import argument, magic_arguments, parse_argstr
 import remotespark.utils.configuration as conf
 from remotespark.livyclientlib.command import Command
 from remotespark.livyclientlib.endpoint import Endpoint
-from remotespark.livyclientlib.sqlquery import SQLQuery
 from remotespark.magics.sparkmagicsbase import SparkMagicBase
 from remotespark.utils.constants import LANGS_SUPPORTED
 from remotespark.utils.sparkevents import SparkEvents
-from remotespark.utils.utils import generate_uuid, get_livy_kind
+from remotespark.utils.utils import generate_uuid, get_livy_kind, wrap_unexpected_exceptions
 
 
 def _event(f):
@@ -60,6 +59,7 @@ class KernelMagics(SparkMagicBase):
         self._spark_events = spark_events
 
     @cell_magic
+    @wrap_unexpected_exceptions
     @_event
     def help(self, line, cell="", local_ns=None):
         help_html = """
@@ -127,6 +127,7 @@ class KernelMagics(SparkMagicBase):
         raise NotImplementedError("UserCodeParser should have prevented code execution from reaching here.")
 
     @cell_magic
+    @wrap_unexpected_exceptions
     @_event
     def info(self, line, cell="", local_ns=None):
         self.ipython_display.writeln("Endpoint:\n\t{}\n".format(self.endpoint.url))
@@ -140,6 +141,7 @@ class KernelMagics(SparkMagicBase):
         self.print_endpoint_info(info_sessions)
 
     @cell_magic
+    @wrap_unexpected_exceptions
     @_event
     def logs(self, line, cell="", local_ns=None):
         if self.session_started:
@@ -154,6 +156,7 @@ class KernelMagics(SparkMagicBase):
     @magic_arguments()
     @cell_magic
     @argument("-f", "--force", type=bool, default=False, nargs="?", const=True, help="If present, user understands.")
+    @wrap_unexpected_exceptions
     @_event
     def configure(self, line, cell="", local_ns=None):
         args = parse_argstring(self.configure, line)
@@ -171,6 +174,7 @@ class KernelMagics(SparkMagicBase):
         self.info("")
 
     @cell_magic
+    @wrap_unexpected_exceptions
     def spark(self, line, cell="", local_ns=None):
         if self._do_not_call_start_session(""):
             (success, out) = self.spark_controller.run_command(Command(cell))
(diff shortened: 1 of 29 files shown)


remotespark/kernels/kernelmagics.py
```diff
@@ -264,9 +285,12 @@ def _do_not_call_start_session(self, line, cell="", local_ns=None):
...
 
     @cell_magic
+    @handle_expected_exceptions
     def _do_not_call_delete_session(self, line, cell="", local_ns=None):
-        if self.session_started:
-            self.spark_controller.delete_session_by_name(self.session_name)
+        try:
+            if self.session_started:
+                self.spark_controller.delete_session_by_name(self.session_name)
+        finally:
```
aggFTW: Can you skip the except?

msftristew: >>> try:
    ...   raise ValueError('hi')
    ... finally:
    ...   print 'yo'
    ...
    yo
      File "<stdin>", line 2, in <module> 
    ValueError: hi

aggFTW: Should we log the failure?

msftristew: The exception will be logged as an event, so I'm unsure if we get anything out of catching and logging the exception here.

remotespark/livyclientlib/exceptions.py
```diff
@@ -0,0 +1,34 @@
+class LivyClientLibException(Exception):
```
aggFTW: What's the model for Exceptions for this library after this change? Should a developer never raise an exception purposefully that is not a `LivyClientLibException`?

Can we document it in this class?

msftristew: Probably not, I don't think directly raising an exception that is not a `LivyClientLibException` is a good idea.  We can document it.

remotespark/utils/constants.py
```diff
@@ -68,3 +68,7 @@
 DELETE_SESSION_ACTION = "delete"
 START_SESSION_ACTION = "start"
 DO_NOTHING_ACTION = "nothing"
+
+INTERNAL_ERROR_MSG = "An internal error was encountered.\n" \
+                     "Please file an issue at https://github.com/jupyter-incubator/sparkmagic\nError:\n{}"
```
aggFTW: What about adding a stack trace here?

aggFTW: I see what you did. Nevermind

msftristew: We write it to the logger rather than exposing it directly to the user: https://github.com/jupyter-incubator/sparkmagic/pull/216/files/ab618e5e802659f92ac55af90614279a61e055ea#diff-dc25b4292d5e26ef9ba1605e37770b67R64

remotespark/utils/utils.py
```diff
@@ -5,21 +5,72 @@
...
 
+def handle_expected_exceptions(f):
+    """A decorator that handles expected exceptions. Self can be any object with
+    an "ipython_display" attribute.
+    Usage:
+    @handle_expected_exceptions
+    def fn(self, ...):
+        etc..."""
+    import remotespark.livyclientlib.exceptions as e
+    exceptions_to_handle = (e.BadUserDataException, e.LivyUnexpectedStatusException, e.FailedToCreateSqlContextException,
```
aggFTW: Should this be in the exceptions file so that it's easy to find?

msftristew: I could go either way on that, doesn't matter much to me.

tests/test_sqlquery.py
```diff
@@ -7,9 +7,11 @@
 from remotespark.utils.constants import LONG_RANDOM_VARIABLE_NAME
 from remotespark.livyclientlib.sqlquery import SQLQuery
 from remotespark.livyclientlib.command import Command
+from remotespark.livyclientlib.exceptions import BadUserDataException
 import remotespark.utils.configuration as conf
 
 
+
```
aggFTW: Nit: third new line is not needed

remotespark/utils/utils.py
aggFTW: I think it should be.

remotespark/livyclientlib/exceptions.py
aggFTW: I agree a dev shouldn't. Let's document it.

remotespark/kernels/kernelmagics.py
aggFTW: Maybe add this code so that any other developer reading it doesn't wonder the same thing?

```
except:
    # Exception is already caught internally, logged, and rethrown
    pass
```

msftristew: Works for me

aggFTW:
Looks good! Did you test all magics with expected/unexpected errors on an actual kernel installation?

remotespark/kernels/kernelmagics.py
msftristew: That should be 

    except:
        # Comment
        raise

Right?

aggFTW: Right

aggFTW: Who is catching this exception?

msftristew: The frontend (the magic), using the `@handle_expected_exceptions` decorator

msftristew:
I did test it in the notebook and everything looks fine, but we should do a quick bug bash on this sometime this week to make sure we're not missing anything.

remotespark/kernels/kernelmagics.py
aggFTW: Ah, true. Sorry!

msftristew:
Addressed all feedback

remotespark/livyclientlib/exceptions.py
```diff
@@ -32,3 +46,53 @@ class SessionManagementException(LivyClientLibException):
...
+
+# == DECORATORS FOR EXCEPTION HANDLING ==
+def handle_expected_exceptions(f):
+    """A decorator that handles expected exceptions. Self can be any object with
+    an "ipython_display" attribute.
+    Usage:
+    @handle_expected_exceptions
+    def fn(self, ...):
+        etc..."""
+    exceptions_to_handle = (BadUserDataException, LivyUnexpectedStatusException, FailedToCreateSqlContextException,
```
aggFTW: I meant we should put this list at the top of the file, like a constant or something so that it's easily discoverable.

aggFTW:
Barring the very minor comment I just made, it's good to go! Merge as you see fit.

remotespark/livyclientlib/exceptions.py
msftristew: Ah, I understand. Want the decorator back in utils.py?

aggFTW: This works too. Keep it here if you want to.

msftristew:
Going to merge after tests pass

PR #217, opened 2016-04-04
Appropriately handle when user passes in invalid JSON object to %%configure

msftristew:
Display an appropriate error message when a bad input to `%%configure` is passed in.


msftristew:
Closing, can address any comments later

PR #220, opened 2016-04-06
Update setup.py to require latest version of plotly

aggFTW:


msftristew:
This took 6 commits? :) 

This is a good catch, will merge once Travis OK's

PR #221, opened 2016-04-06
Fix delete magic

aggFTW:


aggFTW:
Sorry for crazy commits. Can you squash merge when this passes? Will continue the bug bash later. Need to run out!

remotespark/kernels/kernelmagics.py
```diff
@@ -234,7 +234,7 @@ def cleanup(self, line, cell="", local_ns=None):
     @magic_arguments()
     @cell_magic
     @argument("-f", "--force", type=bool, default=False, nargs="?", const=True, help="If present, user understands.")
-    @argument("-s", "--session", type=str, help="Session id number to delete.")
+    @argument("-s", "--session", type=int, help="Session id number to delete.")
```
msftristew: Good catch -- this was a silly mistake.

tests/test_kernel_magics.py
```diff
@@ -531,7 +531,7 @@ def test_cleanup_exception():
 
 @with_setup(_setup, _teardown)
 def test_delete_without_force():
-    session_id = "0"
+    session_id = 0
```
msftristew: I hate dynamically typed languages

msftristew:
Will do

PR #222, opened 2016-04-06
Slightly better error messages for %%configure

msftristew:
Slightly better error messages for %%configure:

image

In general, if you passed bad arguments to the magics before, you got a confusing error message; these are better because they're not propagated to the user as internal exceptions.

We might want to make some further tweaks to `%%configure` messages.  That will probably require some changes to our usercodeparser.  We can address that in a later PR.


aggFTW:
Cool! LGTM!

PR #223, opened 2016-04-06
Magic API & usercodeparser tweaks

msftristew:
Change the API of the usercodeparser slightly to give us better edgecase handling and error messages.

* Check if the cell body is not empty in certain magics (`info`, `cleanup`, `help`, `logs`, `delete`) and throw an error if so

* Change the usage of the `UserCodeParser`.  Now, we only add a newline + space for some of our cell magics which are expected to have empty cell bodies.  This gives us more solid error messages if you (for example) call `%%configure -f {}` with nothing in the cell body.

image

### DIFF
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -18,7 +18,6 @@ from remotespark.livyclientlib.endpoint import Endpoint
 from remotespark.magics.sparkmagicsbase import SparkMagicBase
 from remotespark.utils.constants import LANGS_SUPPORTED
 from remotespark.utils.sparkevents import SparkEvents
-
 from remotespark.utils.utils import generate_uuid, get_livy_kind, parse_argstring_or_throw
 from remotespark.livyclientlib.exceptions import handle_expected_exceptions, wrap_unexpected_exceptions, \
     BadUserDataException
@@ -61,11 +60,13 @@ class KernelMagics(SparkMagicBase):
             spark_events = SparkEvents()
         self._spark_events = spark_events
 
+    @magic_arguments()
     @cell_magic
     @wrap_unexpected_exceptions
     @handle_expected_exceptions
     @_event
     def help(self, line, cell="", local_ns=None):
+        parse_argstring_or_throw(self.help, line)
         self._assure_cell_body_is_empty(KernelMagics.help.__name__, cell)
         help_html = """
 <table>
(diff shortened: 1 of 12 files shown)


remotespark/kernels/kernelmagics.py
```diff
@@ -18,8 +18,10 @@
 from remotespark.magics.sparkmagicsbase import SparkMagicBase
 from remotespark.utils.constants import LANGS_SUPPORTED
 from remotespark.utils.sparkevents import SparkEvents
+
```
aggFTW: Nit: extra new line

remotespark/kernels/wrapperkernel/usercodeparser.py
```diff
@@ -5,36 +5,32 @@
...
+        try:
+            all_but_first_line = code.split(None, 1)[1]
+        except IndexError:
+            all_but_first_line = ""
+
+        if code.startswith("%%local") or code.startswith("%local"):
+            return all_but_first_line
+        elif any(code.startswith("%%" + s) for s in self._magics_with_no_cell_body):
             return "{}\n ".format(code)
-        elif code.startswith(UserCodeParser.get_kernel_magic_names_as_line_magics()):
```
aggFTW: How are you doing this filter now?

```
# We could choose to transform all line magics to cell magics here since we expect that there's no line		
# magics to call, except that there are... for example %autosave and etcetera.		
# So, we'll just transform our magics.
```

remotespark/kernels/kernelmagics.py
```diff
@@ -64,6 +66,7 @@ def __init__(self, shell, data=None, spark_events=None):
     @handle_expected_exceptions
     @_event
     def help(self, line, cell="", local_ns=None):
+        self._assure_cell_body_is_empty(KernelMagics.help.__name__, cell)
```
aggFTW: What about when `line` should be empty?

msftristew: Are you saying you want me to check and assert that `line` is empty?

aggFTW: There are some magics that don't receive parameters (e.g. spark). If someone wrote:

```
%% spark a = 1
print a
```

What error will they get, and do they have any hope of fixing it after this error message?

remotespark/kernels/wrapperkernel/usercodeparser.py
msftristew: The code looks different but we're still doing the same thing.

remotespark/kernels/kernelmagics.py
msftristew: There's no error for that (this has always been the case). I can add extra assertions for that.

aggFTW: Sure, let's.

msftristew:
Now that case is fixed: 
image

remotespark/kernels/kernelmagics.py
```diff
@@ -59,11 +60,14 @@ def __init__(self, shell, data=None, spark_events=None):
             spark_events = SparkEvents()
         self._spark_events = spark_events
 
+    @magic_arguments()
     @cell_magic
     @wrap_unexpected_exceptions
     @handle_expected_exceptions
     @_event
     def help(self, line, cell="", local_ns=None):
+        parse_argstring_or_throw(self.help, line)
```
msftristew: `self.help`

aggFTW: hahahaha

PR #227, opened 2016-04-29
Expose Livy Spark UI url and driver log url in %%info magic.

tc0312:
For better debuggability.

### DIFF
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -140,22 +140,23 @@ class KernelMagics(SparkMagicBase):
     def info(self, line, cell=u"", local_ns=None):
         parse_argstring_or_throw(self.info, line)
         self._assure_cell_body_is_empty(KernelMagics.info.__name__, cell)
+        app_id = driver_log_url = spark_ui_url = None
+        if self.session_started:
+            app_id = self.spark_controller.get_app_id()
+            driver_log_url = self.spark_controller.get_driver_log_url()
+            spark_ui_url = self.spark_controller.get_spark_ui_url()
+
         self.ipython_display.writeln(u"Endpoint:\n\t{}\n".format(self.endpoint.url))
 
         self.ipython_display.writeln(u"Current session ID number:\n\t{}\n".format(
                 self.spark_controller.get_session_id_for_client(self.session_name)))
 
+        self.ipython_display.writeln(u"YARN Application ID:\n\t{}\n".format(app_id))
+
         self.ipython_display.writeln(u"Session configs:\n\t{}\n".format(conf.get_session_properties(self.language)))
 
-        if self.session_started:
-            app_id = self.spark_controller.get_app_id()
-            driver_log_url = self.spark_controller.get_driver_log_url()
-            spark_ui_url = self.spark_controller.get_spark_ui_url()
-        else:
-            app_id = driver_log_url = spark_ui_url = "Session hasn't started."
-        self.ipython_display.writeln("Driver log:\n\t{}\n".format(driver_log_url))
-        self.ipython_display.writeln("Spark UI:\n\t{}\n".format(spark_ui_url))
-        self.ipython_display.writeln("YARN Application ID:\n\t{}\n".format(app_id))
+        self.ipython_display.writeln(u"Driver log:\n\t{}\n".format(driver_log_url))
+        self.ipython_display.writeln(u"Spark UI:\n\t{}\n".format(spark_ui_url))
 
         info_sessions = self.spark_controller.get_all_sessions_endpoint_info(self.endpoint)
         self.print_endpoint_info(info_sessions)
--- a/remotespark/kernels/kernelmagics.py
+++ b/remotespark/kernels/kernelmagics.py
@@ -147,6 +147,16 @@ class KernelMagics(SparkMagicBase):
 
         self.ipython_display.writeln(u"Session configs:\n\t{}\n".format(conf.get_session_properties(self.language)))
 
+        if self.session_started:
+            app_id = self.spark_controller.get_app_id()
+            driver_log_url = self.spark_controller.get_driver_log_url()
+            spark_ui_url = self.spark_controller.get_spark_ui_url()
+        else:
+            app_id = driver_log_url = spark_ui_url = "Session hasn't started."
+        self.ipython_display.writeln("Driver log:\n\t{}\n".format(driver_log_url))
+        self.ipython_display.writeln("Spark UI:\n\t{}\n".format(spark_ui_url))
+        self.ipython_display.writeln("YARN Application ID:\n\t{}\n".format(app_id))
+
         info_sessions = self.spark_controller.get_all_sessions_endpoint_info(self.endpoint)
         self.print_endpoint_info(info_sessions)
 
--- a/remotespark/livyclientlib/livysession.py
+++ b/remotespark/livyclientlib/livysession.py
@@ -50,6 +50,7 @@ class LivySession(ObjectWithGuid):
         else:
             self.status = constants.BUSY_SESSION_STATUS
 
+        self._app_id = None
         self._logs = u""
         self._http_client = http_client
         self._status_sleep_seconds = status_sleep_seconds
@@ -106,10 +107,37 @@ class LivySession(ObjectWithGuid):
         else:
             raise FailedToCreateSqlContextException(u"Failed to create the SqlContext.\nError: '{}'".format(out))
 
+    def get_app_id(self):
+        if self._app_id is None:
+            session = self._http_client.get_session(self.id)
+            self._app_id = session['appId']
+        return self._app_id
+
+    def get_app_info(self):
+        session = self._http_client.get_session(self.id)
+        if 'appInfo' in session:
+            return session['appInfo']
+        else:
+            return {}
+
+    def get_app_info_member(self, member_name, default_value):
+        app_info = self.get_app_info()
+
+        if member_name in app_info and app_info[member_name] is not None:
+            return app_info[member_name]
+        else:
+            return default_value
+
+    def get_driver_log_url(self):
+        return self.get_app_info_member("driverLogUrl", "No driver log URL is available yet.")
+
     def get_logs(self):
         log_array = self._http_client.get_all_session_logs(self.id)[u'log']
         self._logs = "\n".join(log_array)
         return self._logs
+
+    def get_spark_ui_url(self):
+        return self.get_app_info_member("sparkUiUrl", "No Spark UI URL is available yet.")
 
     @property
     def http_client(self):
--- a/remotespark/livyclientlib/livysession.py
+++ b/remotespark/livyclientlib/livysession.py
@@ -62,7 +62,8 @@ class LivySession(ObjectWithGuid):
         self.created_sql_context = sql_created
 
     def __str__(self):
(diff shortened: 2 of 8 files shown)


remotespark/kernels/kernelmagics.py
```diff
@@ -147,6 +147,16 @@ def info(self, line, cell="", local_ns=None):
 
         self.ipython_display.writeln("Session configs:\n\t{}\n".format(conf.get_session_properties(self.language)))
 
+        if self.session_started:
+            app_id = self.spark_controller.get_app_id()
+            driver_log_url = self.spark_controller.get_driver_log_url()
+            spark_ui_url = self.spark_controller.get_spark_ui_url()
+        else:
+            app_id = driver_log_url = spark_ui_url = "Session hasn't started."
+        self.ipython_display.writeln("Driver log:\n\t{}\n".format(driver_log_url))
```
msftristew: The `\n` at the end isn't needed because this is a call to `writeln` right?

remotespark/kernels/kernelmagics.py
```diff
@@ -147,6 +147,16 @@ def info(self, line, cell="", local_ns=None):
...
+        self.ipython_display.writeln("Spark UI:\n\t{}\n".format(spark_ui_url))
+        self.ipython_display.writeln("YARN Application ID:\n\t{}\n".format(app_id))
```
msftristew: I'm unsure about the output here.  The user wants the spark UI link and yarn application id for all sessions, not just the default.

That output should become a part of `LivySession` and included in the `__str__` method for it.

remotespark/kernels/kernelmagics.py
msftristew: Driver logs shouldn't be gotten here, they should be printed as part of the `%%logs` command, or part of another command (`%%sparklogs`? `%%driverlogs`?)

msftristew:
Spoke with Alex offline, still including URL's in `%%info`, add a second `%%sparkui` magic for opening up a window to the Spark UI.

remotespark/livyclientlib/livysession.py
```diff
@@ -106,11 +107,38 @@ def create_sql_context(self):
         else:
             raise FailedToCreateSqlContextException("Failed to create the SqlContext.\nError: '{}'".format(out))
 
+    def get_app_id(self):
+        if self._app_id is None:
+            session = self._http_client.get_session(self.id)
+            self._app_id = session['appId']
```
aggFTW: Can this field be absent?

msftristew: Depends on Livy version obviously.  Not a problem for HDInsight but a problem if a person installs this repo and happens to be running against an old Livy version.  Don't know what the implications of this are, we haven't really discussed our approach to Livy versioning.

remotespark/livyclientlib/livysession.py
```diff
@@ -106,11 +107,38 @@ def create_sql_context(self):
...
+    def get_app_info_member(self, member_name, default_value):
+        app_info = self.get_app_info()
+
+        if member_name in app_info and app_info[member_name] is not None:
+            return app_info[member_name]
+        else:
+            return default_value
+
+    def get_driver_log_url(self):
+        return self.get_app_info_member("driverLogUrl", "No driver log URL is available yet.")
```
aggFTW: The message saying that the url is not available yet is misleading for versions of Livy that do not expose the URL at all...

remotespark/livyclientlib/livysession.py
```diff
@@ -106,11 +107,38 @@ def create_sql_context(self):
...
+
     def get_logs(self):
         log_array = self._http_client.get_all_session_logs(self.id)['log']
         self._logs = "\n".join(log_array)
         return self._logs
 
+    def get_spark_ui_url(self):
+        return self.get_app_info_member("sparkUiUrl", "No Spark UI URL is available yet.")
```
aggFTW: Ditto

remotespark/livyclientlib/livysession.py
aggFTW: There's no Livy versioning at this moment... What we should do is gracefully handle absences at the very least.

msftristew: I mean, we do have Livy versioning (a set of Livy versions we support and don't support), it's just not specified.

I would be comfortable with either 1) trying to do this in a backwards-compatible fashion by checking whether the record includes a session ID and 2) pinning the lowest supported version of Livy for publicpreview0.5 to whatever the current version is

remotespark/livyclientlib/livysession.py
tc0312: Changed it to show "None" like other unavailable fields.

remotespark/livyclientlib/livysession.py
tc0312: I fixed this and added unit test to make sure things are working if appId/appInfo are missing.

remotespark/livyclientlib/livysession.py
tc0312: Ditto

tc0312:
Prototyped %%sparkui. Due to its asynchronous nature, browsers built-in popup blocker blocks notebook from opening up Spark UI. Reverting to the old behavior and showing it in %%info.

For long term, we want to make it as a menu item instead of a magic command.

remotespark/magics/sparkmagicsbase.py
```diff
@@ -47,7 +47,7 @@ def _sqlquery(cell, samplemethod, maxrows, samplefraction):
 
     @staticmethod
     def print_endpoint_info(info_sessions):
-        sessions_info = ["        {}".format(i) for i in info_sessions]
+        sessions_info = ["        {}\n".format(i) for i in info_sessions]
```
aggFTW: Is the new line really necessary?

tests/test_command.py
```diff
@@ -2,10 +2,20 @@
...
 
+def _create_session(kind=SESSION_KIND_SPARK, session_id=-1,
+                    sql_created=False, http_client=None, spark_events=None):
+    if http_client is None:
+        http_client = MagicMock()
+    if spark_events is None:
+        spark_events = MagicMock()
+    ipython_display = MagicMock()
+    session = LivySession(http_client, {"kind": kind}, ipython_display, session_id, sql_created, spark_events)
+    return session
```
aggFTW: Why add it here instead of leaving the old method?

tests/test_livysession.py
```diff
@@ -8,7 +8,6 @@
 import remotespark.utils.configuration as conf
 import remotespark.utils.constants as constants
 
-
```
aggFTW: Please add all removed new lines back so that the file is PEP compliant.

tests/test_livysession.py
```diff
@@ -39,11 +36,18 @@ class TestLivySession(object):
     log_json = json.loads('{"id":6,"from":0,"total":212,"log":["hi","hi"]}')
 
     def __init__(self):
+        self.http_client = None
```
aggFTW: Why make both http_client and spark_events class instances? I really dislike coupling unit tests like this unless there's an extremely good reason.

tests/test_livysession.py
```diff
@@ -64,15 +68,15 @@ def _next_session_response_post(self, *args):
...
-            http_client = MagicMock()
-        if spark_events is None:
-            spark_events = MagicMock()
+    def _create_session(self, kind=constants.SESSION_KIND_SPARK, session_id=-1, sql_created=False):
         ipython_display = MagicMock()
-        session = LivySession(http_client, {"kind": kind}, ipython_display, session_id, sql_created, spark_events)
+        session = LivySession(self.http_client, {"kind": kind}, ipython_display, session_id, sql_created, self.spark_events)
+        return session
+
+    def _create_session_with_fixed_get_response(self, get_session_json):
```
aggFTW: All of these helper methods should be `staticmethod`s to ensure needless coupling.

aggFTW:
Could you post a screenshot of formatting for info when notebook has no session vs when it does so that we have an idea/can comment?

remotespark/magics/sparkmagicsbase.py
tc0312: I splitted Spark UI url and driver log url into its own line. Adding a new line between each session make separation between each session clearer.

tests/test_livysession.py
tc0312: http_client and spark_events used by almost all test cases. Moving them to test fixture makes sense.
They are recreated before each test case is executed. There's no coupling between test cases.

remotespark/magics/sparkmagicsbase.py
aggFTW: It seems we need to reformat to an HTML table or similar, as this has gotten messy already? What do you think?

tests/test_livysession.py
tc0312: They rely on test fixture. The coupling is not needless.

tests/test_livysession.py
aggFTW: How would you parallelize these tests with test fixtures like this?

remotespark/magics/sparkmagicsbase.py
tc0312: I agree but I want to get this out first and do it in another commit. Will create an GH issue for it.

aggFTW: Sounds good!

tests/test_livysession.py
tc0312: By creating test fixture before executing each test case.
Setting _multiprocess_can_split_ = True in the test class will do the trick.

tests/test_command.py
tc0312: I admit that pasting the old method here isn't the cleanest way. But before my change, it called a member method in another test class, which already looks odd. Ideally we should extract the common method into a base test suite/a test helper class.

tc0312:
image

tests/test_livysession.py
aggFTW: Got it. LGTM.

tests/test_command.py
aggFTW: True. Do you want to do that?

aggFTW:
We talked f2f.

Remaining items seem to be:
1. Encoding password to be base64 in config file.
2. Making info output an html table. You might want to use: https://ipython.org/ipython-doc/3/config/integrating.html

tc0312:
Will file another PR very soon for 1.
Filed an issue to track [2](https://github.com/jupyter-incubator/sparkmagic/issues/230)

msftristew:
Please resolve conflicts

remotespark/livyclientlib/livysession.py
```diff
@@ -61,7 +62,8 @@ def __init__(self, http_client, properties, ipython_display,
         self.created_sql_context = sql_created
 
     def __str__(self):
-        return "Session id: {}\tKind: {}\tState: {}".format(self.id, self.kind, self.status)
+        return "Session id: {}\tYARN id: {}\tKind: {}\tState: {}\n\tSpark UI: {}\n\tDriver Log: {}"\
```
msftristew: This is fine for now but let's close #230 soon.  This is obviously getting unwieldy.

msftristew:
LGTM

tc0312:
Resolved conflicts.

msftristew:
Feel free to merge if it looks good to you, Alejandro

PR #231, opened 2016-05-03
Added support for storing endpoint password in base64 encoding in con…

tc0312:
…figuration.

If 'base64_password' is present in config, it's used. Else, it will fallback to 'password'.

### DIFF
--- a/remotespark/livyclientlib/exceptions.py
+++ b/remotespark/livyclientlib/exceptions.py
@@ -48,6 +48,11 @@ class BadUserDataException(LivyClientLibException):
     in some way."""
 
 
+class BadUserConfigurationException(LivyClientLibException):
+    """An exception that is thrown when configuration provided by the user is invalid
+    in some way."""
+
+
 # == DECORATORS FOR EXCEPTION HANDLING ==
 EXPECTED_EXCEPTIONS = [BadUserDataException, LivyUnexpectedStatusException, FailedToCreateSqlContextException,
                        HttpClientException, LivyClientTimeoutException, SessionManagementException]
--- a/remotespark/livyclientlib/exceptions.py
+++ b/remotespark/livyclientlib/exceptions.py
@@ -43,18 +43,18 @@ class SessionManagementException(LivyClientLibException):
     given session name is invalid in some way."""
 
 
-class BadUserDataException(LivyClientLibException):
-    """An exception that is thrown when data provided by the user is invalid
-    in some way."""
-
-
 class BadUserConfigurationException(LivyClientLibException):
     """An exception that is thrown when configuration provided by the user is invalid
     in some way."""
 
 
+class BadUserDataException(LivyClientLibException):
+    """An exception that is thrown when data provided by the user is invalid
+    in some way."""
+
+
 # == DECORATORS FOR EXCEPTION HANDLING ==
-EXPECTED_EXCEPTIONS = [BadUserDataException, LivyUnexpectedStatusException, FailedToCreateSqlContextException,
+EXPECTED_EXCEPTIONS = [BadUserConfigurationException, BadUserDataException, LivyUnexpectedStatusException, FailedToCreateSqlContextException,
                        HttpClientException, LivyClientTimeoutException, SessionManagementException]
 
 
--- a/remotespark/utils/configuration.py
+++ b/remotespark/utils/configuration.py
@@ -3,8 +3,10 @@
 # Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
 
-import json
+import base64
 import copy
+import json
+import sys
 
 from remotespark.utils.constants import CONFIG_JSON
 from remotespark.utils.utils import join_paths, get_magics_home_path, get_livy_kind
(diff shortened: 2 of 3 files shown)


remotespark/utils/configuration.py
```diff
@@ -89,14 +91,38 @@ def session_configs():
...
+        credentials = _override(f)()
+        base64_decoded_credentials = {k: credentials.get(k) for k in ('username', 'password', 'url')}
+        base64_password = credentials.get('base64_password')
+        if base64_password is not None:
+            try:
+                base64_decoded_credentials['password'] = base64.b64decode(base64_password).decode()
+            except Exception:
+                exception_type, exception, traceback = sys.exc_info()
+                msg = "base64_password for %s contains invalid base64 string: %s %s" % (f.__name__, exception_type, exception)
+                raise RuntimeError(msg)
```
aggFTW: Please look at how we do error handling. Only Exceptions of a certain type are handled and bubbled up to users. Look at `handle_expected_exceptions` inside `utils\exceptions.py`

I would say misconfiguration is an expected exception (not a bug) and you should create an exception for this type of error.

remotespark/utils/configuration.py
```diff
@@ -89,14 +91,38 @@ def session_configs():
...
+    def ret():
+        credentials = _override(f)()
+        base64_decoded_credentials = {k: credentials.get(k) for k in ('username', 'password', 'url')}
+        base64_password = credentials.get('base64_password')
+        if base64_password is not None:
+            try:
+                base64_decoded_credentials['password'] = base64.b64decode(base64_password).decode()
+            except Exception:
+                exception_type, exception, traceback = sys.exc_info()
+                msg = "base64_password for %s contains invalid base64 string: %s %s" % (f.__name__, exception_type, exception)
```
aggFTW: Please remove a potential password from the message, as it might get logged somewhere.

remotespark/utils/configuration.py
```diff
@@ -89,14 +91,38 @@ def session_configs():
...
+    # Hack! We do this so that we can query the .__name__ of the function
+    # later to get the name of the configuration dynamically, e.g. for unit tests
+    ret.__name__ = f.__name__
+    return ret
+
+
+@_credentials_override
 def kernel_python_credentials():
-    return {'username': '', 'password': '', 'url': 'http://localhost:8998'}
+    return {'username': '', 'base64_password': '', 'url': 'http://localhost:8998'}
```
aggFTW: Shouldn't the default be returning a `'password'` field too?

remotespark/utils/configuration.py
```diff
@@ -89,14 +91,38 @@ def session_configs():
...
 
 
-@_override
+@_credentials_override
 def kernel_scala_credentials():
-    return {'username': '', 'password': '', 'url': 'http://localhost:8998'}
+    return {'username': '', 'base64_password': '', 'url': 'http://localhost:8998'}
```
aggFTW: Ditto

remotespark/utils/configuration.py
```diff
@@ -89,14 +91,38 @@ def session_configs():
...
 
-@_override
+def _credentials_override(f):
+    """A decorator that provide special handling for credentials. It still calls _override().
+    If 'base64_password' in config is set, it will base64 decode it and returned in return value's 'password' field.
+    If 'base64_password' is not set, it will fallback to to 'password' in config.
+    """
+    def ret():
+        credentials = _override(f)()
+        base64_decoded_credentials = {k: credentials.get(k) for k in ('username', 'password', 'url')}
```
aggFTW: Doesn't `get` throw if there's no key? If that's the case, then the possible exception has to be wrapped in an expected exception so that the problem is properly displayed to the user.

tc0312: No, get returns None or whatever default value you provide.

remotespark/utils/configuration.py
tc0312: Sorry where's the potential password?

remotespark/utils/configuration.py
tc0312: No, 'password' is deprecated. User should use base64_password. If base64_password is not specified, it will fallback to 'password'.

remotespark/utils/configuration.py
aggFTW: Got it, my bad.

remotespark/utils/configuration.py
aggFTW: Please ignore, I misread.

aggFTW:
Cool. So only thing would be to create an exception for bad configuration instead of raising a `RuntimeError`

msftristew:
Why does this PR exist?

msftristew:
Please resolve conflicts

remotespark/utils/configuration.py
msftristew: This would likely work as a `BadUserDataException`.

tc0312:
Resolved conflicts.

remotespark/livyclientlib/exceptions.py
```diff
@@ -48,6 +48,11 @@ class BadUserDataException(LivyClientLibException):
     in some way."""
 
 
+class BadUserConfigurationException(LivyClientLibException):
+    """An exception that is thrown when configuration provided by the user is invalid
+    in some way."""
+
+
 # == DECORATORS FOR EXCEPTION HANDLING ==
 EXPECTED_EXCEPTIONS = [BadUserDataException, LivyUnexpectedStatusException, FailedToCreateSqlContextException,
```
aggFTW: @msftristew 
Do we need to have BadUserConfigurationException be part of `EXPECTED_EXCEPTIONS` so that it's properly handled?

msftristew: Yes.

msftristew:
LGTM once you verify the following:

1. Add `BadUserConfigurationException` to `EXPECTED_EXCEPTIONS`

2. Verify the error message is sensible when someone passes in an invalid base64 string (should be the case if you satisfy (1))

3. Verify that you can still use an empty password to turn off HTTP authentication

tc0312:
Verified 2 manually.

aggFTW:
LGTM! @msftristew, please merge when you take a final look.

Issue #228, opened 2016-05-02
Cannot use non-ASCII characters in notebooks


msftristew:
This is as far as I can tell an issue with `UserCodeParser`.

msftristew:
File "c:\users\ristew\documents\github\sparkmagic\remotespark\livyclientlib\exceptions.py", line 92, in wrapped
    out = f(self, *args, **kwargs)
  File "c:\users\ristew\documents\github\sparkmagic\remotespark\kernels\wrapperkernel\sparkkernelbase.py", line 51, in f
    return self._do_execute(code, silent, store_history, user_expressions, allow_stdin)
  File "c:\users\ristew\documents\github\sparkmagic\remotespark\kernels\wrapperkernel\sparkkernelbase.py", line 61, in _do_execute
    code_to_run = self.user_code_parser.get_code_to_run(code)
  File "c:\users\ristew\documents\github\sparkmagic\remotespark\kernels\wrapperkernel\usercodeparser.py", line 36, in get_code_to_run
    return "%%spark\n{}".format(code)

PR #232, opened 2016-05-10
Add Unicode support

aggFTW:
As reported on http://disq.us/9ntn8x

### DIFF
--- a/remotespark/utils/configuration.py
+++ b/remotespark/utils/configuration.py
@@ -190,5 +190,6 @@ def max_slices_pie_graph():
     return 100
 
 
+@_override
 def pyspark_sql_encoding():
     return u'utf-8'
--- a/remotespark/utils/configuration.py
+++ b/remotespark/utils/configuration.py
@@ -32,8 +32,8 @@ def load(fsrw_class=None):
     config_file = fsrw_class(join_paths(home_path.path, CONFIG_JSON))
     config_file.ensure_file_exists()
     config_text = config_file.read_lines()
-    line = "".join(config_text).strip()
-    if line == "":
+    line = u"".join(config_text).strip()
+    if line == u"":
         overrides = {}
     else:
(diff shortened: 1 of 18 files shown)


aggFTW:
Closes #228

remotespark/utils/configuration.py
```diff
@@ -188,3 +188,7 @@ def default_samplefraction():
 @_override
 def max_slices_pie_graph():
     return 100
+
+
+def pyspark_sql_encoding():
```
msftristew: Missing `_override` decorator?

msftristew:
Does Scala work?

aggFTW:
Yes

msftristew:
LGTM. We should probably use Unicode for all strings as we continue working on the codebase.

guybartal:
Hi
We have the same problem with Hebrew characters in UTF-8, when are you going to publish the fix? 
it is very important for my customer's POC....

Best Regards,
Guy Bertental

msftristew:
Hi Guy, 

This issue was resolved back on May 9.  The most current version of the code should have this fix (including the currently live version of the code on HDInsight).  Are you running the latest version?

guybartal:
Hi @msftristew 
We have created the cluster today on West Europe region so I guess we do have latest version,
you can see my problem here:
[http://stackoverflow.com/questions/37698276/utf-8-text-in-hdinsight-cluster-with-spark-result-encoding-error-ascii-codec-c](url)

msftristew:
Thanks Guy.  The issue you're describing is distinct from this issue -- this issue prevented people from executing code containing Unicode characters in notebooks.  Your issue is an unrelated string encoding error which doesn't seem to have anything to do with the Sparkmagic project itself.  We'll follow up directly at Stack Overflow.

guybartal:
thanks @msftristew hoping to find solution soon....

PR #236, opened 2016-05-17
Downgraded plotly to 1.9.4.

tc0312:


tc0312:
We have a hotfix for the Plotly require fix. Discarding this.

PR #238, opened 2016-05-17
Fix double quotes in sql queries for Python and adds success status

aggFTW:


remotespark/livyclientlib/sqlquery.py
```diff
@@ -80,7 +80,7 @@ def _records_to_dataframe(records_text):
             raise DataFrameParseException(u"Cannot parse object as JSON: '{}'".format(strings))
 
     def _pyspark_command(self):
-        command = u'sqlContext.sql(u"""{}""").toJSON()'.format(self.query)
+        command = u'sqlContext.sql(u"""{} """[:-1]).toJSON()'.format(self.query)
```
msftristew: `[:-1]` is not necessary

remotespark/utils/constants.py
```diff
@@ -60,15 +60,16 @@
 NOT_STARTED_SESSION_STATUS = "not_started"
 STARTING_SESSION_STATUS = "starting"
 BUSY_SESSION_STATUS = "busy"
+SUCCESS_SESSION_STATUS = "success"
```
msftristew: Why?

msftristew: I've spoken to @tc0312 about this and he seemed to indicate that it is a bug if interactive Livy sessions over go to success state

aggFTW: Talked to him just now and he says a session should not be but might be momentarily in success state. He said we should add it and treat it as a final status.

PR #241, opened 2016-05-25
Changed %%info to not show the full url in the table.

tc0312:
Also changed it to not show the table if there's no active sessions.

image


aggFTW:
Perfect

PR #244, opened 2016-06-03
Add example to autoviz

aggFTW:


autovizwidget/examples/autoviz.ipynb
```diff
@@ -0,0 +1,688 @@
...
+    "![widget2](Capture0.PNG)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Convenience methods\n",
+    "\n",
+    "Alternatively, you can let the library inspect your pandas DataFrame and select the X and Y axis automatically for you.\n",
```
msftristew: Something to consider (this is a nitpick) -- I might rather present `display_dataframe` as the default, standard way to display a dataframe, and then offer the `Encoding` method as a more low-level way to do this if you need more control.  I'm fine with either though.

msftristew: My reasoning: this is the `autovizwidget` library, so I think I would expect something to do autoviz without me having to do anything else :)

aggFTW: Done :)

aggFTW:
Merging. I can address other feedback later.

PR #249, opened 2016-06-10
User error on sql query should not be DataFrameParseException

aggFTW:


msftristew:
merge after travis passes

PR #258, opened 2016-06-16
Respect column order for JSON queries. Make sure all null columns sho…

### LINKED ISSUE #256: Sql queries do not respect column orders

aggFTW:
…w up in pandas df

Closes #256 and #257.


msftristew:
lgtm

Issue #255, opened 2016-06-14
%%info does not return because Jupyter is not responsive


aggFTW:
Seems to me we should print this information as soon as context has been created.

Issue #254, opened 2016-06-14
User might think that HiveContext creation is taking longer than it is


aggFTW:
Last cell output says "Creating HiveContext as 'sqlContext'" but does not tell user when context has been created and code is now running.

linchan-ms:
#259

PR #259, opened 2016-06-17
Print the link to Spark UI and driver log as soon as SparkContext is …

linchan-ms:
…created successfully.

### DIFF
--- a/sparkmagic/sparkmagic/livyclientlib/livysession.py
+++ b/sparkmagic/sparkmagic/livyclientlib/livysession.py
@@ -288,3 +288,17 @@ class LivySession(ObjectWithGuid):
         if self._heartbeat_thread is not None:
             self._heartbeat_thread.stop()
             self._heartbeat_thread = None
+
+    def get_row_html(self, current_session_id):
+        return u"""<tr><td>{0}</td><td>{1}</td><td>{2}</td><td>{3}</td><td>{4}</td><td>{5}</td><td>{6}</td></tr>""".format(
+            self.id, self.get_app_id(), self.kind, self.status,
+            self.get_html_link(u'Link', self.get_spark_ui_url()), self.get_html_link(u'Link', self.get_driver_log_url()),
+            u"" if current_session_id is None or current_session_id != self.id else u"\u2714"
+        )
+
+    @staticmethod
+    def get_html_link(text, url):
+        if url is not None:
+            return u"""<a target="_blank" href="{1}">{0}</a>""".format(text, url)
+        else:
+            return u""
--- a/sparkmagic/sparkmagic/livyclientlib/livysession.py
+++ b/sparkmagic/sparkmagic/livyclientlib/livysession.py
@@ -8,7 +8,7 @@ import sparkmagic.utils.configuration as conf
 import sparkmagic.utils.constants as constants
 from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.utils.sparkevents import SparkEvents
-from sparkmagic.utils.utils import get_session_info_html
+from sparkmagic.utils.utils import get_sessions_info_html
 from .command import Command
 from .exceptions import FailedToCreateSqlContextException, LivyClientTimeoutException, \
     LivyUnexpectedStatusException, BadUserDataException
@@ -129,7 +129,7 @@ class LivySession(ObjectWithGuid):
                 raise LivyClientTimeoutException(u"Session {} did not start up in {} seconds."
                                                  .format(self.id, conf.livy_session_startup_timeout_seconds()))
 
-            html = get_session_info_html([self], self.id)
+            html = get_sessions_info_html([self], self.id)
             self.ipython_display.html(html)
 
             if create_sql_context:
--- a/sparkmagic/sparkmagic/livyclientlib/livysession.py
+++ b/sparkmagic/sparkmagic/livyclientlib/livysession.py
@@ -8,6 +8,7 @@ import sparkmagic.utils.configuration as conf
 import sparkmagic.utils.constants as constants
 from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.utils.sparkevents import SparkEvents
+from sparkmagic.utils.utils import get_session_info_html
 from .command import Command
 from .exceptions import FailedToCreateSqlContextException, LivyClientTimeoutException, \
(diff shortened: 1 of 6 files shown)


sparkmagic/sparkmagic/utils/utils.py
```diff
@@ -50,3 +50,24 @@ def coerce_pandas_df_to_numeric_datetime(df):
                 coerced = True
             except (ValueError, TypeError):
                 pass
+
+def get_session_info_html(info_sessions, current_session_id):
```
aggFTW: This receives a list of sessions. Rename to `get_sessions_info`?

MohamedElKamhawy:
Travis Build fails, please check.

aggFTW:
Can you please fix the unit tests and add new unit tests? Running unit test instructions is in README.md

aggFTW:
Also, it'd be good to get a screenshot of how creating a context now looks like and how calling `%%info` now looks like.

linchan-ms:
sessionstart
Sample output

aggFTW:
Closes #255

sparkmagic/sparkmagic/tests/test_livysession.py
```diff
@@ -376,6 +376,9 @@ def test_logs_gets_latest_logs(self):
     def test_wait_for_idle_returns_when_in_state(self):
         self.http_client.post_session.return_value = self.session_create_json
         self.get_session_responses = [self.ready_sessions_json,
+                                      self.ready_sessions_json,
```
aggFTW: I don't think that all these unit tests should be actually using the http_client when retrieving the html table for the session. Maybe make the `get_session_row_html` part of session and simply use a `MagicMock` in every unit test and just verify that it's called?

sparkmagic/sparkmagic/utils/utils.py
```diff
@@ -50,3 +50,24 @@ def coerce_pandas_df_to_numeric_datetime(df):
...
+
+def get_sessions_info_html(info_sessions, current_session_id):
+    html = u"""<table>
+<tr><th>ID</th><th>YARN Application ID</th><th>Kind</th><th>State</th><th>Spark UI</th><th>Driver log</th><th>Current session?</th></tr>""" + \
+    u"".join([get_session_row_html(session, current_session_id) for session in info_sessions]) + \
+    u"</table>"
+
+    return html
+
+def get_session_row_html(session, current_session_id):
```
aggFTW: So you'd take this method out of this module and leave the other two here. Then just call `session.html_row(current_session_id)` from `get_sessions_info_html`. That way, you can substitute the `html_row` method from any session with a `MagicMock` without actually using the http client.

aggFTW:
Can you also add a screenshot of calling %%info?
Thanks!

msftristew:
Do you really want to print all that every single time a session is created?  That's not an amazing UI IMO.

aggFTW:
This is a solution to #255 and closes #254. Do you have another suggestion?

msftristew:
I don't really see a need to dump the entire `%%info` table every time a session is created -- that's a lot of visual noise for something that should ostensibly be very simple (like typing the number `1`). 

If you want to provide the Spark UI link and driver log link, I'd say you can just add the links in the output text:

    Creating SparkContext as sc
    Click here for Spark UI
    Click here for driver log
    Creating HiveContext as sqlContext
    ...

Maybe you all disagree though.

linchan-ms:
@aggFTW  %%info output doesn't change at all. Do you still need a screen shot?

linchan-ms:
@msftristew, This is just using the same format but it always only dumps the info for current session. I do believe the application ID is useful for customer as well. If we really want to make it simpler, we can remove the state and current session column. But I'd rather have them printed so we can use the same function as for %%info. That will be less testing cost and less likely to break later if something changes (especially the URL rewrite logic).

@MohamedElKamhawy, Agree this is not amazing UI. We don't have amazing UI for this yet. This is just for now to provide the utility.

msftristew:
I'm the one who said the UI was not amazing, not @MohamedElKamhawy  :) 

Will defer to your judgment -- however, I'll point out that taking 3 separate lines to print the application ID and the 2 links will take up less vertical and horizontal space than printing out the entire table.

aggFTW:
I don't see anything wrong with displaying the table.

linchan-ms:
@aggFTW, I looked at the mock a bit, adding those ready_sessions_json is essentially creating a mock behavior which doesn't affect the test scenario. It is just mocking at a different place. I don't see any significant benefit of moving the functions inside LivySession and then mock it.

aggFTW:
Yes, except it's much easier to understand than understanding how many mock requests you need and which reply corresponds to which. That's a significant benefit.

linchan-ms:
Updated. Please take a look again.

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -283,3 +288,17 @@ def _stop_heartbeat_thread(self):
...
             self._heartbeat_thread = None
+
+    def get_row_html(self, current_session_id):
+        return u"""<tr><td>{0}</td><td>{1}</td><td>{2}</td><td>{3}</td><td>{4}</td><td>{5}</td><td>{6}</td></tr>""".format(
+            self.id, self.get_app_id(), self.kind, self.status,
+            self.get_html_link(u'Link', self.get_spark_ui_url()), self.get_html_link(u'Link', self.get_driver_log_url()),
+            u"" if current_session_id is None or current_session_id != self.id else u"\u2714"
+        )
+
+    @staticmethod
```
aggFTW: This should be in `utils`

sparkmagic/sparkmagic/tests/test_livysession.py
```diff
@@ -418,10 +422,10 @@ def test_wait_for_idle_throws_when_in_final_status(self):
     @raises(LivyClientTimeoutException)
     def test_wait_for_idle_times_out(self):
         self.http_client.post_session.return_value = self.session_create_json
-        self.get_session_responses = [self.ready_sessions_json,
-                                      self.busy_sessions_json,
-                                      self.busy_sessions_json,
-                                      self.ready_sessions_json]
+        self.get_session_responses = [ self.ready_sessions_json,
```
aggFTW: Nit: remove extra whitespace?

aggFTW:
LGTM!

PR #261, opened 2016-06-20
Print the payload of the HTTP response when the status code doesn't m…

linchan-ms:
…atch the expected.

### DIFF
--- a/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
+++ b/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
@@ -76,7 +76,10 @@ class ReliableHttpClient(object):
                     sleep(self._retry_policy.seconds_to_sleep(retry_count))
                     retry_count += 1
                     continue
+
+                if error:
+                    raise HttpClientException(u"Error sending http request and maximum retry encountered.")
                 else:
-                    raise HttpClientException(u"Invalid status code '{}' or error '{}' from {} with text {}"
-                                              .format(status, error, url, text))
+                    raise HttpClientException(u"Invalid status code '{}' from {} with error payload: {}"
+                                              .format(status, url, text))
             return r
--- a/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
+++ b/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
@@ -1,9 +1,7 @@
 ﻿# Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
-
 import json
 from time import sleep
-
 import requests
 
 import sparkmagic.utils.configuration as conf
--- a/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
+++ b/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
@@ -11,6 +11,7 @@ from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.utils.constants import MAGICS_LOGGER_NAME
 from sparkmagic.livyclientlib.exceptions import HttpClientException
 
+
 class ReliableHttpClient(object):
     """Http client that is reliable in its requests. Uses requests library."""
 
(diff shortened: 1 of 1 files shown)


msftristew:
Can we also fix the "or error 'False'" thing here?  That has always bothered me about the text of this error message.

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -62,18 +63,20 @@ def _send_request_helper(self, url, accepted_status_codes, function, data, retry
...
 
             if error or status not in accepted_status_codes:
                 if self._retry_policy.should_retry(status, error, retry_count):
                     sleep(self._retry_policy.seconds_to_sleep(retry_count))
                     retry_count += 1
                     continue
                 else:
-                    raise HttpClientException(u"Invalid status code '{}' or error '{}' from {}"
-                                              .format(status, error, url))
+                    raise HttpClientException(u"Invalid status code '{}' or error '{}' from {} with text {}"
```
aggFTW: >with text

Somehow, I don't think that's clear enough. Maybe make it "with payload"?

Can you show an example of how r.text looks like for these type of exceptions?

linchan-ms:
parameter error
Changed the text to error payload. Basically we will just print whatever is passed back from REST API.

msftristew:
1. The error message will be incomprehensible if an exception is thrown by the requests library -- this should be fixed in this PR

2. Please remove "or error 'False'" from the error message

3. Please format the output sensibly: the output is full of backslashes and `\n`'s

linchan-ms:
@msftristew 
1. What do you mean by requests library? 
2. Why is the error False check there in the first place? Seems we rely on that to do retry and a few tests actually rely on that to be true? You mean we remove the error check completely from the underlying retry logic as well?
3. I believe those back slashes come from the REST API itself and magic shouldn't do parsing at this point. This is just one particular http request, and the handling is common for all calls. I don't believe we will be able to parse every http request results. Currently from Livy side, the error purely comes from the JSON library, so I don't think we have any quick fix around this yet.

msftristew:
1. We're using the `requests` Python library to make HTTP requests

2. No -- looking at the code, it seems we simply use the "error" variable to record if an exception was thrown by the `requests` library.  There should actually be 2 distinct error cases with 2 different error messages: one if an exception is thrown, and the other if no exception is thrown but Livy gives us a bad error code.  The point I am trying to make is that if `requests` throws an exception, the resultant error message will be incomprehensible.  (To see for yourself, add the line `raise requests.exceptions.RequestException()` at line 50 and see what happens :) )

3. I'm not asking you to parse the output, I'm asking you to format it so that it's readable.

msftristew:
Spoke with @linchan-ms offline, looks like he's going to push a small update to fix up our error messages.

Might be somewhat complicated to fix the string formatting issue from this PR, I'm OK with it looking like this for now

This class + LinearRetryPolicy are structured a little funnily, might want to look at refactoring in a later PR

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -1,15 +1,16 @@
 ﻿# Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
+
```
aggFTW: Can you get rid of these two extra lines in the imports please?

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -1,15 +1,16 @@
...
 from time import sleep
+
 import requests
 
 import sparkmagic.utils.configuration as conf
 from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.utils.constants import MAGICS_LOGGER_NAME
 from sparkmagic.livyclientlib.exceptions import HttpClientException
 
-
```
aggFTW: Add this line back please?

aggFTW:
LGTM

PR #262, opened 2016-06-21
Add ability to choose a y axis with pie chart

aggFTW:
We've learned from users that when doing a query like:

```
SELECT label, COUNT(*) FROM table
```

they expect the pie chart to show proportional areas by the count. Seeing that all labels will get the same amount of area, which is the current behavior, is counterintuitive for them.

That said, previous functionality can still be achieved by selecting None as the Y encoding.

### DIFF
--- a/autovizwidget/autovizwidget/plotlygraphs/piegraph.py
+++ b/autovizwidget/autovizwidget/plotlygraphs/piegraph.py
@@ -6,9 +6,10 @@ from plotly.offline import iplot
 from pandas.core.groupby import DataError
 
 import autovizwidget.utils.configuration as conf
+from .graphbase import GraphBase
 
 
-class PieGraph(object):
+class PieGraph(GraphBase):
     @staticmethod
     def render(df, encoding, output):
         if encoding.x is None:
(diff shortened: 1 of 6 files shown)


aggFTW:
Selecting a Y axis:

y_selected

Previous functionality:

y_not_selected

msftristew:
My only worry is that I don't know if anyone will be able to come up with the "secret recipe" for producing the second pie graph.  FWIW, even if I knew I wanted to construct exactly that pie graph, I would never be able to guess that the way to get it would be by selecting X: date, Y: null, Func: Sum.

aggFTW:
1. Any Func works to get that graph, not only Sum.
2. Your discoverability argument applies to any other graph really, no? I have a hard time creating charts in any other number of products I've used that make charting easy. Do you have a particular suggestion on how to make it better?

msftristew:
No particular suggestion for improvement.  To me, if I were going to try to make this graph, I would choose:

X: date
Y: date
Func: count

However, the implementation to make that work probably wouldn't be simple.

LGTM, feel free to merge as-is, but I think this question is worth investigating further.  (Possible approach: poll some users about how they expect the UI to work, find the most common response(s), and ensure that set of options works the way people expect it to)

PR #272, opened 2016-08-31
Add ability to change endpoint before a session is started

aggFTW:
Magic looks like:

`%_do_not_call_change_endpoint -u admin -p password -s https://cluster.com`


tc0312:
LGTM

serverextension/sparkmagic_ext/handlers.py
```diff
@@ -0,0 +1,58 @@
+import json
+from notebook.utils import url_path_join
+from notebook.base.handlers import IPythonHandler
+
+
+class ReconnectHandler(IPythonHandler):
+    def get(self):
+        path = 'Temp/Untitled3.ipynb'
```
linchan-ms: Why hardcoded path?

aggFTW: This file is not part of this PR.

Issue #171, opened 2016-02-23
Add configuration to turn off HiveContext, use plain SqlContext instead


msftristew:
See title. User should be able to flip a configuration to use a plain SqlContext instead of a full HiveContext.  The context should have the same name (sqlContext) regardless of which kind of context you're using, so the switch would be transparent for the user.

tc0312:
Related change in Livy: [LIVY-94. SQLContext/HiveContext is not available in Interpreter](cloudera/livy/pull/132)

msftristew:
Is my understanding correct that this change automatically creates a SQLContext in Livy interactive sessions and that Livy itself would have a configuration to choose between HiveContext and SqlContext?

linchan-ms:
No. Actually it completely skips the creation of sqlContext variable. That responsibility is completely shifted to Livy. Livy has the config to say you want only SQLContext or you want HiveContext. I guess you could say that with livy 0.3 and this, this issue is resolved.

aggFTW:
Solved by Livy.

PR #277, opened 2016-09-29
Livy0.3compat

linchan-ms:
Changes needed to support livy 0.3


msftristew:
Looks like this fixes #171 ?

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -59,6 +59,12 @@ def __init__(self, http_client, properties, ipython_display,
         super(LivySession, self).__init__()
         assert u"kind" in list(properties.keys())
         kind = properties[u"kind"]
+
+        if not should_heartbeat:
+            properties.pop(u"heartbeatTimeoutInSecond")
```
aggFTW: Add to constants file

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -59,6 +59,12 @@ def __init__(self, http_client, properties, ipython_display,
...
+        elif properties[u"heartbeatTimeoutInSecond"] == 0:
```
aggFTW: Add to constants file

sparkmagic/example_config.json
```diff
@@ -54,5 +54,8 @@
   
   "should_heartbeat": false,
   "heartbeat_refresh_seconds": 30,
-  "heartbeat_retry_seconds": 10
+  "heartbeat_timeout_seconds": 60,
```
aggFTW: livy_server_heartbeat_timeout_seconds

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: pop will raise KeyError if not in the dictionary and a default is not given. Please add a default.

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -59,6 +59,12 @@ def __init__(self, http_client, properties, ipython_display,
...
+            properties.pop(u"heartbeatTimeoutInSecond")
```
aggFTW: pop will raise KeyError if not in the dictionary and a default is not given. Please add a default.

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: Shouldn't this be `<= 0`?

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: We need an `if should_hearbeat:` that adds the property if not present.

sparkmagic/sparkmagic/tests/test_livysession.py
```diff
@@ -140,6 +140,7 @@ def test_constructor_starts_with_existing_session(self):
         assert session.id == session_id
         assert session.created_sql_context
         assert session._heartbeat_thread is None
+        assert u"heartbeatTimeoutInSecond" not in list(session.properties.keys())
```
aggFTW: Someone could break this test unknowingly if `"heartbeatTimeoutInSecond": heartbeat_timeout` was removed from the `_create_session` method.

sparkmagic/sparkmagic/tests/test_livysession.py
```diff
@@ -72,10 +72,10 @@ def _next_session_response_post(self, *args):
         return val
 
     def _create_session(self, kind=constants.SESSION_KIND_SPARK, session_id=-1, sql_created=False,
-                        should_heartbeat=False):
+                        should_heartbeat=False, heartbeat_timeout=60):
         ipython_display = MagicMock()
         session = LivySession(self.http_client,
-                              {"kind": kind},
+                              {"kind": kind, "heartbeatTimeoutInSecond": heartbeat_timeout},
```
aggFTW: What about adding the key to the properties only if `should_heartbeat` is `True`?

sparkmagic/sparkmagic/tests/test_livysession.py
```diff
@@ -150,22 +151,23 @@ def test_constructor_starts_heartbeat_with_existing_session(self):
         })
         session_id = 1
         session = self._create_session(session_id=session_id, sql_created=True,
-                                       should_heartbeat=True)
+                                       should_heartbeat=True, heartbeat_timeout=0)
```
aggFTW: This test is there to make sure that an existing session starts heartbeat correctly if asked to. Your modification breaks that intention. Please add a new test that reflects what you are testing for here.

sparkmagic/sparkmagic/tests/test_livysession.py
```diff
@@ -150,22 +151,23 @@ def test_constructor_starts_heartbeat_with_existing_session(self):
...
+        assert u"heartbeatTimeoutInSecond" not in list(session.properties.keys())
         
     def test_start_with_heartbeat(self):
         self.http_client.post_session.return_value = self.session_create_json
         self.http_client.get_session.return_value = self.ready_sessions_json
 
         conf.override_all({
             "status_sleep_seconds": 0.01,
-            "statement_sleep_seconds": 0.01
+            "statement_sleep_seconds": 0.01,
```
aggFTW: Remove the comma?

sparkmagic/sparkmagic/tests/test_livysession.py
aggFTW: The behavior should be that the key is added/removed based on what configuration.json dictates. You should be overriding conf (via `conf.override_all()`) instead of manually adding things here.

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
```diff
@@ -80,7 +80,7 @@ def test_add_sessions_command_parses():
     magic.spark(line)
 
     add_sessions_mock.assert_called_once_with("name", Endpoint("http://url.com", "sdf", "w"),
-                                              False, {"kind": "pyspark"})
+                                              False, {"kind": "pyspark", "heartbeatTimeoutInSecond": 60})
```
aggFTW: Use the constant

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: What do you mean by that? Why the "kind" is not in constants file?
In real code, it is guaranteed to be present due to the change in get_session_properties. And I do have a default added for that config value.

linchan-ms: What do you mean by that? Why the "kind" is not in constants file?
In real code, it is guaranteed to be present due to the change in get_session_properties. And I do have a default added for that config value.

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: Yes I guess that is safer. Will change that.

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: I deliberately skipped that for backward compat reason. For this version of magic to work with Livy 0.2, user can turn on heartbeat but set timeout to 0. That way, we will still start the heartbeat thread in jupyter but not send the extra param to livy (which would cause 0.2 livy to fail).

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: My bad! I missed the `get_session_properties` change. Let me address my own comments with this in mind now.

As for "kind" not being in constants file, that's just bad coding on my side. No reason to repeat my mistakes though.

aggFTW: On second thoughts, we should have all the logic for `heartbeatTimeoutInSecond` in the same place, and not in two different places. Other people will also miss it.

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -55,15 +55,17 @@ def stop(self):
...
-            properties.pop(u"heartbeatTimeoutInSecond")
+        assert constants.LIVY_KIND_PARAM in list(properties.keys())
+        kind = properties[constants.LIVY_KIND_PARAM]
+
+        should_heartbeat = False
+        if heartbeat_timeout > 0:
+            should_heartbeat = True
+            properties[constants.LIVY_HEARTBEAT_TIMEOUT_PARAM] = heartbeat_timeout
+        elif properties.get(constants.LIVY_HEARTBEAT_TIMEOUT_PARAM, -1) != -1:
+            properties.pop(constants.LIVY_HEARTBEAT_TIMEOUT_PARAM)
```
aggFTW: Add a default value to this pop so that KeyError won't bite you?

PR #284, opened 2016-10-12
Added Kerberos authentication

praveenkanamarlapudi:
We can connect to kerberos enabled Livy now.

### DIFF
--- a/sparkmagic/setup.py
+++ b/sparkmagic/setup.py
@@ -77,6 +77,7 @@ setup(name=NAME,
           'pandas>=0.17.1',
           'numpy',		
           'requests',
+          'requests_kerberos'
           'ipykernel>=4.2.2,<5',		
           'ipywidgets>5.0.0',
           'notebook>=4.2',
--- a/sparkmagic/setup.py
+++ b/sparkmagic/setup.py
@@ -77,7 +77,7 @@ setup(name=NAME,
           'pandas>=0.17.1',
           'numpy',		
           'requests',
-          'requests_kerberos'
+          'requests_kerberos',
           'ipykernel>=4.2.2,<5',		
           'ipywidgets>5.0.0',
           'notebook>=4.2',
--- a/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
+++ b/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
@@ -2,6 +2,7 @@
 # Distributed under the terms of the Modified BSD License.
 from sparkmagic.controllerwidget.abstractmenuwidget import AbstractMenuWidget
 from sparkmagic.livyclientlib.endpoint import Endpoint
+import sparkmagic.utils.constants as constants
 
 
 class AddEndpointWidget(AbstractMenuWidget):
@@ -32,6 +33,11 @@ class AddEndpointWidget(AbstractMenuWidget):
             value='password',
             width=widget_width
         )
+        self.auth_type = self.ipywidget_factory.get_dropdown(
+            options={constants.AUTH_KERBEROS: constants.AUTH_KERBEROS, constants.AUTH_SSL: constants.AUTH_SSL,
+                     constants.NO_AUTH: constants.NO_AUTH},
+            description=u"Auth type:"
+        )
 
         # Submit widget
         self.submit_widget = self.ipywidget_factory.get_submit_button(
@@ -39,14 +45,14 @@ class AddEndpointWidget(AbstractMenuWidget):
         )
 
         self.children = [self.ipywidget_factory.get_html(value="<br/>", width=widget_width),
-                         self.address_widget, self.user_widget, self.password_widget,
+                         self.address_widget, self.auth_type, self.user_widget, self.password_widget,
                          self.ipywidget_factory.get_html(value="<br/>", width=widget_width), self.submit_widget]
 
         for child in self.children:
             child.parent_widget = self
 
     def run(self):
-        endpoint = Endpoint(self.address_widget.value, self.user_widget.value, self.password_widget.value)
+        endpoint = Endpoint(self.address_widget.value, self.auth_type.value, self.user_widget.value, self.password_widget.value)
         self.endpoints[self.address_widget.value] = endpoint
         self.ipython_display.writeln("Added endpoint {}".format(self.address_widget.value))
 
--- a/sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
+++ b/sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
@@ -1,6 +1,11 @@
 # Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
+from threading import Event
+import time
+
 from sparkmagic.controllerwidget.abstractmenuwidget import AbstractMenuWidget
+import sparkmagic.utils.constants as constants
+from sparkmagic.utils.kerberosthread import KerberosThread
 
 
 class ManageEndpointWidget(AbstractMenuWidget):
@@ -10,7 +15,7 @@ class ManageEndpointWidget(AbstractMenuWidget):
 
         self.endpoints = endpoints
         self.refresh_method = refresh_method
-
+        self.kerberos_info = {}
         self.children = self.get_existing_endpoint_widgets()
 
         for child in self.children:
@@ -113,7 +118,7 @@ class ManageEndpointWidget(AbstractMenuWidget):
     def get_info_endpoint_widget(self, endpoint, url):
         # 400 px
         width = "400px"
-
+        self.initialize_kerberos(endpoint)
         info_sessions = self.spark_controller.get_all_sessions_endpoint_info(endpoint)
 
         if len(info_sessions) > 0:
@@ -122,3 +127,12 @@ class ManageEndpointWidget(AbstractMenuWidget):
             text = "No sessions on this endpoint."
 
         return self.ipywidget_factory.get_html(text, width=width)
+
+    def initialize_kerberos(self, endpoint):
+        if endpoint.auth_type == constants.AUTH_KERBEROS and self.kerberos_info.get(endpoint.username) is None:
+            # Stop flag can be used to stop renewing kerberos ticket for the user
+            stop_flag = Event()
+            KerberosThread(stop_flag, endpoint).start()
+            self.kerberos_info[endpoint.username] = stop_flag
+            # Wait for kerberos ticket to be obtained
+            time.sleep(3)
--- a/sparkmagic/sparkmagic/livyclientlib/endpoint.py
+++ b/sparkmagic/sparkmagic/livyclientlib/endpoint.py
@@ -1,14 +1,16 @@
 from .exceptions import BadUserDataException
+import sparkmagic.utils.constants as constants
 
 
 class Endpoint(object):
-    def __init__(self, url, username="", password=""):
+    def __init__(self, url, auth_type=constants.NO_AUTH, username="", password=""):
         if not url:
             raise BadUserDataException(u"URL must not be empty")
         self.url = url.rstrip(u"/")
         self.username = username
         self.password = password
-        self.authenticate = False if username == '' and password == '' else True
+        self.auth_type = auth_type
+        self.authenticate = False if self.auth_type == constants.NO_AUTH else True
 
     def __eq__(self, other):
         if type(other) is not Endpoint:
--- a/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
+++ b/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
@@ -2,11 +2,13 @@
 # Distributed under the terms of the Modified BSD License.
 import json
 from time import sleep
+from requests_kerberos import HTTPKerberosAuth, OPTIONAL
 import requests
 
 import sparkmagic.utils.configuration as conf
 from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.utils.constants import MAGICS_LOGGER_NAME
+import sparkmagic.utils.constants as constants
 from sparkmagic.livyclientlib.exceptions import HttpClientException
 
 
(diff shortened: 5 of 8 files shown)


kpraveen420:
Do I need to get CCLA agreement for contribution?

sparkmagic/sparkmagic/utils/constants.py
```diff
@@ -76,3 +76,11 @@
 
 LIVY_HEARTBEAT_TIMEOUT_PARAM = u"heartbeatTimeoutInSecond"
 LIVY_KIND_PARAM = u"kind"
+
+KERBEROS_KINIT = 'kinit'
+# Time in seconds
+KERBEROS_TIME_INTERVAL = 6000
+
+NO_AUTH = "no auth"
```
aggFTW: `"None"`

sparkmagic/sparkmagic/utils/constants.py
```diff
@@ -76,3 +76,11 @@
...
+AUTH_KERBEROS = "kerberos"
```
aggFTW: `"Kerberos"`

sparkmagic/sparkmagic/utils/constants.py
```diff
@@ -76,3 +76,11 @@
...
+AUTH_SSL = "username/password"
```
aggFTW: `"Basic Access"`

sparkmagic/sparkmagic/utils/constants.py
```diff
@@ -76,3 +76,11 @@
 
 LIVY_HEARTBEAT_TIMEOUT_PARAM = u"heartbeatTimeoutInSecond"
 LIVY_KIND_PARAM = u"kind"
+
+KERBEROS_KINIT = 'kinit'
+# Time in seconds
+KERBEROS_TIME_INTERVAL = 6000
```
aggFTW: This should not be a constant, but a configuration.

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
```diff
@@ -32,21 +33,26 @@ def __init__(self, spark_controller, ipywidget_factory, ipython_display, endpoin
             value='password',
             width=widget_width
         )
+        self.auth_type = self.ipywidget_factory.get_dropdown(
+            options={constants.AUTH_KERBEROS: constants.AUTH_KERBEROS, constants.AUTH_SSL: constants.AUTH_SSL,
+                     constants.NO_AUTH: constants.NO_AUTH},
+            description=u"Auth type:"
+        )
```
aggFTW: On change, we should hide/show the appropriate fields. So:

* On change to NO_AUTH: hide username and password.
* On change to AUTH_KERBEROS: show username and password.
* On change to AUTH_SSL: show username and password.

sparkmagic/sparkmagic/utils/constants.py
aggFTW: Also, please name it so that the user knows what time unit to use: `kerberos_renew_time_interval_seconds` or something similar.

sparkmagic/sparkmagic/utils/kerberosthread.py
```diff
@@ -0,0 +1,26 @@
...
+        self.kinit()
+        while not self.stopped.wait(constants.KERBEROS_TIME_INTERVAL):
+            self.kinit()
+
+    def kinit(self):
+        kinit_args = [constants.KERBEROS_KINIT, self.endpoint.username]
+        kinit = Popen(kinit_args, stdin=PIPE, stdout=PIPE, stderr=PIPE)
+        kinit.stdin.write(('%s\n' % self.endpoint.password).encode())
+        kinit.stdin.flush()
+        kinit.wait()
```
aggFTW: How do we bubble up failures to the user? For example: wrong password.

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
```diff
@@ -122,3 +127,12 @@ def get_info_endpoint_widget(self, endpoint, url):
             text = "No sessions on this endpoint."
 
         return self.ipywidget_factory.get_html(text, width=width)
+
+    def initialize_kerberos(self, endpoint):
```
aggFTW: What about getting the ticket directly from `ReliableHttpClient`? Then, upon creation of the object, the ticket will be obtained (and renewed when needed) and when the client is not to be used anymore, the client would also call `kdestroy`.

We could probably achieve this by creating a `start` and `stop` method on `ReliableHttpClient`. We can also call those methods from `__enter__` and `__exit__` so that the object can also be used with a `with` clause.

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
```diff
@@ -122,3 +127,12 @@ def get_info_endpoint_widget(self, endpoint, url):
...
+        if endpoint.auth_type == constants.AUTH_KERBEROS and self.kerberos_info.get(endpoint.username) is None:
+            # Stop flag can be used to stop renewing kerberos ticket for the user
+            stop_flag = Event()
+            KerberosThread(stop_flag, endpoint).start()
+            self.kerberos_info[endpoint.username] = stop_flag
```
aggFTW: How can the user use the `stop_flag`? This sounds like something we could automate for the user as soon as the kernel dies or the endpoint is removed?

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -17,6 +19,10 @@ def __init__(self, endpoint, headers, retry_policy):
         self._endpoint = endpoint
         self._headers = headers
         self._retry_policy = retry_policy
+        if self._endpoint.auth_type == constants.AUTH_KERBEROS:
+            self._auth = HTTPKerberosAuth(mutual_authentication=OPTIONAL)
```
aggFTW: This logic is not being unit tested at the moment. Could you please add tests?

sparkmagic/setup.py
```diff
@@ -77,6 +77,7 @@ def version(path):
           'pandas>=0.17.1',
           'numpy',		
           'requests',
+          'requests_kerberos',
```
aggFTW: Can you please add it to the `requirements.txt` file too please?

sparkmagic/sparkmagic/livyclientlib/endpoint.py
```diff
@@ -1,14 +1,16 @@
 from .exceptions import BadUserDataException
+import sparkmagic.utils.constants as constants
 
 
 class Endpoint(object):
-    def __init__(self, url, username="", password=""):
+    def __init__(self, url, auth_type=constants.NO_AUTH, username="", password=""):
```
aggFTW: Default should be None. If None, then get the value from a configuration called `default_authentication_method`.

aggFTW:
Let me talk to some people on our team because I'm no Kerberos expert. Thanks!

languy:
Great elegant way to do kerberos auth!
A few comments/suggestions;
- as aggFTW commented: it would be nice to bubble up errors back to the user:
When getting the ticket: wrong credentials, unable to connect to hostname
What happens if the ticket expires (before it was renewed)? Can the end-user understand the issue and recover?
- if I understand the code correctly, a thread is spawned that renews the ticket in an open loop:
  - Once the ticket gets renewed, is there a way to verify that the ticket is actually there? (maybe by checking the ticket cache, or does the request actually return the ticket?) This would help error handling.
  - If you can access the ticket after kinit: check its expiration date in order to adjust the renewal time. Your defined 6000 sec renewal delay seems much shorter than the default TGT validity. This would be ideal but at a minimum. making it configurable and setting this to a more typical value would allow users to adapt it to their specific kerberos setup.

languy:
On deleting tickets: does this require installing kerberos5 on the host? If so, the krb5 utilities will take care of eventually purging expired tickets. However, it won't invalidate tickets if that's what we want to do upon shutdown.

joychak:
What happens when the client already had a Kerberos ticket (by doing kinit) before logging into Jupyter-hub and opening a notebook? In this scenario the Jupyter's KDC Authentication module will use the HTTP service principal to validate/verify the client's Kerberos ticket through HTTP-authenticate/spnego and pass it to SparkMagic (or SparkMagic has to pick it up from jupyter's spawner process) to use it for further communication to Livy?

sparkmagic/sparkmagic/utils/constants.py
praveenkanamarlapudi: @aggFTW, we changed this and updated PR will contain these changes.

sparkmagic/sparkmagic/utils/constants.py
praveenkanamarlapudi: @aggFTW, we are changing this and updated PR will contain these changes.

sparkmagic/sparkmagic/utils/constants.py
praveenkanamarlapudi: @aggFTW, we are changing this and updated PR will contain these changes.

sparkmagic/sparkmagic/utils/constants.py
praveenkanamarlapudi: @aggFTW, we are changing this and updated PR will contain these changes.

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
praveenkanamarlapudi: @aggFTW I think it's better not to destroy the kerberos ticket. There may be multiple processes using the kerberos ticket. Consider livy and jupyter running on the same server, destroying the ticket will effect livy.

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
praveenkanamarlapudi: Sure. Updated PR will contain unit tests for the logic.

sparkmagic/setup.py
praveenkanamarlapudi: Sure. Updated PR will contain this change

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
```diff
@@ -122,3 +127,12 @@ def get_info_endpoint_widget(self, endpoint, url):
...
+            # Wait for kerberos ticket to be obtained
+            time.sleep(3)
```
praveenkanamarlapudi: Sure. Updated PR will contain this change

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
praveenkanamarlapudi: @aggFTW We thought of implementing this change in later. If it's javascript, can you help me understand where can I add javascript for these king of changes.

sparkmagic/sparkmagic/utils/kerberosthread.py
praveenkanamarlapudi: We are working on this. Updated PR will contain this change.

sparkmagic/sparkmagic/livyclientlib/endpoint.py
praveenkanamarlapudi: Do auth_type needs a default value?

praveenkanamarlapudi:
@languy 

> as aggFTW commented: it would be nice to bubble up errors back to the user: When getting the ticket: wrong credentials, unable to connect to hostname What happens if the ticket expires (before it was renewed)? Can the end-user understand the issue and recover?

We are working on checking kerberos ticket expiration time and adjust the renewal time as per it. There is another scenario, some other process might have destroyed the ticket in between the renewal. So we can renew the ticket when we see http response code as 401. 

> Once the ticket gets renewed, is there a way to verify that the ticket is actually there? (maybe by checking the ticket cache, or does the request actually return the ticket?) This would help error handling.

We can check the subprocess exits status. Exit status can help to know the ticket status

> If you can access the ticket after kinit: check its expiration date in order to adjust the renewal time. Your defined 6000 sec renewal delay seems much shorter than the default TGT validity. This would be ideal but at a minimum. making it configurable and setting this to a more typical value would allow users to adapt it to their specific kerberos setup.

Sure. I will change the renewal delay.

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
aggFTW: It's not. Ipywidgets has a way to make widgets visible or not:
https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/autovizwidget/widget/encodingwidget.py#L126

aggFTW:
@joychak do you have a suggestion on how to address that use case?

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
praveenkanamarlapudi: @prabhu1984 can you put your thoughts on this issue.

joychak:
@aggFTW, we are currently working on it and will submit the PR very soon which involves new KDCAuthenticator for JupyterHub, KDCSpawner for JupyterHub, changes in SparkMagic to use the client's kerberos ticket (or optionally using SparkMagic specific keytab) to authenticate with Livy along with adding the proxy-user info in the request for Livy.

praveenkanamarlapudi:
@joychak Thanks for the update. 
I think it's good have kerberos enabled in JupyterHub than spark magic. I am closing the PR and waiting to see kerberos in JupyterHub..

praveenkanamarlapudi:
@joychak Thanks for the update. 
I think it's good have kerberos enabled in JupyterHub than spark magic. I am closing the PR and waiting to see kerberos in JupyterHub..

aggFTW:
Thanks @joychak and @praveenkanamarlapudi!
Looking forward to the PR to continue the discussion.

prabhu1984:
Hi @joychak,

That's good to hear. One quick question.

How will it work if user wants to run only with Jupyter (spark magic) without Jupyterhub?

I.e. It's very easy to setup for anyone to deploy Jupyter and Sparkmagic package on any host without root/admin access and starting jupyter. Whereas, Installing jupyterhub requires pre-requisites like node installation, configurable-http-proxy installation etc. 

Thanks
Prabhu

joychak:
@prabhu1984, let me look into the Jupyter notebook (without hub) authentication part and get back.  But good point indeed.

PR #287, opened 2016-10-18
Update Sparkmagic to use SparkSession for Spark2

jeffersonezra:
In Spark2, we have SparkSession (variable name spark) instead of HiveContext (variable name sqlContext). Updated SqlQuery to support this and added corresponding tests. Also bumped the patch version.


sparkmagic/sparkmagic/tests/test_sqlquery.py
```diff
@@ -194,3 +194,79 @@ def test_unicode_sql():
         assert False
     except NotImplementedError:
         pass
+
+@with_setup(_setup, _teardown)
+def test_pyspark_livy_sql_options_spark2():
+    try:
+
```
aggFTW: Remove extra line

sparkmagic/sparkmagic/tests/test_sqlquery.py
```diff
@@ -194,3 +194,79 @@ def test_unicode_sql():
...
+    finally:
+        defaults = {
+            conf.use_spark_session.__name__: False,
+        }
+        conf.override_all(defaults)
+
+@with_setup(_setup, _teardown)
+def test_scala_livy_sql_options_spark2():
+    try:
+
```
aggFTW: Remove empty line

sparkmagic/sparkmagic/tests/test_sqlquery.py
```diff
@@ -194,3 +194,79 @@ def test_unicode_sql():
...
+                              .format(LONG_RANDOM_VARIABLE_NAME, query,
+                                      LONG_RANDOM_VARIABLE_NAME, conf.pyspark_sql_encoding())))
+
+        sqlquery = SQLQuery(query, samplemethod='sample', samplefraction=0.33, maxrows=3234)
+        assert_equals(sqlquery._pyspark_command(),
+                      Command(u'for {} in spark.sql(u"""{} """).toJSON().sample(False, 0.33).take(3234): '
+                              u'print({}.encode("{}"))'\
+                              .format(LONG_RANDOM_VARIABLE_NAME, query,
+                                      LONG_RANDOM_VARIABLE_NAME, conf.pyspark_sql_encoding())))
+
```
aggFTW: Remove empty line

sparkmagic/sparkmagic/tests/test_sqlquery.py
```diff
@@ -194,3 +194,79 @@ def test_unicode_sql():
...
+                      Command('spark.sql("""{}""").toJSON.collect.foreach(println)'.format(query)))
+
+        sqlquery = SQLQuery(query, samplemethod='sample', samplefraction=0.25, maxrows=-1)
+        assert_equals(sqlquery._scala_command(),
+                      Command('spark.sql("""{}""").toJSON.sample(false, 0.25).collect.foreach(println)'.format(query)))
+
+        sqlquery = SQLQuery(query, samplemethod='sample', samplefraction=0.33, maxrows=3234)
+        assert_equals(sqlquery._scala_command(),
+                      Command('spark.sql("""{}""").toJSON.sample(false, 0.33).take(3234).foreach(println)'.format(query)))
+
```
aggFTW: Remove empty line

tc0312:
I've some concerns on adding a global config for this. I would prefer a per session capability detection approach.

aggFTW:
Is there a way to do detect that once a session is created in Livy?

tc0312:
Yes. We can execute a statement to check does a variable exist.

aggFTW:
When there's more than 2 options (created vs not created), for example a name change, wouldn't this increase the latency?

What about Livy returning the session version on session creation as a field? Would that be a viable approach?

tc0312:
In practice, name change will not happen because everyone is follow spark-shell's naming. There's a pending PR to eliminate the 1 second update latency. Also, it's possible to batch submit statements.
The latency caused by Livy should be sub-second and acceptable.

We have explored the option to expose them in Livy via appInfo. It's definitely doable, I haven't seen anyone opposing the idea. However given the large amount of changes happening right now in Livy (Spark 2.0 and recovery), it will take a while before it gets implemented.

aggFTW:
Got it. Do you want to revisit this once it's exposed via appInfo or do you foresee us running into problems with this approach?

PR #292, opened 2016-11-03
Added PySpark3 kernel.

tc0312:
image


sparkmagic/sparkmagic/kernels/pyspark3kernel/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.0.1'
```
aggFTW: This doesn't need a version number, does it?

msftristew: No

tc0312: I will remove the version.
I cloned this from pysparkkernel. I can remove versions in pysparkkernel and sparkkernel in a different PR if you want.

linchan-ms:
LGTM

sparkmagic/sparkmagic/kernels/pyspark3kernel/__init__.py
aggFTW: Mmm... don't know why that was there in the first place, but thanks! That would be good!

aggFTW:
Looks great after the changes! Merge after Travis passes 👍

tc0312:
Mind if we do a squash merge to hide the feedback commit? BTW I don't have right to push changes to sparkmagics.

PR #298, opened 2016-11-21
Print resource limit warning early.

linchan-ms:
Print a warning message early if detected resource limitation at session startup.


sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -252,12 +260,12 @@ def sleep(self):
...
     def refresh_status(self):
-        status = self._http_client.get_session(self.id)[u'state']
+        response = self._http_client.get_session(self.id)
+        status = response[u'state']
+        log_array = response[u'log']
 
-        if status in constants.POSSIBLE_SESSION_STATUS:
-            self.status = status
-        else:
-            raise LivyUnexpectedStatusException(u"Status '{}' not supported by session.".format(status))
```
aggFTW: Why get rid of the check here?

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -252,12 +260,12 @@ def sleep(self):
...
+        self.status = status
+        self._logs = "\n".join(log_array)
```
aggFTW: We have `self.get_logs` as a way of retrieving logs.

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -242,6 +244,12 @@ def wait_for_idle(self, seconds_to_wait=None):
                 self.logger.error(error)
                 raise LivyClientTimeoutException(error)
 
+            if constants.YARN_RESOURCE_LIMIT_MSG in self._logs and \
+                not self._printed_resource_warning:
+                self.ipython_display.writeln(u"Warning: Session not starting due to resource limitation. {}"\
```
aggFTW: This should be written as an error, not simple line.

aggFTW:
I'm suspecting Travis is failing because of `requests-2.12.1` being installed. I believe requests 2.10.0 works fine.

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: `self.get_logs` gets all logs, not only last.

sparkmagic/sparkmagic/livyclientlib/livysession.py
msftristew: This to me is not a very clear error message. What does "resource limitation" mean?

sparkmagic/sparkmagic/livyclientlib/livysession.py
msftristew: Can this be edited as part of a second PR?  I wouldn't feel comfortable making a change like this without an in-depth discussion about the tradeoffs involved, along with unit tests to go along with it

sparkmagic/sparkmagic/livyclientlib/livysession.py
msftristew: Can the name of the method be updated to reflect the new functionality?

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -242,6 +244,12 @@ def wait_for_idle(self, seconds_to_wait=None):
                 self.logger.error(error)
                 raise LivyClientTimeoutException(error)
 
+            if constants.YARN_RESOURCE_LIMIT_MSG in self._logs and \
```
msftristew: Searching the logs for a particular string in English seems like a bad way to go in the long-term.  This functionality breaks if at any point in the future the error message gets changed on the Livy side even by 1 character.  Is there any other (more structured API-driven) way to accomplish the same goal?

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: You can make it appear in red without throwing an exception:
https://github.com/jupyter-incubator/sparkmagic/blob/master/hdijupyterutils/hdijupyterutils/ipythondisplay.py#L29

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: Agreed with @msftristew. This library is very dependent on Livy states. Seeing one we don't know how to handle will have unknown behavior. If removing the check, we need to have a full code review, like he stated.

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: Let's make the error message a constant, while we are at it.

linchan-ms: Ok, made it an error so it can be more readable.

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: In theory, I can separate it out. But while I am doing this work, I already noticed that there are 2 new Livy state that is not in the list. If I don't add the new states, the test could randomly fail. Those 2 states should not affect how sparkmagic works at all. 
I agree sparkmagic is depending on Livy and should have enough understanding of it. We need to know the ending states: Dead/Success and some key states like Idle. However, it shouldn't have a hard coded list of all Livy states and strictly enforcing them, especially these states are not used by magic anywhere else. Livy could add a new transition state any time, and I don't think we should put magic into a corner where a new update is needed for every such Livy change. It is almost impossible to keep up with that on time. Also, what if Livy removes a state? current way will not guarantee correctness either, so i really don't see a significant value of this assert.

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: I truly don't like methods that have side effects. How bad is it if you just use `self.get_logs`?

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: unfortunately no. This can only be done by parsing the Yarn diagnostic log. No error code or other API.

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: I think separating things out into their own Pull Requests is a standard practice that makes things easier for reviewers. If you truly don't want to, we can all make an effort here to discuss two things at once, provided everyone agrees in doing so.

Do you know where the Livy documentation regarding states is? Couldn't find it. Without it, it's hard to have any discussions, really, like which states are key (is idle the only one?).

The code was written by reading the Livy code regarding states. We started enforcing the list of known states due to a bug that would leave sessions in a non-terminal non-valid state even though there was no chance of recovery. Then, sparkmagic errored out with a non valid state reached.

I'm fine with removing the check as long as we understand the state transition graph from Livy. If you are removing the check though, please remove the variable and other places where it might be used. Do we need unit tests for the change?

I really think this should be its own Pull Request.

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: The problem is we need it to be atomic. And the Yarn diagnostic log changes, it is not accumulative. So the get status call issued in Refresh_status and get_logs could have different states and different logs.
Current refresh_status does have side effect which is changing the state in the object. By changing the name to refresh_status_and_log, it is indicating what it is doing. If you are strict about the return type, we can change this function to not return anything and just refresh the internal variables and the caller can call LivySession.status to retrieve the value. To me that is no ideal either.

sparkmagic/sparkmagic/livyclientlib/livysession.py
linchan-ms: Ok will make this a separate PR.

msftristew: Thanks @linchan-ms.

sparkmagic/sparkmagic/livyclientlib/livysession.py
msftristew: `self.get_logs` does return all logs so it shouldn't affect correctness right?

sparkmagic/sparkmagic/livyclientlib/livysession.py
msftristew: Can we put in a feature request on the Livy side to add a special error code for this case?  I am fine with this in the short-term but it's brittle to put this logic in Sparkmagic.

sparkmagic/sparkmagic/livyclientlib/livysession.py
aggFTW: Wait, are you saying that this won't return the same logs (YARN vs Livy logs)?

sparkmagic/sparkmagic/utils/constants.py
```diff
@@ -76,5 +76,8 @@
                      "Please file an issue at https://github.com/jupyter-incubator/sparkmagic\nError:\n{}"
 EXPECTED_ERROR_MSG = "An error was encountered:\n{}"
 
+YARN_RESOURCE_LIMIT_MSG = "Queue's AM resource limit exceeded."
+RESOURCE_LIMIT_WARNING = "Warning: Session not starting due to resource limitation. {}"
```
msftristew: Still not a big fan of this error message, is there any clearer way of expressing this?

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -251,15 +260,18 @@ def wait_for_idle(self, seconds_to_wait=None):
...
+    # This function will refresh the status and get the logs in a single call.
+    # Only the status will be returned as the return value.
+    def refresh_status_and_info(self):
+        response = self._http_client.get_session(self.id)
+        status = response[u'state']
+        log_array = response[u'log']
 
         if status in constants.POSSIBLE_SESSION_STATUS:
             self.status = status
+            self.session_info = "\n".join(log_array)
```
aggFTW: Change to `u"\n".join(log_array)`

sparkmagic/sparkmagic/utils/constants.py
aggFTW: What about `"Warning: The Spark session does not have enough YARN resources to start up. {}"`

We could then use the `{}` to even provide a link to `/yarnui` in our clusters.

linchan-ms:
I can change the message. The idea of the {} is we will have a documentation of our own to teach them how to check yarn and kill application in yarn. That is specific to HDI so will only come as config.

aggFTW:
Just wait for Travis to pass and it feel free to squash and merge.

PR #331, opened 2017-02-16
Added shutting_down status.

praveenkanamarlapudi:
Added shutting_down status.


aggFTW:
Thanks for your contribution! I would recommend that you bump the versions of the 3 python packages in the repo if you want to be able to pip install with your changes. Look at the version bump here: https://github.com/jupyter-incubator/sparkmagic/pull/326/files

@tc0312 can you please explain the semantics of shutting_down? Is it a terminal state (as in, once it goes into it, it's guaranteed that it won't come back to a non-terminal state)?

tc0312:
ShuttingDown is a transient session state before the session disappears in Livy.

praveenkanamarlapudi:
Removed shutting_down from FINAL_STATUS as it's a transient state.

PR #333, opened 2017-02-21
Store to local for spark

### LINKED ISSUE #327: Alternate method to communicate variables between livy and local (not just using %%sql -o)
It would be great if we also could communicate in some way the contents of a dataframe/pandas in code, not just using the `%sql -o` way.

orestis-ms:
Closes #327 

Possible directions from here:
1. Allow the cell to be executed without any code (make line-magic instead of cell-magic)
2. Refactor SQLQuery, Command and the newly created SparkStoreCommand since they are all 'commands' and all inherit (directly or indirectly) from ObjectWithGuid.


sparkmagic/sparkmagic/magics/sparkmagicsbase.py
```diff
@@ -34,6 +37,21 @@ def __init__(self, shell, data=None, spark_events=None):
...
 
+    def execute_spark(self, cell, output_var, samplemethod, maxrows, samplefraction, session_name):
+        (success, out) = self.spark_controller.run_command(Command(cell), session_name)
+        if not success:
+            self.ipython_display.send_error(out)
+        else:
+            self.ipython_display.write(out)
+            if output_var is not None:
+                spark_store_command = self._spark_store_command(cell, output_var, samplemethod, maxrows, samplefraction)
+                df = self.spark_controller.run_command(spark_store_command, session_name)
```
aggFTW: What if the spark_store_command is not successful: e.g. variable doesn't exist?

aggFTW: As it is, we would simply be catching the exception and displaying it, but I wonder if there's a possibility for a better error message... Ideas?

sparkmagic/sparkmagic/tests/test_kernel_magics.py
```diff
@@ -502,10 +502,50 @@ def test_spark_expected_exception():
...
+    magic.spark(line, cell)
+    assert spark_controller.run_command.call_count == 2
+    spark_controller.run_command.assert_any_call(Command(cell), None)
+    ipython_display.send_error.assert_called_with(constants.INTERNAL_ERROR_MSG
+                                                       .format(side_effect[1]))
+
+
+@with_setup(_setup, _teardown)
+def test_spark_expected_exception_in_storing():
+    line = "-o var_name"
```
aggFTW: This test's name is the same as the one above. Please change the one above to unexpected.

sparkmagic/sparkmagic/tests/test_kernel_magics.py
```diff
@@ -502,10 +502,50 @@ def test_spark_expected_exception():
...
+
+@with_setup(_setup, _teardown)
+def test_spark_sample_options():
+    line = "-o var_name -m sample -n 142 -r 0.3"
+    cell = ""
+    magic.execute_spark = MagicMock()
+
+    ret = magic.spark(line, cell)
+
+    spark_controller.add_session.assert_called_once_with(magic.session_name, magic.endpoint, False,
```
aggFTW: Why are we checking that `add_session` executes?

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
```diff
@@ -269,6 +269,99 @@ def test_run_cell_command_exception():
...
+    meth = "-m"
+    method_name = "sample"
+    output = "-o"
+    output_var = "var_name"
+    line = " ".join([command, name, context, context_name, meth, method_name, output, output_var])
+    cell = "cell code"
+
+    result = magic.spark(line, cell)
+
+    run_cell_method.assert_any_call(Command(cell), name)
```
aggFTW: Assert that the result of Command (`""` in this unit test) is printed to the user.

sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
```diff
@@ -131,3 +131,46 @@ def test_print_empty_endpoint_info():
     magic._print_endpoint_info([], current_session_id)
     magic.ipython_display.html.assert_called_once_with(u'No active sessions.')
 
+
+def test_spark_execution_without_output_var():
+    magic = SparkMagicBase(None)
```
aggFTW: It seems like the setup can be done in a common `setup` method.

sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
```diff
@@ -131,3 +131,46 @@ def test_print_empty_endpoint_info():
...
+    shell = MagicMock()
+    shell.user_ns = {}
+    magic.shell = shell
+
+    session = MagicMock()
+    output_var = "var_name"
+    code = ""
+    df = 'df'
+
+    magic.spark_controller.run_command.side_effect = [(True,'out'), df]
```
aggFTW: We're missing a case here where the second command is an exception.

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
```diff
@@ -269,6 +269,99 @@ def test_run_cell_command_exception():
...
+    ipython_display.send_error.assert_called_once_with(EXPECTED_ERROR_MSG
+                                                       .format(run_cell_method.side_effect))
+
+@with_setup(_setup, _teardown)
+def test_run_spark_command_exception_while_storing():
+    run_cell_method = MagicMock()
```
aggFTW: This unit test should be in SparkMagicBase tests

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
```diff
@@ -269,6 +269,99 @@ def test_run_cell_command_exception():
...
+
+    result = magic.spark(line, cell)
+
+    run_cell_method.assert_any_call(Command(cell), name)
+    run_cell_method.assert_any_call(SparkStoreCommand(cell, output_var, samplemethod=method_name), name)
+
+
+@with_setup(_setup, _teardown)
+def test_run_spark_command_exception():
+    run_cell_method = MagicMock()
```
aggFTW: This unit test should be in SparkMagicBase tests

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
```diff
@@ -269,6 +269,99 @@ def test_run_cell_command_exception():
...
+    meth = "-m"
+    method_name = "sample"
+    output = "-o"
+    output_var = "var_name"
+    line = " ".join([command, name, context, context_name, meth, method_name, output, output_var])
+    cell = "cell code"
+
+    result = magic.spark(line, cell)
+
+    run_cell_method.assert_any_call(Command(cell), name)
```
aggFTW: We should just assert that we are calling the base method.

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
```diff
@@ -269,6 +269,99 @@ def test_run_cell_command_exception():
...
+    context = "-c"
+    context_name = "spark"
+    meth = "-m"
+    method_name = "sample"
+    line = " ".join([command, name, context, context_name, meth, method_name])
+    cell = "cell code"
+
+    result = magic.spark(line, cell)
+
+    run_cell_method.assert_called_once_with(Command(cell), name)
```
aggFTW: We should just assert that we are calling the base method.

sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
```diff
@@ -131,3 +131,46 @@ def test_print_empty_endpoint_info():
...
+    magic = SparkMagicBase(None)
+    magic.spark_controller = MagicMock()
+    magic.ipython_display = MagicMock()
+
+    session = MagicMock()
+    output_var = None
+    
+    magic.spark_controller.run_command.return_value = (True,'out')
+    magic.execute_spark("", output_var, None, None, None, session)
+    magic.ipython_display.write.assert_called_once_with('out')
```
aggFTW: Assert that run_cell_method is not called with SparkStoreCommand

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+def test_to_command_pyspark():
+    variable_name = "var_name"
+    sparkcommand = SparkStoreCommand("Code", variable_name)
+    sparkcommand._pyspark_command = MagicMock(return_value=MagicMock())
+    sparkcommand.to_command("pyspark", variable_name)
+    sparkcommand._pyspark_command.assert_called_with(variable_name)
+
+
+@with_setup(_setup, _teardown)
+def test_to_command_pyspark3():
```
aggFTW: test_to_command for scala and R?

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+@with_setup(_setup, _teardown)
+def test_sparkstorecommand_loads_defaults():
+    defaults = {
+        conf.default_samplemethod.__name__: "sample",
+        conf.default_maxrows.__name__: 419,
+        conf.default_samplefraction.__name__: 0.99,
+    }
+    conf.override_all(defaults)
```
aggFTW: You need to undo the override on `_teardown` above

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+    variable_name = "var_name"
+    code = "hugedataset.collect()"
+    sparkcommand = SparkStoreCommand(code, variable_name)
+    assert_equals(sparkcommand.code, code)
+    assert_equals(sparkcommand.samplemethod, defaults[conf.default_samplemethod.__name__])
+    assert_equals(sparkcommand.maxrows, defaults[conf.default_maxrows.__name__])
+    assert_equals(sparkcommand.samplefraction, defaults[conf.default_samplefraction.__name__])
+
```
aggFTW: Nit: Add extra line

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+    code = "hugedataset.collect()"
+    samplemethod = "foo"
+    _ = SparkStoreCommand(code, variable_name, samplemethod)
+
+
+@with_setup(_setup, _teardown)
+def test_pyspark_livy_sampling_options():
+    variable_name = "abc"
+    code = "hugedataset.collect()"
+    sparkcommand = SparkStoreCommand(code, variable_name, samplemethod='take', maxrows=120)
```
aggFTW: We are missing a test for when sample method is `None` and it should do a `collect` for all 4 languages

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+                                  LONG_RANDOM_VARIABLE_NAME, conf.pyspark_python_encoding())))
+
+    sparkcommand = SparkStoreCommand(code, variable_name, samplemethod='sample', samplefraction=0.33, maxrows=3234)
+    assert_equals(sparkcommand._pyspark_command(variable_name),
+                  Command(u'for {} in {}.toJSON().sample(False, 0.33).take(3234): '
+                          u'print({}.encode("{}"))'\
+                          .format(LONG_RANDOM_VARIABLE_NAME, variable_name,
+                                  LONG_RANDOM_VARIABLE_NAME, conf.pyspark_python_encoding())))
+
+    sparkcommand = SparkStoreCommand(code, variable_name, samplemethod='sample', samplefraction=0.33, maxrows=3234)
```
aggFTW: duplicated

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
```diff
@@ -0,0 +1,147 @@
...
+import sparkmagic.utils.configuration as conf
+import sparkmagic.utils.constants as constants
+from sparkmagic.utils.sparkevents import SparkEvents
+from sparkmagic.livyclientlib.command import Command
+from sparkmagic.livyclientlib.exceptions import DataFrameParseException, BadUserDataException
+
+import ast
+
+class SparkStoreCommand(Command):
+    def __init__(self, code, output_var, samplemethod=None, maxrows=None, samplefraction=None, spark_events=None):
```
aggFTW: It shouldn't take `code`

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
```diff
@@ -0,0 +1,147 @@
...
+            command = u'collect({})'.format(command)
+        command = u'jsonlite::toJSON({})'.format(command)
+        command = u'for ({} in ({})) {{cat({})}}'.format(constants.LONG_RANDOM_VARIABLE_NAME,
+                                                         command,
+                                                         constants.LONG_RANDOM_VARIABLE_NAME)
+        return Command(command)
+
+
+    @staticmethod
+    def _records_to_dataframe(records_text, kind):
```
aggFTW: It should be easy to pull this method out into `utils.py` without much more refactoring and use it for this class and `SQLQuery`.

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
```diff
@@ -0,0 +1,147 @@
...
+        self.samplefraction = samplefraction
+        self.output_var = output_var
+        if spark_events is None:
+            spark_events = SparkEvents()
+        self._spark_events = spark_events
+
+    def execute(self, session):
+        return self.store_to_context(session)
+
+    def store_to_context(self, session):
```
aggFTW: This should just be in execute. No need to have this indirection.

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
```diff
@@ -0,0 +1,147 @@
...
+
+            coerce_pandas_df_to_numeric_datetime(df)
+            return df
+        except ValueError:
+            raise DataFrameParseException(u"Cannot parse object as JSON: '{}'".format(strings))
+
+    # Used only for unit testing
+    def __eq__(self, other):
+        return self.code == other.code and \
+            hasattr(other, 'samplemethod') and \
```
aggFTW: Why the `hasattr`?

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+@with_setup(_setup, _teardown)
+def test_execute_code():
+    spark_events = MagicMock()
+    variable_name = "abc"
+    code = "hugedataset.collect()"
+    sparkcommand = SparkStoreCommand(code, variable_name, "take", 100, 0.2, spark_events=spark_events)
+    sparkcommand.to_command = MagicMock(return_value=MagicMock())
+    result = """{"z":100, "nullv":null, "y":50}
+{"z":25, "nullv":null, "y":10}"""
+    sparkcommand.to_command.return_value.execute = MagicMock(return_value=(True, result))
```
aggFTW: The 3 lines above are hard to parse 😛 
Can we introduce a `command` MagicMock variable. That way you can define what its `execute` method does more clearly and add it as the `return_value` of `to_command`.

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+    result_data = pd.DataFrame([{'z': 100, "nullv": None, 'y': 50}, {'z':25, "nullv":None, 'y':10}], columns=['z', "nullv", 'y'])
+    session = MagicMock()
+    session.kind = "pyspark"
+    result = sparkcommand.execute(session)
+    assert_frame_equal(result, result_data)
+    sparkcommand.to_command.assert_called_once_with(session.kind, variable_name)
+    sparkcommand.to_command.return_value.execute.assert_called_once_with(session)
+
```
aggFTW: Nit: extra line

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+                  Command(u'for {} in {}.toJSON().take(120): print({}.encode("{}"))'\
+                          .format(LONG_RANDOM_VARIABLE_NAME, variable_name,
+                                  LONG_RANDOM_VARIABLE_NAME, conf.pyspark_python_encoding())))
+    assert_equals(sparkcommand._scala_command(variable_name),
+                  Command(u'{}.toJSON.take(120).foreach(println)'.format(variable_name)))
+
+
+@with_setup(_setup, _teardown)
+def test_pyspark_livy_sampling_options_spark2():
+        variable_name = "abc"
```
aggFTW: How are you specifying that these `SparkStoreCommand`s are Spark 2?

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+@with_setup(_setup, _teardown)
+def test_execute_code():
+    spark_events = MagicMock()
+    variable_name = "abc"
+    code = "hugedataset.collect()"
+    sparkcommand = SparkStoreCommand(code, variable_name, "take", 100, 0.2, spark_events=spark_events)
+    sparkcommand.to_command = MagicMock(return_value=MagicMock())
+    result = """{"z":100, "nullv":null, "y":50}
```
aggFTW: Do we have unit tests for the code that transforms the strings to dataframes? I think we do... if so, we should move those tests to the utils tests when you move the function over. That way, you don't have to test that here, and you need only mock the utils function.

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,241 @@
...
+        sparkcommand = SparkStoreCommand(code, variable_name, samplemethod='sample', samplefraction=0.33, maxrows=3234)
+        assert_equals(sparkcommand._pyspark_command(variable_name),
+                      Command(u'for {} in {}.toJSON().sample(False, 0.33).take(3234): '
+                              u'print({}.encode("{}"))'\
+                              .format(LONG_RANDOM_VARIABLE_NAME, variable_name,
+                                      LONG_RANDOM_VARIABLE_NAME, conf.pyspark_python_encoding())))
+
+@with_setup(_setup, _teardown)
+def test_scala_livy_options_spark2():
+        variable_name = "abc"
```
aggFTW: Ditto on Spark 2. Am I going blind? 👅

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -175,7 +175,10 @@ def default_samplefraction():
 def pyspark_sql_encoding():
     return u'utf-8'
     
-    
+@_with_override
+def pyspark_python_encoding():
```
aggFTW: Why not use the same sql encoding variable as above?

sparkmagic/sparkmagic/tests/test_kernel_magics.py
orestis-ms: fixed

sparkmagic/sparkmagic/tests/test_kernel_magics.py
orestis-ms: I assumed there is a reason it happens for sql right below. 
I agree, from the code there is no clear reason why this is happening. Removed the check

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
orestis-ms: comment not clear. We are checking the vanilla-case, simple good old spark command; no storing, hence no storing class created.

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
orestis-ms: same as above.

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
orestis-ms: same for test_run_sql_command_exception() ? Similar tests exist in test_sparkmagicbase.py

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
orestis-ms: same as above

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
orestis-ms: Roger

sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
orestis-ms: Agree; fixed.

sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
orestis-ms: Added check that no SparkStoreCommand is created altogether

sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
orestis-ms: Added

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
orestis-ms: indeed, added also for invalid

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
orestis-ms: Fixed. Should this also be done in test_sqlquery.py ?

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
orestis-ms: Added

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
orestis-ms: fixed (removed)

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
orestis-ms: Since we have additional tests for that, I tried to simplify the test by removing the additional test on the DF format.

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
orestis-ms: Fixed

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
orestis-ms: fixed

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
orestis-ms: Done

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
orestis-ms: SparkStoreCommand inherits from Command. assert_any_call and similar functions throw an exception when comparing instances of the two. Haven't checked if this is bug in nose or expected behavior.

sparkmagic/sparkmagic/utils/configuration.py
orestis-ms: clarity, independence?

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
aggFTW: Hmmm... didn't understand this. Want to chat on Monday?

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
aggFTW: I'm saying that we don't need to check that the run_cell_method is called and only that the sparkmagicsbase `execute_spark` method is being called with the right parameters. This unit test is here only to check that parsing is correct.

sparkmagic/sparkmagic/tests/test_remotesparkmagics.py
aggFTW: I think that there should be unit tests that test only what this class does. Since this class is not actually executing anything against the spark_controller when running spark (sparkmagicsbase is), this should not test an exception there. Instead, it should just test that the exception is caught when it is thrown by `execute_spark`. If this is not clear, we can discuss on Monday.

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
aggFTW: Yes

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,246 @@
...
+                  Command(u'for {} in {}.toJSON().take(120): print({}.encode("{}"))'\
+                          .format(LONG_RANDOM_VARIABLE_NAME, variable_name,
+                                  LONG_RANDOM_VARIABLE_NAME, conf.pyspark_python_encoding())))
+    assert_equals(sparkcommand._scala_command(variable_name),
+                  Command(u'{}.toJSON.take(120).foreach(println)'.format(variable_name)))
+
+
+@with_setup(_setup, _teardown)
+def test_scala_livy_options_spark2():
+    variable_name = "var_name"
```
aggFTW: Where is Spark 2 here?

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,246 @@
...
+
+    sparkcommand = SparkStoreCommand(variable_name, samplemethod='sample', samplefraction=0.33, maxrows=3234)
+    assert_equals(sparkcommand._scala_command(variable_name),
+                  Command('{}.toJSON.sample(false, 0.33).take(3234).foreach(println)'.format(variable_name)))
+
+def test_sample_method_none_all_langs():
+    variable_name = "var_name"
+
+    sparkcommand = SparkStoreCommand(variable_name, samplemethod=None, maxrows=100)
+
```
aggFTW: Too many blank lines.

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
```diff
@@ -0,0 +1,246 @@
...
+    sparkcommand = SparkStoreCommand(variable_name, samplemethod='sample', samplefraction=0.25, maxrows=-1)
+    assert_equals(sparkcommand._scala_command(variable_name),
+                  Command('{}.toJSON.sample(false, 0.25).collect.foreach(println)'.format(variable_name)))
+
+    sparkcommand = SparkStoreCommand(variable_name, samplemethod='sample', samplefraction=0.33, maxrows=3234)
+    assert_equals(sparkcommand._scala_command(variable_name),
+                  Command('{}.toJSON.sample(false, 0.33).take(3234).foreach(println)'.format(variable_name)))
+
+def test_sample_method_none_all_langs():
```
aggFTW: Can this just be added to the different language tests directly?

sparkmagic/sparkmagic/utils/configuration.py
aggFTW: Hmm... I don't think it's needed. If you want, you can call it `dataframe_python_encoding`. Clarity is OK, but I don't see how independence is ever needed for this, do you?

sparkmagic/sparkmagic/utils/utils.py
```diff
@@ -53,6 +56,29 @@ def coerce_pandas_df_to_numeric_datetime(df):
             except (ValueError, TypeError):
                 pass
 
+
+def records_to_dataframe(records_text, kind):
```
aggFTW: Do we have tests for this? Should we move them to the utils test file?

orestis-ms: I'm noticing that there no tests for records_to_dataframe, but only for the sub-routine coerce_pandas_df_to_numeric_datetime. The latter is tested extensively, but not the former.

aggFTW: 😞 It seems like we were indirectly testing that function by having empty content, invalid content, and etcetera go through the function.

It's OK for now to leave it untested. Please just create an issue specifying that the function has no unit tests.

sparkmagic/sparkmagic/tests/test_sparkstorecommand.py
aggFTW: You don't need Spark 2 tests because you are not referencing the name of the sql context, which changes from spark 1 to spark 2.

aggFTW:
So, TODO:

- [ ] Did you test it on an actual notebook? Can you list the tests you performed?
- [ ] Create an issue on missing JSON to dataframe unit tests
- [ ] Bump library version

Do we have anything else?

autovizwidget/autovizwidget/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '0.10.4'
+__version__ = '0.10.5'
```
aggFTW: Please bump to 0.11.0

You can read about semantic versioning here: http://semver.org/
Since we haven't reached 1.0 yet, we are bumping minor version whenever we introduce a new feature, and bumping build version whenever we make a bug fix.

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
```diff
@@ -0,0 +1,120 @@
+from sparkmagic.utils.utils import coerce_pandas_df_to_numeric_datetime, records_to_dataframe
```
msftristew: Nit: Rename file to `sparkstorecommand.py`?

sparkmagic/sparkmagic/livyclientlib/sparkcommand.py
```diff
@@ -0,0 +1,120 @@
...
+import sparkmagic.utils.configuration as conf
+import sparkmagic.utils.constants as constants
+from sparkmagic.utils.sparkevents import SparkEvents
+from sparkmagic.livyclientlib.command import Command
+from sparkmagic.livyclientlib.exceptions import DataFrameParseException, BadUserDataException
+
+import ast
+
+class SparkStoreCommand(Command):
```
msftristew: Structurally this class looks quite a lot like the `SqlQuery` class.  Can/should this be refactored?

orestis-ms: We have discussed offline about this. There are a couple of ways to go. The straight-forward is to refactor so that the new class and SQLQuery inherit from some hybrid. 

However. if we include Command, they are all 'commands', and all (directly or indirectly) inherit from ObjectWithGuid. Hence, one can argue that SQLQuery should inherit from Command.

Nithanaroy:
This is amazing! Is there an example on how to use this to send a Scala Spark variable / dataframe to local Python in Spark kernel?

philmassie:
@Nithanaroy this is a bit late, but I found this works:
```
%%spark -o pdf -n -1
pdf = your_spark_dataframe

pdf.head()
```
(watch out for -n -1)

PR #341, opened 2017-03-23
Update .travis.yml to test on python 3.6 and not 3.3

aggFTW:
3.6 is the official release. Testing to see what would happen to unit tests running on 3.6 via Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
 - '2.7'
+- '3.3'
 - '3.6'
 install:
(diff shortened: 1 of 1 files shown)


aggFTW:
It seems like what's failing in our unit tests is `pandas.to_datetime`. I'll look at this later.

aggFTW:
Note to self, when this is done. update: https://github.com/jupyter-incubator/sparkmagic/blob/python36support/sparkmagic/setup.py#L78

PR #350, opened 2017-04-27
Make location of config.json file configurable using environment variables

jimdowling:
If you want to run more than one jupyter notebook server as the same user with the sparkmagic kernel, you are stuck with a single config file for both notebook server instances. 
This PR will enable users to set a non-default location for the config.json configuration file using one of the environment variables: SPARKMAGIC_CONF_DIR or SPARKMAGIC_CONF_FILE


sparkmagic/sparkmagic/utils/constants.py
```diff
@@ -1,9 +1,11 @@
 # Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
 
+import os
 
-HOME_PATH = "~/.sparkmagic"
-CONFIG_FILE = "config.json"
+
+HOME_PATH = os.environ.get("SPARKMAGIC_CONF_DIR", "~/.sparkmagic")
```
aggFTW: Looks good. You just need to rev the minor version of the 3 packages in the repo:

https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/autovizwidget/__init__.py
https://github.com/jupyter-incubator/sparkmagic/blob/master/hdijupyterutils/hdijupyterutils/__init__.py
https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/__init__.py

aggFTW:
Updated at https://pypi.python.org/pypi/sparkmagic/0.11.3
Thanks for your contribution!

PR #353, opened 2017-05-08
Fix column drop issue when first row has missing value

### LINKED ISSUE #346: SQL Magic %%sql drops column if first row value is null/NaN
It seems that the dataframe visualization for an SQL query will exclude a column outright if the first row contains a `NULL` or `NaN` value for that column.

Consider this example, where the only difference between the queries is the `ORDER` of the results.
image

## Observed with
notebook 4.4.1
sparkmagic 0.11.1
kernel PySpark

## Courtesy Text for Copy/Paste:
```
%%sql
select * from (
  select 1 as foo, 2 as bar
  union all
  select 3, NULL
) x
order by bar asc
```

aggFTW:
Fix #346 

@danfike @edoardovivo


sparkmagic/sparkmagic/utils/utils.py
```diff
@@ -69,7 +69,18 @@ def records_to_dataframe(records_text, kind):
             data_array = data_array[0]
 
         if len(data_array) > 0:
-            df = pd.DataFrame(data_array, columns=data_array[0].keys())
+            df = pd.DataFrame(data_array)
```
jeffersonezra: Move outside if

PR #355, opened 2017-05-10
Kerberos authentication support

### LINKED ISSUE #282: Add Kerberos Support
Need to add kerberos support for spark magic to contact kerberized Livy Server.

Jupyter SparkMagic - Kerberos Authentication Error:
Image of Jupyter SparkMagic - Kerberos Authentication Error

praveenkanamarlapudi:
Fix #282 

1. Enabled kerberos authentication on sparkmagic
2. Updated test cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,12 @@ Request Body example:
                 'username': 'username',
                 'password': 'password',
                 'endpoint': 'url',
+                'auth': 'Kerberos',
                 'kernelname': 'pysparkkernel'
         }
         ```
 
-*Note that the kernelname parameter is optional and defaults to the one specified on the config file or pysparkkernel if not on the config file.*
+*Note that the auth can be either None, Basic_Access or Kerberos based on the authentication enabled in livy. The kernelname parameter is optional and defaults to the one specified on the config file or pysparkkernel if not on the config file.*
 Returns `200` if successful; `400` if body is not JSON string or key is not found; `500` if error is encountered changing clusters.
 
 Reply Body example:
@@ -106,6 +107,18 @@ However, there are some important limitations to note:
 
 2. Since all code is run on a remote driver through Livy, all structured data must be serialized to JSON and parsed by the Sparkmagic library so that it can be manipulated and visualized on the client side.
 In practice this means that you must use Python for client-side data manipulation in `%%local` mode.
+
+## Authentication Methods
+
+Sparkmagic supports:
+
+* No auth
+* Basic authentication
+* Kerberos
+
+Kerberos support is implemented via the [requests-kerberos](https://github.com/requests/requests-kerberos) package. Sparkmagic expects a kerberos ticket to be available in the system. Requests-kerberos will pick up the kerberos ticket from a cache file. For the ticket to be available, the user needs to have run [kinit](https://web.mit.edu/kerberos/krb5-1.12/doc/user/user_commands/kinit.html) to create the kerberos ticket.
+
+Currently, sparkmagic does not support passing a kerberos principal/token, but we welcome pull requests.
 
 ## Contributing
 
--- a/sparkmagic/setup.py
+++ b/sparkmagic/setup.py
@@ -88,6 +88,7 @@ setup(name=NAME,
           'ipykernel>=4.2.2,<5',		
           'ipywidgets>5.0.0,<7.0',
           'notebook>=4.2,<5.0',
-          'tornado>=4'
+          'tornado>=4',
+          'requests_kerberos'
       ])
 
--- a/sparkmagic/setup.py
+++ b/sparkmagic/setup.py
@@ -89,6 +89,6 @@ setup(name=NAME,
           'ipywidgets>5.0.0,<7.0',
           'notebook>=4.2,<5.0',
           'tornado>=4',
-          'requests_kerberos'
+          'requests_kerberos>=0.8.0'
       ])
 
--- a/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
+++ b/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
@@ -44,6 +44,8 @@ class AddEndpointWidget(AbstractMenuWidget):
             description='Add endpoint'
         )
 
+        self.auth_type.on_trait_change(self.update_endpoint)
+
         self.children = [self.ipywidget_factory.get_html(value="<br/>", width=widget_width),
                          self.address_widget, self.auth_type, self.user_widget, self.password_widget,
                          self.ipywidget_factory.get_html(value="<br/>", width=widget_width), self.submit_widget]
@@ -59,3 +61,12 @@ class AddEndpointWidget(AbstractMenuWidget):
         # We need to call the refresh method because drop down in Tab 2 for endpoints wouldn't refresh with the new
         # value otherwise.
         self.refresh_method()
+
+    def update_endpoint(self):
+        if self.auth_type.value == constants.NO_AUTH or self.auth_type.value == constants.AUTH_KERBEROS:
+            self.user_widget.layout.display = 'none'
+            self.password_widget.layout.display = 'none'
+        else:
+            self.user_widget.layout.display = 'flex'
+            self.password_widget.layout.display = 'flex'
+
--- a/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
+++ b/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
@@ -33,7 +33,7 @@ class AddEndpointWidget(AbstractMenuWidget):
             value='password',
             width=widget_width
         )
-        self.auth_type = self.ipywidget_factory.get_dropdown(
+        self.auth = self.ipywidget_factory.get_dropdown(
             options={constants.AUTH_KERBEROS: constants.AUTH_KERBEROS, constants.AUTH_BASIC: constants.AUTH_BASIC,
                      constants.NO_AUTH: constants.NO_AUTH},
             description=u"Auth type:"
@@ -44,17 +44,19 @@ class AddEndpointWidget(AbstractMenuWidget):
             description='Add endpoint'
         )
 
-        self.auth_type.on_trait_change(self.update_endpoint)
+        self.auth.on_trait_change(self._show_correct_endpoint_fields)
 
         self.children = [self.ipywidget_factory.get_html(value="<br/>", width=widget_width),
-                         self.address_widget, self.auth_type, self.user_widget, self.password_widget,
+                         self.address_widget, self.auth, self.user_widget, self.password_widget,
                          self.ipywidget_factory.get_html(value="<br/>", width=widget_width), self.submit_widget]
 
         for child in self.children:
             child.parent_widget = self
 
+        self._show_correct_endpoint_fields()
+
     def run(self):
-        endpoint = Endpoint(self.address_widget.value, self.auth_type.value, self.user_widget.value, self.password_widget.value)
+        endpoint = Endpoint(self.address_widget.value, self.auth.value, self.user_widget.value, self.password_widget.value)
         self.endpoints[self.address_widget.value] = endpoint
         self.ipython_display.writeln("Added endpoint {}".format(self.address_widget.value))
 
@@ -62,8 +64,8 @@ class AddEndpointWidget(AbstractMenuWidget):
         # value otherwise.
         self.refresh_method()
 
-    def update_endpoint(self):
-        if self.auth_type.value == constants.NO_AUTH or self.auth_type.value == constants.AUTH_KERBEROS:
+    def _show_correct_endpoint_fields(self):
+        if self.auth.value == constants.NO_AUTH or self.auth.value == constants.AUTH_KERBEROS:
             self.user_widget.layout.display = 'none'
             self.password_widget.layout.display = 'none'
         else:
--- a/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
+++ b/sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
@@ -2,6 +2,7 @@
 # Distributed under the terms of the Modified BSD License.
 from sparkmagic.controllerwidget.abstractmenuwidget import AbstractMenuWidget
 from sparkmagic.livyclientlib.endpoint import Endpoint
+import sparkmagic.utils.constants as constants
 
 
 class AddEndpointWidget(AbstractMenuWidget):
@@ -32,6 +33,11 @@ class AddEndpointWidget(AbstractMenuWidget):
             value='password',
             width=widget_width
         )
+        self.auth_type = self.ipywidget_factory.get_dropdown(
+            options={constants.AUTH_KERBEROS: constants.AUTH_KERBEROS, constants.AUTH_BASIC: constants.AUTH_BASIC,
+                     constants.NO_AUTH: constants.NO_AUTH},
+            description=u"Auth type:"
+        )
 
         # Submit widget
         self.submit_widget = self.ipywidget_factory.get_submit_button(
@@ -39,14 +45,14 @@ class AddEndpointWidget(AbstractMenuWidget):
         )
 
         self.children = [self.ipywidget_factory.get_html(value="<br/>", width=widget_width),
-                         self.address_widget, self.user_widget, self.password_widget,
+                         self.address_widget, self.auth_type, self.user_widget, self.password_widget,
                          self.ipywidget_factory.get_html(value="<br/>", width=widget_width), self.submit_widget]
 
         for child in self.children:
             child.parent_widget = self
 
     def run(self):
-        endpoint = Endpoint(self.address_widget.value, self.user_widget.value, self.password_widget.value)
+        endpoint = Endpoint(self.address_widget.value, self.auth_type.value, self.user_widget.value, self.password_widget.value)
         self.endpoints[self.address_widget.value] = endpoint
         self.ipython_display.writeln("Added endpoint {}".format(self.address_widget.value))
 
--- a/sparkmagic/sparkmagic/kernels/kernelmagics.py
+++ b/sparkmagic/sparkmagic/kernels/kernelmagics.py
@@ -12,6 +12,7 @@ from IPython.core.magic_arguments import argument, magic_arguments
 from hdijupyterutils.utils import generate_uuid
 
 import sparkmagic.utils.configuration as conf
+from sparkmagic.utils import constants
 from sparkmagic.utils.utils import get_livy_kind, parse_argstring_or_throw
 from sparkmagic.utils.sparkevents import SparkEvents
 from sparkmagic.utils.constants import LANGS_SUPPORTED
@@ -349,25 +350,25 @@ class KernelMagics(SparkMagicBase):
     @argument("-u", "--username", type=str, help="Username to use.")
     @argument("-p", "--password", type=str, help="Password to use.")
     @argument("-s", "--server", type=str, help="Url of server to use.")
-    @argument("-t", "--auth_type", type=str, help="Auth type for authentication")
+    @argument("-t", "--auth", type=str, help="Auth type for authentication")
     @_event
     def _do_not_call_change_endpoint(self, line, cell="", local_ns=None):
         args = parse_argstring_or_throw(self._do_not_call_change_endpoint, line)
         username = args.username
         password = args.password
         server = args.server
-        auth_type = args.auth_type
+        auth = args.auth
 
         if self.session_started:
             error = u"Cannot change the endpoint if a session has been started."
             raise BadUserDataException(error)
 
-        self.endpoint = Endpoint(server, auth_type, username, password)
+        self.endpoint = Endpoint(server, auth, username, password)
 
     def refresh_configuration(self):
         credentials = getattr(conf, 'base64_kernel_' + self.language + '_credentials')()
-        (username, password, url, auth_type) = (credentials['username'], credentials['password'], credentials['url'], credentials['auth_type'])
-        self.endpoint = Endpoint(url, auth_type, username, password)
+        (username, password, auth, url) = (credentials['username'], credentials['password'], credentials['auth'], credentials['url'])
+        self.endpoint = Endpoint(url, auth, username, password)
 
     def get_session_settings(self, line, force):
         line = line.strip()
--- a/sparkmagic/sparkmagic/kernels/kernelmagics.py
+++ b/sparkmagic/sparkmagic/kernels/kernelmagics.py
@@ -349,23 +349,25 @@ class KernelMagics(SparkMagicBase):
     @argument("-u", "--username", type=str, help="Username to use.")
     @argument("-p", "--password", type=str, help="Password to use.")
     @argument("-s", "--server", type=str, help="Url of server to use.")
+    @argument("-t", "--auth_type", type=str, help="Auth type for authentication")
     @_event
     def _do_not_call_change_endpoint(self, line, cell="", local_ns=None):
         args = parse_argstring_or_throw(self._do_not_call_change_endpoint, line)
         username = args.username
         password = args.password
         server = args.server
+        auth_type = args.auth_type
 
         if self.session_started:
             error = u"Cannot change the endpoint if a session has been started."
             raise BadUserDataException(error)
 
-        self.endpoint = Endpoint(server, username, password)
+        self.endpoint = Endpoint(server, auth_type, username, password)
 
     def refresh_configuration(self):
         credentials = getattr(conf, 'base64_kernel_' + self.language + '_credentials')()
-        (username, password, url) = (credentials['username'], credentials['password'], credentials['url'])
-        self.endpoint = Endpoint(url, username, password)
+        (username, password, url, auth_type) = (credentials['username'], credentials['password'], credentials['url'], credentials['auth_type'])
+        self.endpoint = Endpoint(url, auth_type, username, password)
 
     def get_session_settings(self, line, force):
         line = line.strip()
--- a/sparkmagic/sparkmagic/livyclientlib/endpoint.py
+++ b/sparkmagic/sparkmagic/livyclientlib/endpoint.py
@@ -1,14 +1,15 @@
 from .exceptions import BadUserDataException
+import sparkmagic.utils.configuration as conf
 
 
 class Endpoint(object):
-    def __init__(self, url, username="", password=""):
+    def __init__(self, url, auth_type=conf.default_livy_endpoint_auth_type(), username="", password=""):
         if not url:
             raise BadUserDataException(u"URL must not be empty")
         self.url = url.rstrip(u"/")
         self.username = username
         self.password = password
-        self.authenticate = False if username == '' and password == '' else True
+        self.auth_type = auth_type
 
     def __eq__(self, other):
         if type(other) is not Endpoint:
--- a/sparkmagic/sparkmagic/livyclientlib/endpoint.py
+++ b/sparkmagic/sparkmagic/livyclientlib/endpoint.py
@@ -1,15 +1,14 @@
 from .exceptions import BadUserDataException
-import sparkmagic.utils.configuration as conf
 
 
 class Endpoint(object):
-    def __init__(self, url, auth_type=conf.default_livy_endpoint_auth_type(), username="", password=""):
+    def __init__(self, url, auth, username="", password=""):
         if not url:
             raise BadUserDataException(u"URL must not be empty")
         self.url = url.rstrip(u"/")
         self.username = username
         self.password = password
-        self.auth_type = auth_type
+        self.auth = auth
 
     def __eq__(self, other):
         if type(other) is not Endpoint:
--- a/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
+++ b/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
@@ -4,13 +4,13 @@ import json
 from time import sleep
 import requests
 from requests_kerberos import HTTPKerberosAuth, REQUIRED
-import subprocess
 
 import sparkmagic.utils.configuration as conf
 from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.utils.constants import MAGICS_LOGGER_NAME
 import sparkmagic.utils.constants as constants
 from sparkmagic.livyclientlib.exceptions import HttpClientException
+from sparkmagic.livyclientlib.exceptions import BadUserConfigurationException
 
 
 class ReliableHttpClient(object):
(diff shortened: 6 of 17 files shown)


aggFTW:
Thanks so much! Nice to see 👍 
Looking into how busy I'm at work, I cannot get to it now, but I will get to the review not later than Friday.

@pranayhasan, I hope this you and your team can get started on this review. Can you help us with the review to make sure it works for your scenario as well? I would suggest taking the commit and trying it out on your system. Feel free to make any code/test comments as well.

Thanks!

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -44,7 +45,7 @@ def session_configs():
 
 @_with_override
 def kernel_python_credentials():
-    return {u'username': u'', u'base64_password': u'', u'url': u'http://localhost:8998'}
+    return {u'username': u'', u'base64_password': u'', u'url': u'http://localhost:8998', u'auth_type': constants.NO_AUTH}
```
aggFTW: For this to be back compatible with old config.json files, we need to move the logic of:

`constants.NO_AUTH if username == '' and password == '' else constants.BASIC_AUTH`

sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -12,19 +13,19 @@ def _setup():
...
                                      conf.status_sleep_seconds.__name__: 1 })
     assert_equals(conf.status_sleep_seconds(), 1)
-    assert_equals(conf.base64_kernel_python_credentials(), { 'username': 'U', 'password': 'password', 'url': 'L' })
+    assert_equals(conf.base64_kernel_python_credentials(), { 'username': 'U', 'password': 'password', 'url': 'L', 'auth_type': AUTH_BASIC })
 
 
 @with_setup(_setup)
 def test_configuration_override_fallback_to_password():
-    kpc = { 'username': 'U', 'password': 'P', 'url': 'L' }
+    kpc = { 'username': 'U', 'password': 'P', 'url': 'L', 'auth_type': None }
```
aggFTW: We need some extra tests:

1. when auth_type is not there at all in the config file, `conf.base64_kernel_python_credentials` (and all other languages) should return `auth_type` as `constants.NO_AUTH` if the username and password are the empty string
1. when auth_type is not there at all in the config file, `conf.base64_kernel_python_credentials` (and all other languages) should return `auth_type` as `constants.AUTH_BASIC` if the username or password are not the empty string

This will ensure backwards compatibility to existing `config.json` files.

sparkmagic/sparkmagic/tests/test_reliablehttpclient.py
```diff
@@ -2,18 +2,21 @@
 # Distributed under the terms of the Modified BSD License.
 
 from mock import patch, PropertyMock, MagicMock
-from nose.tools import raises, assert_equals, with_setup
+from nose.tools import raises, assert_equals, with_setup, assert_is_not_none
 import requests
+from pandas.util.testing import assertIsInstance
```
aggFTW: Please use `assert isinstance()` instead: https://docs.python.org/2/library/functions.html#isinstance since you are not checking Pandas dataframes here.

sparkmagic/sparkmagic/tests/test_reliablehttpclient.py
```diff
@@ -207,3 +210,18 @@ def test_will_retry_error_no():
             retry_policy.should_retry.assert_called_once_with(None, True, 0)
 
 
+@with_setup(_setup, _teardown)
+def test_basic_auth_check_auth():
```
aggFTW: We need tests for NO_AUTH and for non-valid auth methods.

aggFTW:
@joychak you may want to take a look :)

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
```diff
@@ -32,21 +33,26 @@ def __init__(self, spark_controller, ipywidget_factory, ipython_display, endpoin
             value='password',
             width=widget_width
         )
+        self.auth_type = self.ipywidget_factory.get_dropdown(
+            options={constants.AUTH_KERBEROS: constants.AUTH_KERBEROS, constants.AUTH_BASIC: constants.AUTH_BASIC,
+                     constants.NO_AUTH: constants.NO_AUTH},
+            description=u"Auth type:"
+        )
```
aggFTW: I'm looking at #284, and we had this comment:

>On change, we should hide/show the appropriate fields. So:
>- On change to NO_AUTH: hide username and password.
>- On change to AUTH_KERBEROS: hide username and password.
>- On change to AUTH_SSL: show username and password.

>Ipywidgets has a way to make widgets visible or not:
https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/autovizwidget/widget/encodingwidget.py#L126

sparkmagic/sparkmagic/kernels/kernelmagics.py
```diff
@@ -349,23 +349,25 @@ def _do_not_call_change_language(self, line, cell="", local_ns=None):
     @argument("-u", "--username", type=str, help="Username to use.")
     @argument("-p", "--password", type=str, help="Password to use.")
     @argument("-s", "--server", type=str, help="Url of server to use.")
+    @argument("-t", "--auth_type", type=str, help="Auth type for authentication")
```
aggFTW: Let's change the long name to `--auth` to be consistent with the other magic

sparkmagic/sparkmagic/magics/remotesparkmagics.py
```diff
@@ -112,7 +114,7 @@ def spark(self, line, cell="", local_ns=None):
         # info
```
aggFTW: We need to update the documentation in the comment above to add the `-t` parameter for the `add` subcommand.

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -63,7 +64,7 @@ def base64_kernel_python3_credentials():
 
 @_with_override
 def kernel_scala_credentials():
-    return {u'username': u'', u'base64_password': u'', u'url': u'http://localhost:8998'}
+    return {u'username': u'', u'base64_password': u'', u'url': u'http://localhost:8998', u'auth_type': constants.NO_AUTH}
```
aggFTW: Ditto

aggFTW: We need the same for the other kernels too:

- kernel_r_credentials
- kernel_python3_credentials

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -202,7 +203,7 @@ def _credentials_override(f):
     If 'base64_password' is not set, it will fallback to to 'password' in config.
     """
     credentials = f()
```
aggFTW: This is where you could see if auth_type is there to begin with or not. Please also test it with an actual file.

sparkmagic/sparkmagic/livyclientlib/endpoint.py
```diff
@@ -1,14 +1,15 @@
 from .exceptions import BadUserDataException
+import sparkmagic.utils.configuration as conf
 
 
 class Endpoint(object):
-    def __init__(self, url, username="", password=""):
+    def __init__(self, url, auth_type=conf.default_livy_endpoint_auth_type(), username="", password=""):
```
aggFTW: I don't think we need a default value here, do we? If we don't need it, then we could get rid of the `default_livy_endpoint_auth_type` setting altogether.

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -212,3 +213,8 @@ def _credentials_override(f):
             msg = "base64_password for %s contains invalid base64 string: %s %s" % (f.__name__, exception_type, exception)
             raise BadUserConfigurationException(msg)
     return base64_decoded_credentials
+
+
+@_with_override
+def default_livy_endpoint_auth_type():
```
aggFTW: I think this is not needed.

joychak:
Let me look into this tomorrow or this weekend.

Thanks,
Joy

aggFTW:
Thanks @joychak. I am mostly interested in being able to support your use case presented here: https://spark-summit.org/east-2017/events/secured-kerberos-based-spark-notebook-for-data-science/

I think we could simply add a new authentication method (e.g. PROXY_USER_AUTH or some similar name), and follow the entire Endpoint/HttpClient pattern here in the future.

praveenkanamarlapudi:
Let me look into the comments and update the PR this weekend.

joychak:
I have a general question (after reviewing the pull request) about this approach of Kerberized mechanism for the following 2 scenarios-

Scenario one:
===========
When SparkMagic is being used inside JupyterHub then the kerberos authentication happens at JupyterHub (userid/passwd are authenticated at JupyterHub before opening any notebook) and the token (not any user-id/passwd) is normally passed to the SparkMagic for further authentication. 

Scenario two:
==========
Kerberos authentication can also being done using a keytab file which is very common in corporate environment. When you get authenticated using keytab file (using "KINIT") then you just need to pass the kerberos principal name to the HTTPKerberosAuth object inside "reliablehttpclient.py" file at "_send_request_helper" method. The HTTPKerberosAuth object then picks up the kerberos ticket from the KRB cache file for the given principal.

The above 2 scenarios doesnt need any mechanism within SparkMagic (inside notebook) context to accept userId/passwd.

We need to make sure that we cover these scenarios which is missing currently because the changes in reliablehttpclient.py can only work with UserId/Passwd based Kerberos authentication but not for keytab or KRB-ticket based Kerberos authentication.

sparkmagic/setup.py
```diff
@@ -88,6 +88,7 @@ def version(path):
           'ipykernel>=4.2.2,<5',		
           'ipywidgets>5.0.0,<7.0',
           'notebook>=4.2,<5.0',
-          'tornado>=4'
+          'tornado>=4',
+          'requests_kerberos'
```
aggFTW: +1

Thanks for catching this!

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
```diff
@@ -32,21 +33,26 @@ def __init__(self, spark_controller, ipywidget_factory, ipython_display, endpoin
             value='password',
             width=widget_width
         )
+        self.auth_type = self.ipywidget_factory.get_dropdown(
+            options={constants.AUTH_KERBEROS: constants.AUTH_KERBEROS, constants.AUTH_BASIC: constants.AUTH_BASIC,
```
aggFTW: Let's not use any alpha/dev versions of ipywidgets please.

I was not aware of these display issues with 6.0.0. Is this introduced by this PR or has it always been like this with 6.0.0?

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -11,6 +11,7 @@
 from .constants import HOME_PATH, CONFIG_FILE, MAGICS_LOGGER_NAME, LIVY_KIND_PARAM
 from .utils import get_livy_kind
 from sparkmagic.livyclientlib.exceptions import BadUserConfigurationException
```
aggFTW: We also need to modify the example_config.json file

aggFTW:
@pranayhasan, regarding

>We need to make the authentication type, endpoints configurable from config.json so that there shall be no need to re-enter each time.

I believe this PR already addresses it. Look at the configuration.py file. Do you have something else in mind?

aggFTW:
@joychak Thanks for reviewing!

Those scenarios make sense and are needed. Is there anything on this PR that would prevent those scenarios from being implemented in a later PR? I'm thinking we could have a 4th and 5th authentication type (e.g. `KERBEROS_TOKEN` and `KERBEROS_PRINCIPAL` or something clearer if you have suggestions).

It seems you have implemented at least one of this. Would you be able to submit a PR after this one to add this support or is this design restrictive?

joychak:
Adding KERBEROS_TOKEN and KERBEROS_PRINCIPAL makes sense. I would be able to submit the KERBEROS_PRINCIPAL one immediately after the PR and currently working on the KERBEROS_TOKEN. In case of KERBEROS_TOKEN, the notebook UI should have the provision to provide the keytab file path.

aggFTW:
Thank @joychak! That would be wonderful. For the path, could we use a config based approach?

joychak:
+1 => config based approach

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
pranayhasan: What I meant is that we need to the functionality with 7.0.0a2, so that we can use 7.0.0 once it  gets published. We already have display issues with 6.0.0 version where the fields are displayed in a row as above. Can we revert back the commit and address UI issue to output the fields in columns? We can use the password widget only once we fix these.

pranayhasan:
> I believe this PR already addresses it. Look at the configuration.py file. Do you have something else in mind?

@aggFTW In a corporate environment where all the users need to use Kerberos by default, we don't want to show users the authentication type. It should be defaulted to Kerberos with no display of auth_type in widget. So, we should be able to handle that using **auth**/**auth_type** by specifying in  **~/.sparkmagic/config.json**

aggFTW:
> we don't want to show users the authentication type. It should be defaulted to Kerberos with no display of auth_type in widget. So, we should be able to handle that using auth/auth_type by specifying in ~/.sparkmagic/config.json

You want to add a config to not display the authentication type with the regular python kernel widget magics and instead use the default `default_livy_endpoint_auth_type` specified in the config (I previously asserted this was not needed, but this change would make it needed).

I think this could be a separate PR altogether. @pranayhasan would you mind sending that out if @praveenkanamarlapudi  does not tackle it?

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
aggFTW: Yes, we need to fix the issue with 6.0 first. I just created an issue for it (#358). I do not think I can prioritize it now. @pranayhasan do you want to take a crack at it?

pranayhasan:
@aggFTW Yes, sure.

praveenkanamarlapudi:
Let me look into the comments and update the PR soon.

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
praveenkanamarlapudi: @aggFTW Can you help me understand how to use that API? Any examples are appreciated.

aggFTW: Line number changed: https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/autovizwidget/widget/encodingwidget.py#L134

Basically, you can take any widget and change its `.layout.display` property to `flex` or `none` to show/hide. You would have to create a callback that is called whenever the value changes and knows to show/hide the appropriate widgets. Here, we add a callback, for example:

```
y_column_view = self.ipywidget_factory.get_dropdown(options=options_y_view,
                                                            description="Y", value=self.encoding.y)
y_column_view.on_trait_change(self._y_changed_callback, 'value')
```

praveenkanamarlapudi: Ok thanks. Let me check and update the PR.

aggFTW:
@praveenkanamarlapudi, any updates?

praveenkanamarlapudi:
@aggFTW Sorry I was overloaded with other stuff. Let me update the PR by this weekend. Thanks

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
pranayhasan: @aggFTW Took this up. Thanks for creating an issue!

meatheadmike:
Is there anything left on this one before the PR gets approved? I'd love to start playing with this build!

praveenkanamarlapudi:
@aggFTW  can you please review the PR.

aggFTW:
Reviewing

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
```diff
@@ -59,3 +61,12 @@ def run(self):
         # We need to call the refresh method because drop down in Tab 2 for endpoints wouldn't refresh with the new
         # value otherwise.
         self.refresh_method()
+
+    def update_endpoint(self):
```
aggFTW: Let's rename this method to `_show_correct_endpoint_fields` and I believe that its signature should be: `self, name, old_value, new_value` from looking at other callbacks: https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/autovizwidget/widget/encodingwidget.py#L115

README.md
```diff
@@ -68,12 +68,14 @@ Request Body example:
                 'username': 'username',
                 'password': 'password',
                 'endpoint': 'url',
+                'auth': 'Kerberos',
                 'kernelname': 'pysparkkernel'
         }
         ```
 
-*Note that the kernelname parameter is optional and defaults to the one specified on the config file or pysparkkernel if not on the config file.*
+*Note: The auth can be either None, Basic_Access or Kerberos based on the authentication enabled in livy. kernelname parameter is optional and defaults to the one specified on the config file or pysparkkernel if not on the config file.*
```
aggFTW: This is good for the README section on the server extension, but can you add a bullet point to the Features section above?

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -24,6 +24,8 @@ def __init__(self, endpoint, headers, retry_policy):
             self._auth = HTTPKerberosAuth(mutual_authentication=REQUIRED)
         elif self._endpoint.auth_type == constants.AUTH_BASIC:
             self._auth = (self._endpoint.username, self._endpoint.password)
+        elif self._endpoint.auth_type != constants.NO_AUTH:
+            raise ValueError("Unsupported auth_type.")
```
aggFTW: Please replace ValueError for Bad Config exception at https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/exceptions.py#L41

sparkmagic/sparkmagic/serverextension/handlers.py
```diff
@@ -38,6 +39,12 @@ def post(self):
             username = self._get_argument_or_raise(data, 'username')
             password = self._get_argument_or_raise(data, 'password')
             endpoint = self._get_argument_or_raise(data, 'endpoint')
+            auth_type = self._get_argument_if_exists(data, 'auth_type')
```
aggFTW: When you documented the API, you specified the field as `auth`, not `auth_type`. Please change the code here to `auth`.

sparkmagic/sparkmagic/tests/test_reliablehttpclient.py
```diff
@@ -2,18 +2,21 @@
 # Distributed under the terms of the Modified BSD License.
 
 from mock import patch, PropertyMock, MagicMock
-from nose.tools import raises, assert_equals, with_setup
+from nose.tools import raises, assert_equals, with_setup, assert_is_not_none, assert_false, assert_true
 import requests
+from pandas.util.testing import assertIsInstance
```
aggFTW: You are still using `assertIsInstance`. Previous comment:

>Please use `assert isinstance()` instead: https://docs.python.org/2/library/functions.html#isinstance since you are not checking Pandas dataframes here.

README.md
aggFTW: Actually, a full section on Kerberos auth is needed, as mentioned in a previous comment. Please add the note you have below regarding ticket management to that section, and explain that it is simply using the requests-kerberos package. Basically, give an overview of what type of kerberos authentications are supported vs which ones are not so that users know what to expect and how to configure the system.

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -3,10 +3,13 @@
 import json
 from time import sleep
 import requests
+from requests_kerberos import HTTPKerberosAuth, REQUIRED
+import subprocess
```
aggFTW: Remove subprocess

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -202,7 +203,7 @@ def _credentials_override(f):
     If 'base64_password' is not set, it will fallback to to 'password' in config.
     """
     credentials = f()
-    base64_decoded_credentials = {k: credentials.get(k) for k in ('username', 'password', 'url')}
+    base64_decoded_credentials = {k: credentials.get(k) for k in ('username', 'password', 'url', 'auth_type')}
```
aggFTW: What happens if `auth_type` isn't there? Wouldn't this throw?

sparkmagic/sparkmagic/tests/test_configuration.py
aggFTW: This is still not being tested. Please add tests for when `auth_type` is not part of the configuration specified by the user.

sparkmagic/sparkmagic/tests/test_handlers.py
```diff
@@ -62,7 +64,7 @@ def setUp(self):
 
         # Mock request
         self.request = MagicMock()
-        self.request.body = json.dumps({"path": self.path, "username": self.username, "password": self.password, "endpoint": self.endpoint})
+        self.request.body = json.dumps({"path": self.path, "username": self.username, "password": self.password, "endpoint": self.endpoint, "auth_type": self.auth_type})
```
aggFTW: There's no tests here to test for when `auth` is not in the payload at all. There should be two tests for that case:

1. `auth` is not on payload but due to empty username and password, auth should default to None.
1. `auth` is not on payload but due to non-empty username and password, auth should default to Basic Auth.

aggFTW:
@praveenkanamarlapudi Please look at all comments.
Also please be aware of this PR: https://github.com/jupyter-incubator/sparkmagic/pull/362
which will have collisions with yours.

praveenkanamarlapudi:
@aggFTW Update code. Sorry we had differences with our branch, so missed to push few changes earlier. I will update the README tomorrow evening.

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
praveenkanamarlapudi: I believe adding old_value, new_value are optional and the values can be accessed in the callback as well.

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
```diff
@@ -32,24 +33,42 @@ def __init__(self, spark_controller, ipywidget_factory, ipython_display, endpoin
...
+
         self.children = [self.ipywidget_factory.get_html(value="<br/>", width=widget_width),
-                         self.address_widget, self.user_widget, self.password_widget,
+                         self.address_widget, self.auth, self.user_widget, self.password_widget,
                          self.ipywidget_factory.get_html(value="<br/>", width=widget_width), self.submit_widget]
 
         for child in self.children:
             child.parent_widget = self
 
+        self._show_correct_endpoint_fields()
```
praveenkanamarlapudi: Calling initially to display right parameters initially as per the auth type.

sparkmagic/sparkmagic/kernels/kernelmagics.py
```diff
@@ -349,23 +350,32 @@ def _do_not_call_change_language(self, line, cell="", local_ns=None):
...
             raise BadUserDataException(error)
 
-        self.endpoint = Endpoint(server, username, password)
+        self.endpoint = Endpoint(server, auth_type, username, password)
 
     def refresh_configuration(self):
         credentials = getattr(conf, 'base64_kernel_' + self.language + '_credentials')()
         (username, password, url) = (credentials['username'], credentials['password'], credentials['url'])
-        self.endpoint = Endpoint(url, username, password)
+        if hasattr(credentials, 'auth'):
```
aggFTW: credentials is a dictionary, so the check would be `if 'auth' in credentials`

aggFTW: This check would not be here. When you call conf.base64_kernel_X_credentials(), the contract is that auth is always returned. So this code that sets auth if it's not specified by the user needs to be in configuration.py

aggFTW: Please add unit tests for this code in test_configuration.py. This back compat is extremely important to get right.

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -17,6 +19,13 @@ def __init__(self, endpoint, headers, retry_policy):
         self._endpoint = endpoint
         self._headers = headers
         self._retry_policy = retry_policy
+        if self._endpoint.auth == constants.AUTH_KERBEROS:
+            self._auth = HTTPKerberosAuth(mutual_authentication=REQUIRED)
+        elif self._endpoint.auth == constants.AUTH_BASIC:
+            self._auth = (self._endpoint.username, self._endpoint.password)
+        elif self._endpoint.auth != constants.NO_AUTH:
+            raise ValueError(u"Unsupported auth."+self._endpoint.auth)
```
aggFTW: This is still ValueError when I asked to change it to Bad Config exception at https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/exceptions.py#L41

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
```diff
@@ -32,24 +33,42 @@ def __init__(self, spark_controller, ipywidget_factory, ipython_display, endpoin
...
-        endpoint = Endpoint(self.address_widget.value, self.user_widget.value, self.password_widget.value)
+        endpoint = Endpoint(self.address_widget.value, self.auth.value, self.user_widget.value, self.password_widget.value)
         self.endpoints[self.address_widget.value] = endpoint
         self.ipython_display.writeln("Added endpoint {}".format(self.address_widget.value))
 
         # We need to call the refresh method because drop down in Tab 2 for endpoints wouldn't refresh with the new
         # value otherwise.
         self.refresh_method()
+
+    def _show_correct_endpoint_fields(self):
```
aggFTW: Did you try this signature on trait change? Did it work? I would have expected it to throw and for the fields not to be updated if the signature of the method were wrong.

aggFTW:
Please review my comments again. Let's be thorough on reviewing comments and addressing them.
I will review when you have addressed all comments and updated the README too.

sparkmagic/sparkmagic/controllerwidget/addendpointwidget.py
praveenkanamarlapudi: I tested it and works fine.

README.md
```diff
@@ -107,6 +108,9 @@ However, there are some important limitations to note:
 2. Since all code is run on a remote driver through Livy, all structured data must be serialized to JSON and parsed by the Sparkmagic library so that it can be manipulated and visualized on the client side.
 In practice this means that you must use Python for client-side data manipulation in `%%local` mode.
 
+## Kerberos support
+Sparkmagic uses [requests-kerberos](https://github.com/requests/requests-kerberos) package for kerberos support. The sparkmagic expects kerberos ticket to be avaibale in the system. It picks up the kerbros ticket from cache file. User needs to run kinit to create the kerberos ticket. The sparkmagic doesn't support passing kerberos principal/token.
```
aggFTW: Can we please change this paragraph to:

## Authentication Methods

Sparkmagic supports:

* No auth
* Basic authentication
* Kerberos

Kerberos support is implemented via the [requests-kerberos](https://github.com/requests/requests-kerberos) package. Sparkmagic expects a kerberos ticket to be available in the system. Requests-kerberos will pick up the kerberos ticket from a cache file. For the ticket to be available, the user needs to have run [kinit](https://web.mit.edu/kerberos/krb5-1.12/doc/user/user_commands/kinit.html) to create the kerberos ticket.

Currently, sparkmagic does not support passing a kerberos principal/token, but we welcome pull requests.

praveenkanamarlapudi: Sure.

sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -13,19 +13,43 @@ def _setup():
...
+
+@with_setup(_setup)
+def test_configuration_auth_missing_basic_auth():
+    kpc = { 'username': 'U', 'password': 'P', 'url': 'L'}
+    overrides = { conf.kernel_python_credentials.__name__: kpc }
+    conf.override_all(overrides)
+    conf.override(conf.status_sleep_seconds.__name__, 1)
+    assert_equals(conf.d, { conf.kernel_python_credentials.__name__: kpc,
+                                     conf.status_sleep_seconds.__name__: 1 })
+    assert_equals(conf.status_sleep_seconds(), 1)
```
aggFTW: Great to see these tests! 😄 Thanks for adding them!
Why are you `overriding status_sleep_seconds` here?

sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -13,19 +13,43 @@ def _setup():
...
+    assert_equals(conf.base64_kernel_python_credentials(), { 'username': 'U', 'password': 'P', 'url': 'L', 'auth': AUTH_BASIC })
+
+
+@with_setup(_setup)
+def test_configuration_auth_missing_no_auth():
+    kpc = { 'username': '', 'password': '', 'url': 'L'}
+    overrides = { conf.kernel_python_credentials.__name__: kpc }
+    conf.override_all(overrides)
+    conf.override(conf.status_sleep_seconds.__name__, 1)
```
aggFTW: Same, why override `status_sleep_seconds` at all?

praveenkanamarlapudi: Let me correct it.

praveenkanamarlapudi: I thought it will be like extra check. I can remove it if not needed.

sparkmagic/sparkmagic/tests/test_handlers.py
```diff
@@ -62,7 +64,7 @@ def setUp(self):
 
         # Mock request
         self.request = MagicMock()
-        self.request.body = json.dumps({"path": self.path, "username": self.username, "password": self.password, "endpoint": self.endpoint})
```
aggFTW: Please add two tests where you don't send the `auth` parameter at all and still get no auth and basic auth as values for auth.

praveenkanamarlapudi: Ok

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -212,9 +212,10 @@ def _credentials_override(f):
             exception_type, exception, traceback = sys.exc_info()
             msg = "base64_password for %s contains invalid base64 string: %s %s" % (f.__name__, exception_type, exception)
             raise BadUserConfigurationException(msg)
+    if base64_decoded_credentials['auth'] is None:
```
aggFTW: Awesome! 👍

sparkmagic/sparkmagic/tests/test_configuration.py
aggFTW: Yes, I think the extra check is not needed. Thank you!

sparkmagic/sparkmagic/tests/test_handlers.py
praveenkanamarlapudi: Done! 👍

aggFTW:
Ready for merge! 👍 Thanks so much @praveenkanamarlapudi. I hope a lot of people are going to find this very useful!

@joychak, we would love to see those code reviews for other authentication methods 😃

ghoshtir:
I am trying to get Kerberos authentication working between livy 0.3 (in a Linux edge node of a cluster) and spakmagic 0.12.5 as part of jupyter notebook running on my windows box. On the windows side I am happy to kinit (I don't see the username/password on the UI when I select auth-type as Kerberos). Can you please confirm if the versions mentioned will allow Kerberos authentication between jupyter and livy?

PR #361, opened 2017-05-24
Adding a working Docker setup for developing sparkmagic

apetresc:
It includes the Jupyter notebook as well as the Livy+Spark endpoint.
Documentation is in the README.

Tested and works on my own machine (which, since this is Docker, means it should work anywhere). You just `docker-compose build && docker-compose up` and then launch http://localhost:8888 and add http://spark:8998 as a Livy endpoint from `%manage_spark`.

### DIFF
--- a/Dockerfile.jupyter
+++ b/Dockerfile.jupyter
@@ -6,6 +6,7 @@ RUN pip install sparkmagic
 
 RUN mkdir /home/$NB_USER/.sparkmagic
 COPY sparkmagic/example_config.json /home/$NB_USER/.sparkmagic/config.json
+RUN sed -i 's/localhost/spark/g' /home/$NB_USER/.sparkmagic/config.json
 RUN jupyter nbextension enable --py --sys-prefix widgetsnbextension
 RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/pyspark3kernel
 RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/sparkrkernel
--- a/Dockerfile.jupyter
+++ b/Dockerfile.jupyter
@@ -19,6 +19,8 @@ RUN mkdir /home/$NB_USER/.sparkmagic
 COPY sparkmagic/example_config.json /home/$NB_USER/.sparkmagic/config.json
 RUN sed -i 's/localhost/spark/g' /home/$NB_USER/.sparkmagic/config.json
 RUN jupyter nbextension enable --py --sys-prefix widgetsnbextension
+RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/sparkkernel
+RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/pysparkkernel
 RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/pyspark3kernel
 RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/sparkrkernel
 RUN jupyter serverextension enable --py sparkmagic
--- a/Dockerfile.jupyter
+++ b/Dockerfile.jupyter
@@ -1,0 +1,16 @@
+FROM jupyter/base-notebook:d0b2d159cc6c
+
+USER $NB_USER
+
+RUN pip install sparkmagic
+
+RUN mkdir /home/$NB_USER/.sparkmagic
+COPY sparkmagic/example_config.json /home/$NB_USER/.sparkmagic/config.json
+RUN jupyter nbextension enable --py --sys-prefix widgetsnbextension
+RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/pyspark3kernel
+RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/sparkrkernel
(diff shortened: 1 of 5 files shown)


Dockerfile.jupyter
```diff
@@ -0,0 +1,17 @@
...
+
+USER $NB_USER
+
+RUN pip install sparkmagic
+
+RUN mkdir /home/$NB_USER/.sparkmagic
+COPY sparkmagic/example_config.json /home/$NB_USER/.sparkmagic/config.json
+RUN sed -i 's/localhost/spark/g' /home/$NB_USER/.sparkmagic/config.json
+RUN jupyter nbextension enable --py --sys-prefix widgetsnbextension
+RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/pyspark3kernel
```
aggFTW: I think we should be installing:
* sparkkernel
* pysparkkernel
* pyspark3kernel
* sparkrkernel (?)

Is R properly set up for Livy on the image? I'll try it as soon as I'm done with the review

Dockerfile.spark
```diff
@@ -0,0 +1,29 @@
...
+
+RUN apt-get update && apt-get install -yq --no-install-recommends --force-yes \
+    git \
+    openjdk-7-jdk \
+    maven && \
+    rm -rf /var/lib/apt/lists/*
+
+ENV LIVY_BUILD_VERSION livy-server-0.3.0
+ENV LIVY_APP_PATH /apps/$LIVY_BUILD_VERSION
+ENV LIVY_BUILD_PATH /apps/build/livy
```
aggFTW: Something similar to this:

```
ANACONDA_DEST_PATH=/usr/bin/anaconda
CONDA_PATH=$ANACONDA_DEST_PATH/bin/conda

# TODO wget anaconda to ANACONDA_DEST

bash $INSTALLER_PATH -p $ANACONDA_DEST_PATH -b -f

# Create an env for Python 3.
$CONDA_PATH create -n py35 python=3.5 anaconda
```

README.md
```diff
@@ -54,6 +54,27 @@ See [Pyspark](examples/Pyspark Kernel.ipynb) and [Spark](examples/Spark Kernel.i
...
+[Docker Compose](https://docs.docker.com/compose/) both installed, and
+then simply run:
+
+    docker-compose build
+    docker-compose up
+
+You will then be able to access the Jupyter notebook in your browser at
+http://localhost:8888. Inside this notebook, you can configure a
+sparkmagic endpoint at http://spark:8998. This endpoint is able to
+launch both Scala and Python sessions.
```
aggFTW: We should also mention that the managed sparkmagic kernels will be available.

Dockerfile.jupyter
aggFTW: Yeah, so R is not set up properly for Livy. Let's disable installing the SparkR kernel.

README.md
aggFTW: All except SparkR.

README.md
```diff
@@ -54,6 +54,27 @@ See [Pyspark](examples/Pyspark Kernel.ipynb) and [Spark](examples/Spark Kernel.i
...
+extensions installed, and a Livy server backed by a local-mode Spark instance.
+(This is just for testing and developing sparkmagic itself; in reality,
+sparkmagic is not very useful if your Spark instance is on the same machine!)
+
+In order to use it, make sure you have [Docker](https://docker.com) and
+[Docker Compose](https://docs.docker.com/compose/) both installed, and
+then simply run:
+
+    docker-compose build
+    docker-compose up
```
aggFTW: I think it would also be a good idea to add instructions for exiting...

`Ctrl-C` + `docker-compose down`?

Dockerfile.jupyter
apetresc: Fixed, by setting up R properly. Tested and works!

README.md
apetresc: Done

README.md
apetresc: Done

Dockerfile.spark
apetresc: I've made the other changes, but I'm not sure I understand the need for this one. The base image I'm using for the Livy container (`gettyimages/spark:2.1.0-hadoop-2.7`) already has a Python3 installation which seems to be working just fine with Livy and sparkmagic as-is.

Do you just mean creating a virtualenv for it? I can do that, though I'm not sure it's necessary if we're just running inside a single-purpose container.

aggFTW: Sorry I wasn't clear. I meant we need to install the following two kernels, besides the kernels already being installed:

* pysparkkernel
* sparkkernel

You can just add `sparkkernel` to the list of kernels to install, and it should just work.

`pysparkkernel` is different in that it should run in a python 2.7 environment whereas `pyspark3kernel` would run in a python 3 environment. Because the container's python installation is python 3, we need to tell livy where to find the python 2 installation in the image (created via virtualenv or anaconda or something else). The way livy finds the two installations is via the env variables i mentioned above.

The point would be to try to show users that one can have two python versions running fine side by side and how to set it up. This is some extra work, and I would be OK with you just adding `sparkkernel` for now and me creating an issue to add `pysparkkernel` to the docker image to tackle later.

aggFTW:
Thanks for the updates! If you can just add `sparkkernel` and do something about `pysparkkernel`, whereas create an issue or create a python 2 virtualenv or similar, I think we'd be ready for a merge.

Thanks for your contribution!

Dockerfile.spark
apetresc: Ah, I see what you mean then :) Okay, let me add that real quick.

apetresc:
Comments addressed :) Debian has separate packages+environments for python2 and python3, so it was just a matter of making sure both are installed and that `PYSPARK_PYTHON` and `PYSPARK3_PYTHON` were pointing at the right ones.

I confirmed manually that they are in fact in separate environments and the right kernel is pointing to the right version.

aggFTW:
Thanks for the awesome contribution! I know a lot of people will find it very useful.

PR #362, opened 2017-06-02
Automatically configure wrapper-kernel endpoints in widget

### LINKED ISSUE #359: Automatically add endpoint for the one configured in kernel_*_credentials
If you set a Livy server in `kernel_scala_credentials` or `kernel_python_credentials`, it works automatically when running the corresponding wrapper kernel. But if you start a normal Python notebook on the same Jupyter instance and run the `%manage_spark` magic, it doesn't take that configuration into account at all; you have to re-enter it before you can create a session.

Is there any good reason not to pre-configure an endpoint entry if the user has already gone to the trouble of setting it up for the wrapper kernel? Of course, they still can specify other endpoints in `%manage_spark` but this seems like a convenient shortcut for 90% of use cases (with very little damage done to the other 10%).

Would a pull request along these lines be accepted, or are you opposed to this idea? Thanks :)

apetresc:
This resolves #359. Basically, it pre-populates the endpoint widgets with any valid configurations found in the `~/.sparkmagic/config.json` for any of the wrapper kernels. This just removes a step from the workflow in cases where you've set up `config.json` for any other kernels.

Tested and works in the Docker environment, and my own deployment :smile:


sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
```diff
@@ -5,21 +5,39 @@
...
 
         self._refresh()
 
     def run(self):
         pass
 
+    @staticmethod
+    def _get_default_endpoints():
+        default_endpoints = set()
+        kernel_types = ['python', 'python3', 'scala', 'r']
```
aggFTW: This list should be in the constants file.

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
```diff
@@ -5,21 +5,39 @@
...
+
+        for kernel_type in kernel_types:
+            endpoint_config = getattr(conf, 'kernel_%s_credentials' % kernel_type)()
```
aggFTW: The defaults will return `localhost:8998` on URL, making this check to always add the endpoints, even when not configured. I'm trying to come up with a way to tell if the user configured the kernel endpoints or not for sure and failing to....

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
 from sparkmagic.controllerwidget.abstractmenuwidget import AbstractMenuWidget
-
+from sparkmagic.livyclientlib.exceptions import HttpClientException
```
aggFTW: Please add an extra line here (2 lines in between sections)

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
```diff
@@ -30,7 +30,13 @@ def get_existing_endpoint_widgets(self):
 
             # Endpoints
             for url, endpoint in self.endpoints.items():
-                endpoint_widgets.append(self.get_endpoint_widget(url, endpoint))
+                try:
+                    endpoint_widgets.append(self.get_endpoint_widget(url, endpoint))
+                except HttpClientException:
+                    # If we can't reach one of the default endpoints, just skip over it
+                    # silently
+                    if not endpoint.is_default:
```
aggFTW: Please log the error here too.

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
aggFTW: I think that's why you added the `is_default` field. Is that so?

sparkmagic/sparkmagic/livyclientlib/linearretrypolicy.py
```diff
@@ -11,6 +11,8 @@ def __init__(self, seconds_to_sleep, max_retries):
         self._max_retries = max_retries
 
     def should_retry(self, status_code, error, retry_count):
+        if None in (status_code, retry_count):
```
aggFTW: How could this happen?

sparkmagic/sparkmagic/livyclientlib/endpoint.py
```diff
@@ -2,19 +2,23 @@
...
 class Endpoint(object):
-    def __init__(self, url, username="", password=""):
+    def __init__(self, url, username="", password="", is_default=False):
         if not url:
             raise BadUserDataException(u"URL must not be empty")
         self.url = url.rstrip(u"/")
         self.username = username
         self.password = password
         self.authenticate = False if username == '' and password == '' else True
+        self.is_default = is_default
```
aggFTW: Please add comments as to when `is_default` should be true and when it should be false, and implications for it.

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
apetresc: Yes, exactly. I want to treat errors connecting to the endpoint differently based on whether the endpoint was explicitly added, or just added through this default mechanism. If they added it themselves, throwing an error to stdout is okay, but otherwise I think it should just be silently dropped from the list.

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
apetresc: Okay, I don't mind logging it here too, but I just want to make sure it's a good idea. As you guessed, the problem I'm trying to fix is someone copying in a default `example_config.json` and getting strange errors about an endpoint they aren't even aware of.

I think the current behavior is good: if they _did_ explicitly configure an endpoint in `config.json` that cannot be connected to, they silently won't see it in the default endpoint list – then when they explicitly add it, they see the connection error and understand what needs to be fixed. The alternative seems worse.

sparkmagic/sparkmagic/livyclientlib/linearretrypolicy.py
apetresc: It was happening to me when connecting to an endpoint that didn't exist. `requests.get` would return `None` for `status_code` and stdout would return these strange opaque errors about `NoneType` in `linearretrypolicy.py`, which isn't very descriptive to the end user. I can fetch a screenshot of the situation if you'd like :)

sparkmagic/sparkmagic/livyclientlib/endpoint.py
apetresc: Will do!

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
apetresc: Ah beautiful, thanks!

apetresc:
Thank you for the comments. I'm out most of the day so I won't get a chance to actually commit the changes until later tonight, but I will get to it shortly! :smile:

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
aggFTW: Should we call the field `implicitly_added`?

sparkmagic/sparkmagic/livyclientlib/linearretrypolicy.py
aggFTW: Got it. Thanks!

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
aggFTW: I think logging would be a good idea for the case when it's silently ommited from the list and the user files an issue saying it's being ommited. We will then be able to look at the log file and explain what's going on. Logging shouldn't interfere with what the UX is, since it's just writting to a file/logging system.

aggFTW:
No worries, thanks 😄

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
```diff
@@ -5,21 +5,39 @@
...
+            if all([p in endpoint_config for p in ["url", "password", "username"]]) and endpoint_config["url"] != "":
+                default_endpoints.add(Endpoint(
```
aggFTW: Enpoint now has an auth parameter after the kerberos PR was merged, and the widget should have an extra dropdown now.

aggFTW:
Any news on the PR? I'd like to push sparkmagic 0.12 soon with this goodness on it 😃 
If not possible, that's fine, and we can always ship it as 0.12.1 or something like that.

aggFTW:
Gentle ping

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
apetresc: Fixed

sparkmagic/sparkmagic/controllerwidget/magicscontrollerwidget.py
apetresc: Renamed

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
apetresc: Whoops, done

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
apetresc: Fair enough. Added a log to that effect. I hope `SparkLogger` was the right thing to use here? It seems to just be a simple subclass of the hdijupyterutils Log object, but the name confused me... made me think it somehow logged things on/about the Spark cluster.

sparkmagic/sparkmagic/livyclientlib/endpoint.py
apetresc: Done.

apetresc:
Sorry again for the delay - was out of town and didn't find time until I got back. All comments addressed, I hope :) Manual tests still look good.

apetresc:
Ah, I will up-merge with master too, one minute...

apetresc:
Upmerged with master. Everything still works, although I've done 0 testing using the new Kerberos `auth` parameter, mainly because I have no idea how Kerberos works :fearful: I can dive into it if needed, it should probably be added to the Docker image anyway if it's going to be supported going forward.

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
aggFTW: No, that's the right one 👍

aggFTW:
Thanks!

Issue #358, opened 2017-05-16
Display issues when adding an endpoint via widget


aggFTW:
Look at 

![display issue with widget](https://cloud.githubusercontent.com/assets/5507845/26034074/6fb3ddc2-38d4-11e7-8829-60c86b8d3612.png)

pranayhasan:
@aggFTW I'm taking up fixing this issue.

pranayhasan:
#368

PR #368, opened 2017-06-15
Widget UI updates, fixing ipywidgets backward incompatible changes, config for adding endpoints and overriding auth

pranayhasan:
* Adding support for Cookies
* Upgraded ipywidgets to use Password Widget
* Setting password blank once user is authenticated
* Update widget width
* Adding authentication related properties. It will read these properties from config.json
* UI enhancements
* Changes to configure Auth Type
* Adding endpoints as a configuration
* Addresses #358


.travis.yml
```diff
@@ -4,6 +4,7 @@ python:
 - '3.3'
 install:
 - pip install six
+- pip install --upgrade pip
```
aggFTW: Why?

autovizwidget/setup.py
```diff
@@ -58,8 +58,8 @@ def version(path):
         'Programming Language :: Python :: 3.4'],
       install_requires=[
           'plotly>=1.10.0,<1.11.0',
-          'ipywidgets>5.0.0,<7.0',
+          'ipywidgets>=7.0.0a4,<8.0',
```
aggFTW: Sorry, but let's not bump to any alpha version whatsoever. It should be `>=6.1.0,<7.0`

sparkmagic.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
```
aggFTW: Please remove all idea files from the review.

pranayhasan:
@aggFTW Thanks! I will break them up and let you know.

aggFTW:
Thanks so much! I will close this review then.

PR #371, opened 2017-06-21
Ability to add custom headers to HTTP calls

### LINKED ISSUE #285: Add ability to specify headers to Livy
I got the following error when try to connect to livy server on HDP 2.5:

HttpClientException: Invalid status code '400' from http://host:8998/sessions with error payload: <html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 400 </title>
</head>
<body>

<h2>HTTP ERROR: 400</h2>

<p>Problem accessing /sessions. Reason:
<pre>    Missing Required Header for CSRF protection.</pre></p>

<hr /><i><small>Powered by Jetty://</small></i>
</body>
</html>

karuppayya:
Closes #285

adding ability to add custom headers to HTTP calls
The following json block has to be added to config.json
```
  "custom_headers": {
    "header1":"value1",
    "header2":"value2",
  }
```

### DIFF
--- a/sparkmagic/sparkmagic/tests/test_configuration.py
+++ b/sparkmagic/sparkmagic/tests/test_configuration.py
@@ -78,10 +78,3 @@ def test_share_config_between_pyspark_and_pyspark3():
     kpc = { 'username': 'U', 'password': 'P', 'base64_password': 'cGFzc3dvcmQ=', 'url': 'L', 'auth': AUTH_BASIC }
     overrides = { conf.kernel_python_credentials.__name__: kpc }
     assert_equals(conf.base64_kernel_python3_credentials(), conf.base64_kernel_python_credentials())
-
-@with_setup(_setup)
-def test_custom_headers():
-    headers = { 'header1': 'value1'}
-    overrides = { conf.custom_headers.__name__: headers }
-    conf.override_all(overrides)
-    assert_equals(conf.custom_headers(), { 'header1': 'value1'})
--- a/sparkmagic/sparkmagic/tests/test_configuration.py
+++ b/sparkmagic/sparkmagic/tests/test_configuration.py
@@ -84,4 +84,4 @@ def test_custom_headers():
     headers = { 'header1': 'value1'}
     overrides = { conf.custom_headers.__name__: headers }
     conf.override_all(overrides)
-    assert_equals(conf.get_custom_headers(), { 'header1': 'value1'})
+    assert_equals(conf.custom_headers(), { 'header1': 'value1'})
--- a/sparkmagic/sparkmagic/tests/test_configuration.py
+++ b/sparkmagic/sparkmagic/tests/test_configuration.py
@@ -78,3 +78,10 @@ def test_share_config_between_pyspark_and_pyspark3():
     kpc = { 'username': 'U', 'password': 'P', 'base64_password': 'cGFzc3dvcmQ=', 'url': 'L', 'auth': AUTH_BASIC }
     overrides = { conf.kernel_python_credentials.__name__: kpc }
     assert_equals(conf.base64_kernel_python3_credentials(), conf.base64_kernel_python_credentials())
+
+@with_setup(_setup)
+def test_custom_headers():
+    headers = { 'header1': 'value1'}
+    overrides = { conf.custom_headers.__name__: headers }
+    conf.override_all(overrides)
+    assert_equals(conf.get_custom_headers(), { 'header1': 'value1'})
--- a/sparkmagic/sparkmagic/tests/test_livyreliablehttpclient.py
+++ b/sparkmagic/sparkmagic/tests/test_livyreliablehttpclient.py
@@ -2,6 +2,8 @@ from mock import MagicMock
 from nose.tools import assert_equals
 
 from sparkmagic.livyclientlib.livyreliablehttpclient import LivyReliableHttpClient
+from sparkmagic.livyclientlib.endpoint import Endpoint
+import sparkmagic.utils.configuration as conf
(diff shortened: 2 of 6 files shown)


aggFTW:
Looks good! Could you add a unit test that checks that custom headers are added when specified please? You can use the override configuration methods. Check other tests for examples.

karuppayya:
@aggFTW Thanks for the quick review. I was checking the code in [test_reliablehttpclient.py](https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/tests/test_reliablehttpclient.py#L59).
Not sure how to get the request headers from the patched obj(Unlike in actual `response` object sent which has a wrapper [request](https://github.com/requests/requests/blob/master/requests/models.py#L585) object containing headers, there is none in the patched obj ). Any pointer might be helpful.

But I have written a basic test to confirm that configuration.py send the request heades as is without modifying.(af027aa)

sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -78,3 +78,10 @@ def test_share_config_between_pyspark_and_pyspark3():
     kpc = { 'username': 'U', 'password': 'P', 'base64_password': 'cGFzc3dvcmQ=', 'url': 'L', 'auth': AUTH_BASIC }
     overrides = { conf.kernel_python_credentials.__name__: kpc }
     assert_equals(conf.base64_kernel_python3_credentials(), conf.base64_kernel_python_credentials())
+
+@with_setup(_setup)
+def test_custom_headers():
```
aggFTW: Sorry if I was not clear, I meant that you should add a unit test to test that the from_endpoint method in LivyReliableHttpClient returns a client with custom headers if some custom headers are specified in the config.

aggFTW:
You don't need to mess with the patched objects or anything. You just need a test that tests the `from_endpoint` method to check that the client will be sending the headers.

karuppayya:
@aggFTW Thanks for the inputs. 
I have added the test to verify that headers are part of the end point.

sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -78,3 +78,10 @@ def test_share_config_between_pyspark_and_pyspark3():
     kpc = { 'username': 'U', 'password': 'P', 'base64_password': 'cGFzc3dvcmQ=', 'url': 'L', 'auth': AUTH_BASIC }
     overrides = { conf.kernel_python_credentials.__name__: kpc }
     assert_equals(conf.base64_kernel_python3_credentials(), conf.base64_kernel_python_credentials())
+
```
aggFTW: Add an extra white line before this method.

sparkmagic/sparkmagic/tests/test_configuration.py
aggFTW: You can remove this test.

sparkmagic/sparkmagic/tests/test_livyreliablehttpclient.py
```diff
@@ -60,3 +63,11 @@ def test_get_all_session_logs():
...
     http_client.get.assert_called_once_with("/sessions/42/log?from=0", [200])
 
+def test_custom_headers():
+    custom_headers = {"header1": "value1"}
+    overrides = { conf.custom_headers.__name__: custom_headers }
+    conf.override_all(overrides)
+    endpoint = Endpoint("http://url.com", constants.NO_AUTH)
+    client = LivyReliableHttpClient.from_endpoint(endpoint)
+    headers = client.get_headers()
+    assert_equals(custom_headers.items() <= headers.items(), True)
```
aggFTW: Please check that the content type header is present explicitly and that the header1 header is also there. And that there is only two headers. You can remove the assert you have now. I think it's too weak of an assertion.

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -244,3 +244,6 @@ def _credentials_override(f):
         base64_decoded_credentials['auth'] = get_auth_value(base64_decoded_credentials['username'], base64_decoded_credentials['password'])
     return base64_decoded_credentials
 
+@_with_override
```
aggFTW: Add an extra white line before this method.

karuppayya:
@aggFTW I have tried to address the review comments, Please let me know if it looks fine.

karuppayya:
I am not sure why CI didnt get triggered

aggFTW:
Hmm... weird that CI build didn't get triggered indeed. Can we try to push a new commit and see if it gets triggered? Maybe add to example_config.json:

 `"custom_headers": { }`

karuppayya:
This time, it got triggered :)

aggFTW:
Thanks for the contribution!

karuppayya:
Thanks for taking time to review my changes 👍 @aggFTW 
Can you please let me know the release cycle followed?
Also i was trying to pip install from git using following command
```
pip install ssh+https://github.com/jupyter-incubator/sparkmagic.git#subdirectory=sparkmagic/sparkmagic
``` 
Bu fails with no matches found, any idea what i am doing wrong here?

aggFTW:
I'll push the 0.12.0 release today.
Haven't installed directly from git, but I'd probably follow the instructions here instead, since you need the 3 packages, in the order specified there: https://github.com/jupyter-incubator/sparkmagic#contributing

aggFTW:
Released now!

karuppayya:
Thanks!!! @aggFTW

PR #374, opened 2017-06-27
Don't ask for write permissions during read

kroq-gar78:
The `read_lines` function only seems to only need read permissions, not write permissions. Requesting read and write permissions (`"r+a"`) causes issues on multi-user systems, where the `config.json` file would require write permissions for all users needing to read it.

See documentation for permissions in `open`:
 * [Python 2.7](https://docs.python.org/2/library/functions.html#open)
 * [Python 3.6](https://docs.python.org/3/library/functions.html#open)

All unit tests are passing after this change.
```
$ nosetests hdijupyterutils autovizwidget sparkmagic
................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 352 tests in 5.072s

OK

```

### DIFF
--- a/hdijupyterutils/hdijupyterutils/filesystemreaderwriter.py
+++ b/hdijupyterutils/hdijupyterutils/filesystemreaderwriter.py
@@ -20,7 +20,7 @@ class FileSystemReaderWriter(object):
 
     def read_lines(self):
         if os.path.isfile(self.path):
-            with open(self.path, "r+") as f:
+            with open(self.path, "r") as f:
                 return f.readlines()
         else:
             return ""


aggFTW:
Thanks! Do you need this released ASAP or can we bulk push with next version? We don't follow a schedule for releases.

kroq-gar78:
If you could release it sooner that would be great!

On Jun 27, 2017 10:54 AM, "Alejandro Guerrero Gonzalez" <
notifications@github.com> wrote:

Thanks! Do you need this released ASAP or can we bulk push with next
version? We don't follow a schedule for releases.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jupyter-incubator/sparkmagic/pull/374#issuecomment-311435578>,
or mute the thread

aggFTW:
Publishing!

kroq-gar78:
Excellent, thank you!

PR #389, opened 2017-07-28
Fix endpoints widget when deleting a session

### LINKED ISSUE #386: Sessions in Manage Endpoints and Manage sessions not in sync
The sessions in manage endpoints tab are not in synch with the sessions in manage sessions tab in widget UI as show below. 

The session 6 has been deleted from manage endpoints as shown below, but the UI continues to display the session in the endpoint list.

screen shot 2017-07-24 at 11 42 56 am

Even the manage sessions UI displays the deleted session as below:
screen shot 2017-07-24 at 11 42 38 am

This is primarily due to the methods that are used in spark controller.

### Method called when deleting session from manage endpoint widget
_Class: Spark Controller_
`delete_session_by_id(endpoint, session_id) -calls-> creates a new LivySession object, say ls with LivyReliableHTTPClient and does ls.delete()`
This method doesn't go through the session manager route and hence the sessions in session manager continue to exist even after being deleted. The status shown in manage sessions tab

### Method called when deleting from manage sessions widget
_Class: Spark Controller_
`delete_session_by_name(name) -calls-> session_manager.delete_client(name)`
This method cleanly deletes the session by removing session from the sessions managed by session manager.

So, to make the experience consistent the method **delete_session_by_id** needs to be updated to go through session manager route.

aggFTW:
Closes #386 

* Bumps to 0.12.3
* Fixes adding the default endpoint (bug was that a None was being passed)
* Adds sample notebooks to docker file for easier testing
* Deletes session from manager if session is being managed

Still missing unit tests. Will add later.


aggFTW:
@pranayhasan if you could please review

sparkmagic/sparkmagic/controllerwidget/manageendpointwidget.py
```diff
@@ -114,6 +114,7 @@ def delete_endpoint(button):
             except ValueError as e:
                 self.ipython_display.send_error(str(e))
                 return
+            self.refresh_method()
```
aggFTW: Cool, thanks for catching this!

PR #403, opened 2017-08-24
Adding default CSRF header as a good security practice.

JeffRodriguez:


aggFTW:
@JeffRodriguez I think that this functionality can easily be set with Custom Headers, without a need to add any new code/tests. I'm inclined to not merge and just leave it to the user to set the headers that they require. This default value might not be needed by some (I know my organization won't use it, for example).

aggFTW:
Hi Jeff, somehow I missed the email? :( I'm sorry! I still think that custom headers is best here.
We had this issue before:
https://github.com/jupyter-incubator/sparkmagic/issues/285

And this PR was pushed for it: https://github.com/jupyter-incubator/sparkmagic/pull/371

I'm closing for now. Please feel free to post a comment if you disagree and we can talk about it.

Thanks!

PR #408, opened 2017-09-18
Bump up supported Jupyter/Notebook versions to the latest supported

apetresc:
Sparkmagic uses version specifiers to cap supported Jupyter/IPython/Notebook server versions to supported/tested versions. Those versions have fallen quite a bit out of date.

I've done a bunch of testing, both on the included Docker Compose set up as well as in my own production cluster, to confirm that sparkmagic works just fine with newer versions of Jupyter, so I've updated those version qualifiers accordingly. You can use the provided Dockerfiles (which have also been bumped to the latest versions) to confirm this.

Here's a screenshot to "prove" it :)
screen shot 2017-09-18 at 3 34 02 pm


aggFTW:
Thanks @apetresc! I'm checking that everything works too :) I'll update if I find anything. In the meantime, can you please also update the `requirements.txt` files please?

apetresc:
@aggFTW Whoops, somehow missed the `requirements.txt` files there. Brought them in line with `setup.py`.

aggFTW:
Hey, I'm actually seeing an error: 
image

I don't have time to debug this right now (maybe not even this week). I can take a look later, but either it doesn't work out of the box or I have some funky setup. I believe that it may be the former, unfortunately.

aggFTW:
I also see this deprecation error:

image

And I'm unable to create a Scala session:

```
spark_1    | 17/09/18 22:43:55 INFO ContextLauncher: 17/09/18 22:43:55 INFO repl.SparkInterpreter: Created Spark sess
ion.
spark_1    | 17/09/18 22:43:55 INFO ContextLauncher: java.util.concurrent.RejectedExecutionException: Task scala.conc
urrent.impl.CallbackRunnable@678404e2 rejected from java.util.concurrent.ThreadPoolExecutor@22c2674b[Shutting down, p
ool size = 1, active threads = 1, queued tasks = 0, completed tasks = 0]
```

apetresc:
1. The "IOPub data rate exceeded" error is something that happens with visualizations on 5.0.0 to everyone, not specific to sparkmagic ([ticket here](https://github.com/jupyter/notebook/issues/2287)). It's fixed with a configuration setting, I can apply that to the Docker image if needed? (It's also mitigated in 5.1.0, but that was only released a few days ago so there isn't a base Docker image for it yet, so this PR is still using 5.0.0 as the test case. A real user doing `pip install notebook` would get 5.1.0 right now, though)
2. I can fix this in `autovizwidget`, although again, some libraries (like a non-latest-version pyplot) will display this regardless of sparkmagic.
3. Whoops! I didn't do any testing of the scala kernel myself, since I never really use Scala. I will look into this!

apetresc:
Pushed fixes/workarounds for the deprecation and iopub problems. Looking into Scala now.

screen shot 2017-09-19 at 2 17 00 pm

autovizwidget/autovizwidget/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '0.12.4'
+__version__ = '0.12.5'
```
aggFTW: You need to bump all libraries together, since we always push all to pip, and build would fail otherwise.

autovizwidget/autovizwidget/widget/utils.py
```diff
@@ -13,7 +13,7 @@ def infer_vegalite_type(data):
     data: Numpy array or Pandas Series
     """
 
-    typ = pd.lib.infer_dtype(data)
+    typ = pd.api.types.infer_dtype(data)
```
aggFTW: Do we need to bump the minor required version of pandas? I don't know what pandas version this api got introduced with.

apetresc: Good catch. Looking at the pandas repository, it seems this change was made in 0.21.0.

Actually I just noticed that autovizwidget doesn't have an explicit dependency on _any_ version of pandas. Presumably it is relying on a transitive dependency through hdijupyterutils. Given that autovizwidget explicitly imports pandas, it seems that a dependency should be added, but I am asking to make sure before I make the change :)

autovizwidget/autovizwidget/__init__.py
apetresc: Done!

autovizwidget/autovizwidget/widget/utils.py
aggFTW: 😢 Yes, it should have an explicit dependency on it and not rely on the transitive dependency, since it's calling it directly. Probably missed adding it at some point back in time.

apetresc:
It seems that the Scala kernel failure was caused by an incompatibility with Spark 2.2.0. Changing the base image of Dockerfile.spark to `gettyimages/spark:2.1.0-hadoop-2.7` resolves the issue.

Obviously we want sparkmagic to be compatible with 2.2.0 as well, but that's outside of the scope of this PR (which is just focused on allowing sparkmagic to co-exist with newer versions of Jupyter), so I'll just leave the Spark version at 2.1.0 where it was originally.

aggFTW:
Cool. It seems that this livy version is not yet compatible with 2.2 then.

apetresc:
Ah yeah, [looks like you're right](https://github.com/apache/incubator-livy/pull/18). I guess we can revisit the Spark bump when Livy 0.4.0 comes out of incubation. Or are `-incubating` releases considered stable in Apache parlance?

autovizwidget/autovizwidget/widget/utils.py
apetresc: Done!

apetresc:
Ah it seems that 0.4.0 *is* considered stable. In that case, I might as well try to implement that here as well.

aggFTW:
Yes, it is considered stable (same way 0.3.0 is). That said, you may or may not bump into API changes from Livy 😄. I'm not sure if the protocol changed or not. Feel free to give it a try then. I won't try the docker image until you tell me to.

apetresc:
Hm, indeed, it appears to not be a drop-in replacement - I'm getting incompatibilities at the Livy <-> Sparkmagic layer after solving the Spark <-> Livy layer. In that case, I will file a separate pull request for that change. I think this one is ready from my end.

autovizwidget/autovizwidget/__init__.py
aggFTW: Actually, sorry to have missed this, but we should ideally bump the 12 to 13 altogether. So new version would be 0.13.0, according to rules of semantic versioning. Can you please bump to 0.13.0 instead?

autovizwidget/requirements.txt
```diff
@@ -1,4 +1,4 @@
-plotly>=1.10.0,<1.11.0
-ipywidgets>5.0.0,<7.0
+plotly>=1.10.0,<3.0
+ipywidgets>5.0.0,<8.0
 hdijupyterutils>=0.6
-notebook>=4.2,<5.0
+notebook>=4.2,<6.0
```
aggFTW: We are missing the pandas requirement here now.

hdijupyterutils/requirements.txt
```diff
@@ -1,8 +1,8 @@
 nose
 mock
 ipywidgets>5.0.0
-ipython==4.0.2
-ipykernel==4.2.2
+ipython>=4.0.2,<7
+ipykernel>=4.2.2,<5
 pandas==0.17.1
```
aggFTW: With pandas being tied to this specific version, requirements files are in conflict. Can you update here? I don't remember why we pinned it like that in the past though...

aggFTW: If you update here, you also need to udpate the setup.py file

aggFTW:
Hey, sorry that it's taken me a while to review. I'm currently super busy at work and my docker installation seems to be messed up. I'm getting:

```
Hardware assisted virtualization and data execution protection must be enabled in the BIOS. See https://docs.docker.com/docker-for-windows/troubleshoot/#virtualization-must-be-enabled
```

when I try to run docker, even though virtualization is indeed enabled. I'll try to look at this ASAP. Are you blocked on it?

aggFTW:
Just ran some hyper-v commands, rebooted... it worked. Merging. Thanks so much!

Issue #410, opened 2017-10-06
build error


harschware:
I ran into an error building docker containers, via `docker-compose build` on master.  I then thought to try the latest release instead so I did `git checkout 0.12.4` and tried again.  This is what I get:

```[INFO] --- exec-maven-plugin:1.2.1:exec (python-api install) @ livy-python-api ---
Downloading: https://repo.maven.apache.org/maven2/org/apache/maven/maven-toolchain/1.0/maven-toolchain-1.0.pom
Downloaded: https://repo.maven.apache.org/maven2/org/apache/maven/maven-toolchain/1.0/maven-toolchain-1.0.pom (4 KB at 115.0 KB/sec)
Downloading: https://repo.maven.apache.org/maven2/org/codehaus/plexus/plexus-utils/2.0.5/plexus-utils-2.0.5.jar
Downloaded: https://repo.maven.apache.org/maven2/org/codehaus/plexus/plexus-utils/2.0.5/plexus-utils-2.0.5.jar (218 KB at 3823.5 KB/sec)
/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
your setuptools is too old (<12)
setuptools_scm functionality is degraded
zip_safe flag not set; analyzing archive contents...

Installed /tmp/easy_install-3ezcy40a/flake8-3.4.1/pytest_runner-2.12.1-py3.4.egg
warning: no previously-included files matching '*.pyc' found anywhere in distribution
no previously-included directories found matching 'docs/build/'
zip_safe flag not set; analyzing archive contents...

Installed /apps/build/livy/python-api/flake8-3.4.1-py3.4.egg
Searching for pytest-runner
Reading https://pypi.python.org/simple/pytest-runner/
Best match: pytest-runner 2.12.1
Downloading https://pypi.python.org/packages/43/bb/def5d83ce6abe507300e013562d3e1229633e86296d3f7f76bd39bf3d2b9/pytest-runner-2.12.1.tar.gz#md5=a8dd57c80cde37a587dd8c37740c1ab1
Processing pytest-runner-2.12.1.tar.gz
Writing /tmp/easy_install-6iyad24k/pytest-runner-2.12.1/setup.cfg
Running pytest-runner-2.12.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-6iyad24k/pytest-runner-2.12.1/egg-dist-tmp-l8p13c5h
[...]

PR #411, opened 2017-10-07
fixes issue #410

harschware:
the build problem described in issue #410 was due to a very out of date setup tools.  At one point in testing I got a warning about pip out of date so I upgrade that too...

### DIFF
--- a/Dockerfile.jupyter
+++ b/Dockerfile.jupyter
@@ -9,6 +9,9 @@ RUN conda install requests-kerberos -y
 
 USER $NB_USER
 
+RUN pip install --upgrade pip
+RUN pip install --upgrade --ignore-installed setuptools
+
 COPY examples /home/jovyan/work
 
 # Install sparkmagic - if DEV_MODE is set, use the one in the host directory.
--- a/Dockerfile.spark
+++ b/Dockerfile.spark
@@ -10,6 +10,8 @@ RUN apt-get update && apt-get install -yq --no-install-recommends --force-yes \
     r-base-core && \
     rm -rf /var/lib/apt/lists/*
 
+RUN pip install --upgrade setuptools
+
 ENV LIVY_BUILD_VERSION livy-server-0.3.0
 ENV LIVY_APP_PATH /apps/$LIVY_BUILD_VERSION
 ENV LIVY_BUILD_PATH /apps/build/livy


Dockerfile.jupyter
```diff
@@ -9,6 +9,9 @@ RUN conda install requests-kerberos -y
 
 USER $NB_USER
 
+RUN pip install --upgrade pip
+RUN pip install --upgrade --ignore-installed setuptools
```
aggFTW: Is the `--ignore-installed` needed?

harschware: yes. it was required.

aggFTW: Why?

Dockerfile.spark
```diff
@@ -10,6 +10,8 @@ RUN apt-get update && apt-get install -yq --no-install-recommends --force-yes \
     r-base-core && \
     rm -rf /var/lib/apt/lists/*
 
+RUN pip install --upgrade setuptools
```
aggFTW: Why not use ignore installed here too then? Should we update pip here as well?

harschware: --ignore-installed was not needed here.  I don't know why.  My environment is OSX Sierra with Docker Community Edition Version 17.09.0-ce-mac35 (19611)

As for upgrading pip here..  this image is built via `FROM gettyimages/spark:2.1.0-hadoop-2.7` which had pip up to date.  Adding the upgrade will yield:
Step 3/12 : RUN pip install --upgrade pip
 ---> Running in 58351e04718b
Requirement already up-to-date: pip in /usr/local/lib/python3.4/dist-packages/pip-9.0.1-py3.4.egg
 whereas the jupyter notebook is built via `FROM jupyter/base-notebook:d0b2d159cc6c` and that pinned version is not up to date.

With this image built from an unpinned version of gettyimages, we may be able to defer the pip version to the base docker image.

Dockerfile.jupyter
harschware: without it I got this error:  

```Step 7/25 : RUN pip install --upgrade setuptools
 ---> Running in a103e27aaeca
Collecting setuptools
  Downloading setuptools-36.6.0-py2.py3-none-any.whl (481kB)
Installing collected packages: setuptools
  Found existing installation: setuptools 23.0.0
Cannot remove entries from nonexistent file /opt/conda/lib/python3.5/site-packages/easy-install.pth
The command '/bin/sh -c pip install --upgrade setuptools' returned a non-zero code: 1
```

I googled until I found the --ignore-installed could get around it.  See:   https://github.com/ContinuumIO/anaconda-issues/issues/542

Dockerfile.spark
harschware: Hi @aggFTW, can you review again?

aggFTW:
Done and thanks so much for the explanations!

PR #413, opened 2017-10-15
WIP: pass %local data to %spark

spaszek:


sparkmagic/sparkmagic/kernels/wrapperkernel/usercodeparser.py
```diff
@@ -18,7 +18,7 @@ def get_code_to_run(self, code):
         except IndexError:
             all_but_first_line = ""
 
-        if code.startswith("%%local") or code.startswith("%local"):
+        if (code.startswith("%%local") or code.startswith("%local")) and not "-o" in code:
```
aggFTW: This seems risky...

sparkmagic/sparkmagic/livyclientlib/sendtosparkcommand.py
```diff
@@ -0,0 +1,50 @@
...
+        self.input_variable_name = input_variable_name
+        self.input_variable_value = input_variable_value
+        self.output_variable_name = output_variable_name
+
+    def execute(self, session):
+        try:
+            command = self.to_command(session.kind, self.input_variable_name, self.input_variable_value, self.output_variable_name)
+            (success, result) = command.execute(session)
+            result = self.input_variable_value #todo ISSUE#412
+            if not success:
```
aggFTW: I wonder what kind of issues would result in not success and whether it's up to each implementation of this abstract class to handle the failures (e.g. JSON serialization failures could only be detected by commands that JSON serialize) and etcetera. What do you think?

sparkmagic/sparkmagic/livyclientlib/sendtosparkcommand.py
```diff
@@ -0,0 +1,50 @@
...
+            return self._scala_command(input_variable_name, input_variable_value, output_variable_name)
+        elif kind == constants.SESSION_KIND_SPARKR:
+            return self._r_command(input_variable_name, input_variable_value, output_variable_name)
+        else:
+            raise BadUserDataException(u"Kind '{}' is not supported.".format(kind))
+
+    @abstractmethod
+    def _scala_command(self, input_variable_name, input_variable_value, output_variable_name):
+        raise NotImplementedError #override and provide proper implementation in supertype!
+
```
aggFTW: Single space between methods inside of classes 😃

aggFTW: I just realized we did not follow this for other classes... we need to fix it. I can do that in another review, or feel free to do it yourself.

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
```diff
@@ -37,6 +39,29 @@ def __init__(self, shell, data=None, spark_events=None):
...
+    def execute_local(self, cell, input_variable_name, type, output_variable_name, session_name):
+        input_variable_value = self.shell.user_ns[input_variable_name]
+
+        if not input_variable_value:
+            raise BadUserDataException("Value of specified -o variable is None!")
+
+        if not output_variable_name:
+            output_variable_name = input_variable_name
+
+        if type == 'str':
```
aggFTW: `type` is a reserved word; how about `var_type`?
Also, let's make the types constants and compare after lowercasing the user input?

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
```diff
@@ -37,6 +39,29 @@ def __init__(self, shell, data=None, spark_events=None):
...
+            command = SendStringToSparkCommand(input_variable_name, input_variable_value, output_variable_name)
+        elif type == 'df':
+            command = SendPandasDfToSparkCommand(input_variable_name, input_variable_value, output_variable_name)
+        else:
+            raise BadUserDataException("Invalid -t type. Available are: `str` or `df`.")
```
aggFTW: Let's construct the available types by joining an array from constants

sparkmagic/sparkmagic/livyclientlib/sendtosparkcommand.py
aggFTW: Why check for success here instead of simply passing it up?

aggFTW:
Solid, solid WIP! 😃 I'm so glad to see this! I haven't been able to pull and try it myself, but it looks good from an overall approach. I see you are already handling multiple languages, so I'm very happy about that!

I'd like to see tests for:
* escaping JSON right in all languages, and failure handling so user gets an actionable message
* handle multiple encodings correctly (in python 2 vs python3, for example)

We should also add:
* a way to limit the amount of data you can send
* support for different versions of spark (2.X onwards at least) -- sparkmagic is currently not versioned, as Livy is not 1.0 yet, but we need to probably start adding a spark to livy to sparkmagic version table to the README...
* support for magic-based ipython kernel scenarios
* unit tests
* documentation in README on how to use this new feature
* documentation in sample notebooks on how to use this new feature to:
    1. generate a pandas df locally, serialize it to the cluster and then pass it to the remote spark cluster
    2. read a local file containing a serialized trained model into a string, pass it to the remote cluster, and deserialize there

Please let me know what you need from me to get this to the finish line and I'll try to make it happen as best as I can.

sparkmagic/sparkmagic/livyclientlib/sendtosparkcommand.py
spaszek: This section is heavily work-in-progressish and I literally have no idea why I did that, lol. Will surely refactor this part later. :)

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
spaszek: sure, good idea

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
spaszek: will do

sparkmagic/sparkmagic/kernels/wrapperkernel/usercodeparser.py
spaszek: Yes, I know this code smells, but so does the entire method(imo). I just followed the convention. I would have parsed users' first input line more sophisticatedly overall. Not sure if this PR is a good place to change it tho.

What would you suggest instead of this line?

spaszek:
@aggFTW 
Thanks for the small review and I'm glad you said it looks good from overall approach. There is just one thing - I would like to allow only one kind of user input, looking like:
```python
%%local -o var
```
That is - user mustn't put anything into the cell. Right now the code above does not work - IPython forces me to put something in the cell, otherwise throwing error along the lines `cellmagics cell cannot be empty`.

I have to use it like this:
```python
%%local -o var
# empty comment
```
Putting any code in parameterized cell `%%local` is now very misleading too - because it **DOES NOT** execute locally(even though you would exepect it does).

How do I change it, to work like:
- if the cell is just `%%local` without any params, require it to be nonempty and execute the code locally (current behavior)
- if the cell is `%%local` with params like `-o`, require it to be empty

sparkmagic/sparkmagic/livyclientlib/sendtosparkcommand.py
spaszek: uh, should we use single or double newline then? :)

sparkmagic/sparkmagic/kernels/wrapperkernel/usercodeparser.py
aggFTW: Yes... I've never been to happy with this class overall. How would you make it more sophisticated? I don't have suggestions as of now.

sparkmagic/sparkmagic/livyclientlib/sendtosparkcommand.py
aggFTW: Single for methods in classes. Double for methods outside of classes.

aggFTW:
>How do I change it, to work like:
>- if the cell is just `%%local` without any params, require it to be nonempty and execute the code locally (current behavior)
>- if the cell is `%%local` with params like `-o`, require it to be empty

I would leave `local` as the magic it was, and add a new magic altogether for this. What about calling it something like `sparksend`... feel free to suggest a better name.

Then you could just add this new magic to the list of empty body cells in UserCodeParser called `_magics_with_no_cell_body `, and it would work. `-o` would be a required param for this new magic.

This way you avoid the misleading behavior.

P.S. I may be a little bit more unresponsive this week.

spaszek:
@aggFTW I managed to implemented working(wohoo!) solution for both Python and Scala. Now I'm pretty much stuck on R - I have never seen that language in my life before. Could you help me a bit?

deserialization in both Scala:
```scala
val seralizedPandasStr = """[{"int":1,"string":"a"},{"int":2,"string":"b"},{"int":3,"string":"c"},{"int":4,"string":"d"},{"int":5,"string":"e"}]"""

val rdd_json_array = spark.sparkContext.makeRDD(serializedPandasStr :: Nil)
return spark.read.json(rdd_json_array) /// returns a DataFrame object
```
and Python:
```python
serialized_pandas_str = '''[{"int":1,"string":"a"},{"int":2,"string":"b"},{"int":3,"string":"c"},{"int":4,"string":"d"},{"int":5,"string":"e"}]'''

json_array = json.loads(serialized_pandas_str)
rdd_json_array = spark.sparkContext.parallelize(json_array)
return spark.read.json(rdd_json_array) # returns a DataFrame object
```
is pretty easy and works well. What would be the equivalent of that in R/SparkR?

I have access to the variable in R:
```r
dataframe_json <- '"[{"int":1,"string":"a"},{"int":2,"string":"b"},{"int":3,"string":"c"},{"int":4,"string":"d"},{"int":5,"string":"e"}]"'
```

I tried googling but the solutions either require external library(`require(RJSONIO)` which the kernel does not have) or load the json from file(`read.json` method from `SparkR` does not allow me to use String - it wants a path to file). 

Perhaps I should transform my json string into RDD and then load it by `data.frame`? That was my first thought, because that is what I have done with Scala/Python, but the API reference https://spark.apache.org/docs/latest/api/R/index.html has literally no occurence of `rdd` word, lol.

aggFTW:
Sorry I took a long time to reply. I was out and just got back. First thought is to save the string into a file and then read via read.json: https://spark.apache.org/docs/latest/api/R/read.json.html

I cannot seem to find anything else 😢 (I'm no SparkR or R expert myself)...

I also see this: https://spark.apache.org/docs/latest/api/R/from_json.html but don't know which one is better. Maybe post to the spark R user/dev distribution lists?

spaszek:
https://spark.apache.org/docs/latest/api/R/from_json.html looks exactly like it, although it is not avaialble in Spark 2.1 :cry: 

I have implemented your idea of saving the json to file and it worked correctly. Unfortunately, I also had to `.persist()` the loaded dataframe. If I had not done that, the temporary file would have sticked in the system(because I could not have deleted it before user actually made Spark run the code - and I have no control over it/no indication when it happens :P).

Also, because R has no triple-quotes mechanism like Scala/Python, the string escaping is pretty clunky as you can see in `sparkmagic/livyclientlib/sendstringtosparkcommand.py:19`. Feel free to guide me to do it better :). I tried using
```python
import re
re.escape(input)
```
but then R kernel did not parse it correctly(and I have no idea where the stacktraces are, lol)

So, summing up, right now:
- user can send any String to PySpark/Scala kernel (as long as it has no triple-quotes?)
- ...and most of the Strings to R kernel :P
- they also are able to transfer Pandas DF - (in PySpark/Scala without `.persist()` invocation, in R with it)

What needs to be done:

- allow pre-2.0 Spark <- how would you do that? Should I first check the existence of `sqlContext` or `spark` object in kernel and based on that decide my next steps? or is there a cleaner way? btw imho for now this PR could just allow Spark>2.0 and I would add Spark<2.0 support in another one
- `allow user to limit the data` <- what did you mean?
- add unit tests :P
- fill missing parts in example notebooks

right?

aggFTW:
Hey @spaszek, nice to see all the progress! Thank you so much for working on this!

It made sense that you had to persist the dataframe after saving to disk. My worries with this approach are that problems arise due to data volume or that we run into file permission issues. One alternative to all of this is to say that this feature is only supported from Spark 2.2 onwards for R, and let user fail otherwise. This may be a better tradeoff than a longer term "not-so-clean" approach?

I think it's OK to not implement this feature altogether for pre 2.X versions, so long as user gets a clear error message saying that it's not supported. You can see our (very hacky) way of checking whether it's Spark 2.X or not here (spark variable presence means it's 2.X). Feel free to refactor the check so that you can use it if you need to: https://github.com/jupyter-incubator/sparkmagic/blob/f65571a3fb9092afbc03212587ce0710ed79cadf/sparkmagic/sparkmagic/livyclientlib/livysession.py#L138

As for escaping things properly in all languages, I would just create a JSON payload and let JSON libraries do the serialization/deserialization in all languages, similar to how we do auto visualization today. I wouldn't try to escape things manually. That way, you ensure all strings are supported, regardless of quote types or other factors. Note that we use jsonlite on the R kernel: https://github.com/jupyter-incubator/sparkmagic/blob/f65571a3fb9092afbc03212587ce0710ed79cadf/sparkmagic/sparkmagic/livyclientlib/sqlquery.py#L112

As for what needs to be done, we had this list. Feel free to disagree with me on which items should be on the list, and we can discuss. Once we get to the final list, you can check box what you've done so far to keep a tally of it, and once you've completed it, I will take a final review:

- [ ] escaping JSON right in all languages
- [ ] JSON escaping failure handling so user gets an actionable message
- [ ] handle multiple encodings correctly (in python 2 vs python3, for example)
- [ ] a way to limit the amount of data you can send. This means that a user can choose to send X rows from the pandas DF or all rows, or a sample, or etcetera. This is similar to the way users can currently choose how much data to visualize automatically from a dataframe.
- [ ] support for different versions of spark (2.X onwards at least) -- maybe 2.2 onwards for R...
- [ ] support for magic-based ipython kernel scenarios
- [ ] unit tests
- [ ] a spark to livy to sparkmagic version table to the README
- [ ] documentation in README on how to use this new feature
- [ ] documentation in sample notebooks on how to use this new feature to:
- [ ] generate a pandas df locally, serialize it to the cluster and then pass it to the remote spark cluster
- [ ] read a local file containing a serialized trained model into a string, pass it to the remote cluster, and deserialize there

What do you think? Sounds good? Please do reach out if you need to speak to me or if you need anything else.

Thanks!!!

spaszek:
-  _escaping JSON right in all languages_
sure
- _JSON escaping failure handling so user gets an actionable message_
will do
- _handle multiple encodings correctly (in python 2 vs python3, for example)_
could you provide an example in code? i already have different methods for python 2 and 3 kernels in both SparkCommands
- _a way to limit the amount of data you can send. This means that a user can choose to send X rows from the pandas DF or all rows, or a sample, or etcetera. This is similar to the way users can currently choose how much data to visualize automatically from a dataframe._
user can just `.head(10)` his dataframe to limit it, don't he? :) or is there another idea behind that?
- _support for different versions of spark (2.X onwards at least) -- maybe 2.2 onwards for R..._
its done already I think
- _support for magic-based ipython kernel scenarios_
not sure what's a magic-based kernel, but will check
- _unit tests_
😒 i usually love tests but somehow hate them(especially mocking) in python. will do my best though
- _a spark to livy to sparkmagic version table to the README_
could you do it instead? i'm not really sure about all the versions compatible. i can provide info about my `%send_to_spark`, you do the `%spark -o` and stuff
- _documentation in README on how to use this new feature_
done
- _documentation in sample notebooks on how to use this new feature to:
 generate a pandas df locally, serialize it to the cluster and then pass it to the remote spark cluster
 read a local file containing a serialized trained model into a string, pass it to the remote cluster, and deserialize there_
sure, but it would require another library(e.g. `scikit-learn` with `pickle`) installed on kernel. or should I install it just one time, to provide an example-working-solution, save the notebook and then `pip remove`?

No need to thanks me, I thank you for creating it in the first place. I use `sparkmagics` a lot right now. 

I have little free time this week, but I think the PR will be finished in November. :)

spaszek:
@aggFTW ping :)

aggFTW:
>- _handle multiple encodings correctly (in python 2 vs python3, for example)_
could you provide an example in code? i already have different methods for python 2 and 3 kernels in both SparkCommands

Look at some tests for what we had to do here: https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/sqlquery.py#L82

>- _a way to limit the amount of data you can send. This means that a user can choose to send X rows from the pandas DF or all rows, or a sample, or etcetera. This is similar to the way users can currently choose how much data to visualize automatically from a dataframe._
user can just `.head(10)` his dataframe to limit it, don't he? :) or is there another idea behind that?

This is to be consistent with the way we allow users to limit data with other similar commands, like sql or spark.

>- _a spark to livy to sparkmagic version table to the README_
could you do it instead? i'm not really sure about all the versions compatible. i can provide info about my `%send_to_spark`, you do the `%spark -o` and stuff

Yes, I can add that.

>- _documentation in sample notebooks on how to use this new feature to:
 generate a pandas df locally, serialize it to the cluster and then pass it to the remote spark cluster
 read a local file containing a serialized trained model into a string, pass it to the remote cluster, and deserialize there_
sure, but it would require another library(e.g. `scikit-learn` with `pickle`) installed on kernel. or should I install it just one time, to provide an example-working-solution, save the notebook and then `pip remove`?

You just need to write the notebook and add it to the examples folder (you can specify which libraries to have installed for it to work in the notebook itself in the first paragraph). Users would then install in their own computers.

spaszek:
allright, (hopefully) last question before the final review - how do I instal a python package inside PySpark kernel? I tried slipping in few lines:
```
RUN pip install -U numpy==1.13.3
RUN pip install -U scipy==1.0.0
RUN pip install -U scikit-learn==0.19.1
```
to Dockerfile.spark but it did not work:
```python
No module named numpy
ImportError: No module named numpy
```

aggFTW:
Try removing the -U maybe?

spaszek:
Allright, one more set of unit tests, encoding fix and I'm ready to go! Should be over next week.

BTW. I have added an example file how one should use the new feature - https://github.com/spaszek/sparkmagic/blob/35619b3e2686d1f007cff0f2bcbc685b532e23d9/examples/Send%2Blocal%2Bdata%2Bto%2BSpark%2Bexample.ipynb 

is it fine?

aggFTW:
Woohoo! It looks really neat! I have a couple of very minor comments, but I'll just add them to the final review once you tell me to review.

Do note that I'll be going on vacations soon, so please be patient with me if I disappear for a little while 😄

spaszek:
Sorry - I had a lot of things going on with little free time to spare.

Gonna finish this PR ASAP, I promise :)

spaszek:
@aggFTW - its (almost) ready to merge - please do your final code review. :)

I did all the checklist tasks except for:

-  "encode python2 vs python3" - because I do not print the executed commands anywhere(and I guess that encoding-to-utf-8-trick was only for that).
- actionable message for json encode failures - if I guessed correctly, I would have to surround following block with try-catch:
```python
try:            
    return pandas_df.head(self.max_rows).to_json(orient=u'records')
except ???:
    raise BadUserDataException("Failed to convert df to correct json.")
```
what exception should I catch in  `except` though?

README.md
```diff
@@ -19,6 +19,7 @@ The Sparkmagic project includes a set of magics for interactively running Spark
 * Automatic visualization of SQL queries in the PySpark, PySpark3, Spark and SparkR kernels; use an easy visual interface to interactively construct visualizations, no code required
 * Easy access to Spark application information and logs (`%%info` magic)
 * Ability to capture the output of SQL queries as Pandas dataframes to interact with other Python libraries (e.g. matplotlib)
+* Send local files or dataframes to a remote cluster (e.g. sending pretrained local ML model straight to the Spark cluster)
```
harschware: nice work!! don't forget to update the output of %%help to show the new -o param to %%local and the subsequent change to the README.

spaszek: Actually, there is a new cell magic dedicated to that - `%send_to_spark`

It has been already added to `%help` and there is also an example `.ipynb` usage attached: https://github.com/spaszek/sparkmagic/blob/35619b3e2686d1f007cff0f2bcbc685b532e23d9/examples/Send%2Blocal%2Bdata%2Bto%2BSpark%2Bexample.ipynb

spaszek:
@aggFTW bump for the CR

spaszek:
@aggFTW :P

spaszek:
@aggFTW @harschware  hello?

roberto-fierimonte:
Hey - any news here? This is one of the best feature we can get and it seems just a step away from being merged!

harschware:
@spaszek I can't approve merges on this project and I only took a cursory look at the PR.  Maybe @aggFTW will be able to help.

alfozan:
@aggFTW

phil-trinh:
bumping this up since we're so close!
@aggFTW
@Carreau
@jdfreder

sangramga:
@aggFTW @spaszek @Carreau @jdfreder Could you please take a look at this PR. It's a really needed feature!!

Tagar:
@spaszek would it be possible to resolve the conflicts? 

Thanks

djememahamat:
@spaszek Could we except having this merge this week? Pleaaaaaase

itamarst:
I will also see if I can get this in when I next do maintenance work on Sparkmagic.

itamarst:
This has been superseded by #574, so I can edit the code myself.

itamarst:
I would appreciate testing of the version in #574, especially with Scala and R.

itamarst:
Reminder: would appreciate user testing of the version of this PR in #574, so I can get it merged.

spaszek:
@roberto-fierimonte @harschware  @djememahamat @Tagar @sangramga @phil-trinh @alfozan @harschware I am happy to announce that https://github.com/jupyter-incubator/sparkmagic/pull/574 ( a continouation of this task) was just merged to master branch.

PR #421, opened 2017-11-21
Fix the issue that pyspark3 not supported by Livy 0.4

taoliseki:
Livy 0.4 onwards removed the "pyspark3" as the session kind, so sparkmagic receives a 400 response when we create a pyspark3 notebook on Jupyter. So pyspark3 kernel is not working due to this Livy breaking change.

In this PR, we are specifying "spark.yarn.appMasterEnv.PYSPARK_PYTHON" to "python" or "python3" to kick off different python versions, which is the new approach to differentiate pyspark/pyspark3 sessions. This approach works with the old livy version (e.g. 0.3) as well as 0.4 onwards. Also "spark.yarn.appMasterEnv.PYSPARK_PYTHON"  is supported both spark 1.6 and spark 2.

### DIFF
--- a/sparkmagic/sparkmagic/livyclientlib/livysession.py
+++ b/sparkmagic/sparkmagic/livyclientlib/livysession.py
@@ -113,6 +113,10 @@ class LivySession(ObjectWithGuid):
     def _translate_to_livy_kind(self, properties, kind):
         # Livy does not support "pyspark3" as the kind from 0.4 onwards and only "pyspark" is valid.
         # That's why we are leveraging PYSPARK_PYTHON_PARAM here to specifyf different python versions.
+
+        if kind != constants.SESSION_KIND_PYSPARK and kind != constants.SESSION_KIND_PYSPARK3:
+            return
+
         sparkConfigs = {}
         if constants.LIVY_CONF_PARAM in properties:
             sparkConfigs = properties[constants.LIVY_CONF_PARAM]
--- a/sparkmagic/sparkmagic/livyclientlib/livysession.py
+++ b/sparkmagic/sparkmagic/livyclientlib/livysession.py
@@ -110,11 +110,22 @@ class LivySession(ObjectWithGuid):
         return u"Session id: {}\tYARN id: {}\tKind: {}\tState: {}\n\tSpark UI: {}\n\tDriver Log: {}"\
             .format(self.id, self.get_app_id(), self.kind, self.status, self.get_spark_ui_url(), self.get_driver_log_url())
 
+    def _translate_to_livy_kind(self, properties, kind):
+        # Livy does not support "pyspark3" as the kind from 0.4 onwards and only "pyspark" is valid.
+        # That's why we are leveraging PYSPARK_PYTHON_PARAM here to specifyf different python versions.
+        if kind == constants.SESSION_KIND_PYSPARK:
+            python_properties = {constants.PYSPARK_PYTHON_PARAM: constants.LANG_PYTHON}
+            properties[constants.LIVY_CONF_PARAM] = python_properties
+        elif kind == constants.SESSION_KIND_PYSPARK3:
+            properties[constants.LIVY_KIND_PARAM] = constants.SESSION_KIND_PYSPARK
+            python_properties = {constants.PYSPARK_PYTHON_PARAM: constants.LANG_PYTHON3}
+            properties[constants.LIVY_CONF_PARAM] = python_properties
+
     def start(self):
         """Start the session against actual livy server."""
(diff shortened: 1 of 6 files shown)


sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -32,7 +32,7 @@ def override_all(obj):
 
 # Helpers
 
-def get_livy_kind(language):
+def get_session_kind(language):
```
aggFTW: Can we please revert the change in name? Was it done for any particular reason?

taoliseki: Yes there is a reason for that change. 

Basically Livy 0.4 onwards has removed "pyspark3" as session kind, which is a breaking change. For sparkmagic, we switch to use the pyspark setting "spark.pyspark.python" to specify python2/python3 instead of the old approach by using the Livy session kind.

As you can see from my change, we are sticking with pyspark/pyspark3 as the session kind definition for sparkmagic, and then translate that to the correct config settings which are sent to Livy when creating a session. I am renaming the method name because this method is now using the new session definition (for sparkmagic), as opposed to livy session definition (which does not have pyspark3 now).

Does it make sense?

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -110,11 +110,22 @@ def __str__(self):
...
 
+    def _translate_to_livy_kind(self, properties, kind):
+        # Livy does not support "pyspark3" as the kind from 0.4 onwards and only "pyspark" is valid.
+        # That's why we are leveraging PYSPARK_PYTHON_PARAM here to specifyf different python versions.
+        if kind == constants.SESSION_KIND_PYSPARK:
+            python_properties = {constants.PYSPARK_PYTHON_PARAM: constants.LANG_PYTHON}
+            properties[constants.LIVY_CONF_PARAM] = python_properties
+        elif kind == constants.SESSION_KIND_PYSPARK3:
+            properties[constants.LIVY_KIND_PARAM] = constants.SESSION_KIND_PYSPARK
+            python_properties = {constants.PYSPARK_PYTHON_PARAM: constants.LANG_PYTHON3}
```
aggFTW: The python3 path should be configurable, and not a constant.

mpekalski: Also this kind of replacement of properties breaks configuration JSON, e.g. below is before and after. Basically it lost 'spark.jars.packages' I had passed over in the conf. 


before 
```
 properties: {'driverMemory': '1G', 'executorCores': 4, 'kind': 'pyspark3', 'conf': {'spark.pyspark.python': 'python3', 'spark.jars.packages': 'com.databricks:spark-avro_2.11:4.0.0'}, 'heartbeatTimeoutInSecond': 120}
```
after
```
 properties: {'driverMemory': '1G', 'executorCores': 4, 'kind': 'pyspark', 'conf': {'spark.pyspark.python': 'python3'}, 'heartbeatTimeoutInSecond': 120}
```

sparkmagic/sparkmagic/utils/configuration.py
aggFTW: This change is not merged.

chuyqa: 0.4 Release Still has `pyspark3` https://github.com/apache/incubator-livy/blob/branch-0.4/core/src/main/scala/org/apache/livy/sessions/Kind.scala

0.5 Removes it https://github.com/apache/incubator-livy/blob/branch-0.5/core/src/main/scala/org/apache/livy/sessions/Kind.scala#L40

gm-spacagna:
Hi I have tried to to fix this compatibility issue in a different way in this PR: https://github.com/jupyter-incubator/sparkmagic/pull/507
Would you mind having a look?

ranjitiyer:
@gm-spacagna I have a PR that's failing build https://github.com/jupyter-incubator/sparkmagic/pull/462. Any tips on how to resolve it?

sparkmagic/sparkmagic/utils/configuration.py
kanastasiou4signifyd: Has anyone tested this on livy 0.5?

foscraig:
@gm-spacagna @aggFTW Can you please answer the question? Would like someone to look at that PRs and it's been almost a year, so want to get Travis CI working.

foscraig:
Also tagging @apetresc in case he knows the answer to @ranjitiyer's question. Thanks!

itamarst:
This is superseded by #540, which I hope to get merged soon. Please try it!

PR #427, opened 2017-12-19
Add python3 option in %manage_spark magic

### LINKED ISSUE #420: %manage_spark only provides 'scala' and 'python' language options
As noted in #357 , the %manage_spark user interface only provides 'scala' and 'python' language options. In order for Python 3 to be used and unencoded JSON dataframes to be returned to the local environment (via %spark -o), 'python3' should also be included.

b11z:
Fixes #420
screen shot 2017-12-18 at 5 00 25 pm


aggFTW:
Thanks!

Issue #404, opened 2017-08-29
SSL certificate verify loading server certificate


JeffRodriguez:
SparkMagic could load server certificate, ca_certs, and use it to verify SSL Server certificate. Instead of relying on a DEFAULT_CA_BUNDLE_PATH. This will require to load the server certificate and validate with requests.
The user case would be in the cases where we can to connect to an SSL intermediary to access liv

aggFTW:
I think that we could expose a `custom_certfiles_path` as a configuration and use it with the `verify` parameter for requests. The default value in python if no path is specified in the config file should be `None`.

If you are implementing this, please send a PR with unit tests.

JeffRodriguez:
Thanks Alejandro. That make a lot of sense, I like the solution using the conf.json file

polomarcus:
Hi @JeffRodriguez :smile: 

I would be interested by the same feature, I can propose a PR if you haven't had the time to work on it

bhadrip:
Is there any work needed to get this in ? I can work on it, we also need this fix.

PR #435, opened 2018-01-31
SSL with custom certfiles path

polomarcus:
Regarding #404 

Using verify to pass the certificate path as shown here 
* http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

Configure it using ` custom_certfiles_path`

### DIFF
--- a/sparkmagic/example_config.json
+++ b/sparkmagic/example_config.json
@@ -48,7 +48,7 @@
   "fatal_error_suggestion": "The code failed because of a fatal error:\n\t{}.\n\nSome things to try:\na) Make sure Spark has enough available resources for Jupyter to create a Spark context.\nb) Contact your Jupyter administrator to make sure the Spark magics library is configured correctly.\nc) Restart the kernel.",
 
   "ignore_ssl_errors": false,
-  "custom_certfiles_path": "ca-bundle.crt",
+  "custom_certfiles_path": null,
 
   "session_configs": {
     "driverMemory": "1000M",
--- a/sparkmagic/example_config.json
+++ b/sparkmagic/example_config.json
@@ -48,6 +48,7 @@
   "fatal_error_suggestion": "The code failed because of a fatal error:\n\t{}.\n\nSome things to try:\na) Make sure Spark has enough available resources for Jupyter to create a Spark context.\nb) Contact your Jupyter administrator to make sure the Spark magics library is configured correctly.\nc) Restart the kernel.",
 
   "ignore_ssl_errors": false,
+  "custom_certfiles_path": "cacerts.pem",
 
   "session_configs": {
     "driverMemory": "1000M",
--- a/sparkmagic/example_config.json
+++ b/sparkmagic/example_config.json
@@ -48,7 +48,7 @@
   "fatal_error_suggestion": "The code failed because of a fatal error:\n\t{}.\n\nSome things to try:\na) Make sure Spark has enough available resources for Jupyter to create a Spark context.\nb) Contact your Jupyter administrator to make sure the Spark magics library is configured correctly.\nc) Restart the kernel.",
 
   "ignore_ssl_errors": false,
-  "custom_certfiles_path": "cacerts.pem",
+  "custom_certfiles_path": "ca-bundle.crt",
 
   "session_configs": {
     "driverMemory": "1000M",
--- a/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
+++ b/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
@@ -33,6 +33,10 @@ class ReliableHttpClient(object):
         if not self.verify_ssl:
             self.logger.debug(u"ATTENTION: Will ignore SSL errors. This might render you vulnerable to attacks.")
             requests.packages.urllib3.disable_warnings()
+        else:
+            if conf.custom_certfiles_path is not None:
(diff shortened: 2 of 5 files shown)


sparkmagic/example_config.json
```diff
@@ -48,6 +48,7 @@
   "fatal_error_suggestion": "The code failed because of a fatal error:\n\t{}.\n\nSome things to try:\na) Make sure Spark has enough available resources for Jupyter to create a Spark context.\nb) Contact your Jupyter administrator to make sure the Spark magics library is configured correctly.\nc) Restart the kernel.",
 
   "ignore_ssl_errors": false,
+  "custom_certfiles_path": "cacerts.pem",
```
aggFTW: Let's change default value so that it ends up with `None` value in python.

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, endpoint, headers, retry_policy):
         if not self.verify_ssl:
             self.logger.debug(u"ATTENTION: Will ignore SSL errors. This might render you vulnerable to attacks.")
             requests.packages.urllib3.disable_warnings()
+        else:
+            if conf.custom_certfiles_path is not None:
```
aggFTW: This should be a method call

aggFTW:
LGTM otherwise!

@apetresc, do you have any comments?

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -164,6 +164,9 @@ def resource_limit_mitigation_suggestion():
 def ignore_ssl_errors():
     return False
 
+@_with_override
+def custom_certfiles_path():
+    return None
```
polomarcus: @aggFTW this is how we handle None value for `custom_certfiles_path` , so I'm not sure what to do for ` example_config.json`

aggFTW: If you did:

`"custom_certfiles_path": null,`

would it end up returning `None`?

I think it would. Maybe try that out.

polomarcus: What do you think about not using ` "custom_certfiles_path": null` in the configuration file but telling in the readme that we can set this value ? 
I, personally,  don't like to have null value set somewhere

aggFTW: The purpose of having an example_config.json in the repo is to have a place that lists all the configurable options for the library and show what the default values are, so this option would need to be there and match the python `None` value. Let's add it so that it translates to `None` if used.

polomarcus: with ` json.loads(path)` null translates to None

https://docs.python.org/3/library/json.html#encoders-and-decoders

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -94,3 +96,8 @@ def _send_request_helper(self, url, accepted_status_codes, function, data, retry
                     raise HttpClientException(u"Invalid status code '{}' from {} with error payload: {}"
                                               .format(status, url, text))
             return r
+
+    def _set_custom_certfiles_path(self):
+        if conf.custom_certfiles_path is not None:
```
aggFTW: `conf.custom_certfiles_path` is not a method call. This needs to invoke the method to retrieve the value of the configuration. So it would change to `conf.custom_certfiles_path()`. That's what I meant with my previous comment 😃

aggFTW: So just put this method in line 37, like you had it before, but make sure to actually retrieve the value of the configuration. Does that make sense?

sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -78,3 +78,13 @@ def test_share_config_between_pyspark_and_pyspark3():
     kpc = { 'username': 'U', 'password': 'P', 'base64_password': 'cGFzc3dvcmQ=', 'url': 'L', 'auth': AUTH_BASIC }
     overrides = { conf.kernel_python_credentials.__name__: kpc }
     assert_equals(conf.base64_kernel_python3_credentials(), conf.base64_kernel_python_credentials())
+
+@with_setup(_setup)
+def test_custom_certfiles_path():
+    overrides = { }
```
aggFTW: There's no need for this test.

sparkmagic/sparkmagic/tests/test_reliablehttpclient.py
```diff
@@ -232,3 +232,4 @@ def test_kerberos_auth_check_auth():
     client = ReliableHttpClient(endpoint, {}, retry_policy)
     assert_is_not_none(client._auth)
     assert isinstance(client._auth, HTTPKerberosAuth)
+
```
aggFTW: Adding a test here to see that the verify parameter in the request is honoring the value of `conf.custom_certfiles_path()` when it's set would be useful and would have caught the lack of method call I mentioned in the other file. Let's add it.

JunLyu:
Hi @polomarcus! I came across this [thread](https://github.com/jupyter-incubator/sparkmagic/issues/404) and thought your PR can be helpful. Do you still plan to complete this PR?

PR #439, opened 2018-02-12
Load plotly.js from CDN rather than by inclusion

### LINKED ISSUE #438: Autovizwidget includes many copies of plotly.js
The autovizwidget includes multiple copies of the plotly.js source code in Jupyter notebook files that are saved. Any cell that displays a dataframe or query output has two copies of plotly attached to it in its JSON representation, one keyed by `text/html` and another by `text/vnd.plotly.v1+html`.

Because each copy of the plotly source comprises 2.3 MB, and notebooks can easily contain dozens of these outputs, this leads to ballooning notebook file sizes (up to 93 MB in my experience). Jupyter's autosave functionality can stop working when notebooks become this large, leading to unexpected data loss.

baldwint:
Greatly reduces notebook file size by setting `connected=True` in the call to `init_notebook_mode`, as suggested [here](https://github.com/plotly/plotly.py/blob/master/plotly/offline/offline.py#L92). Fixes #438.

### DIFF
--- a/autovizwidget/requirements.txt
+++ b/autovizwidget/requirements.txt
@@ -1,4 +1,4 @@
-plotly>=1.10.0,<3.0
+plotly>=1.13.0,<3.0
 ipywidgets>5.0.0,<8.0
 hdijupyterutils>=0.6
 notebook>=4.2,<6.0
--- a/autovizwidget/setup.py
+++ b/autovizwidget/setup.py
@@ -57,7 +57,7 @@ setup(name=NAME,
         'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4'],
       install_requires=[
-          'plotly>=1.10.0,<3.0',
+          'plotly>=1.13.0,<3.0',
           'ipywidgets>5.0.0,<8.0',
           'hdijupyterutils>=0.6',
           'notebook>=4.2,<6.0',


aggFTW:
Hey, thanks so much for sending this 👍 

We have autovizwidget listing plotly version requirements here:
https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/requirements.txt#L1
and here:
https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/setup.py#L60

`1.10.0` doesn't have that parameter available. It was introduced in `1.13.0`. Could you please update the requirements there? I think it's good after that!

baldwint:
Thank you for the review! And good catch. I modified the two files to update the minimum plotly version required.

aggFTW:
LGTM. Waiting for @apetresc's feedback.

apetresc:
Did some manual testing and everything seems to work fine :smile:

That being said, I think it would be optimal to allow this to be configurable in `~/.sparkmagic/config.json`, maybe with a key like `embed_plotly` - and I agree that `connected=True` (i.e, `embed_plotly=False`) should be the default value.

The tradeoff being made is that if we force `connected=True`, exported notebooks won't work at all if opened on a laptop without an internet connection, or on a particularly restrictive firewall, with no way to fix it.

I can make this change, just wanted to get @aggFTW's approval (on both the idea and the name of the key in `config.json`).

aggFTW:
Yes, that sounds good! 👍

thesuperzapper:
@apetresc @aggFTW  is this going to be merged?

thesuperzapper:
@baldwint

PR #456, opened 2018-05-11
Filter out unsupported session types

rytis:
Right now SparkMagic cannot establish a new connection if there's already a connection of an unsupported kind (for example Zeppelin "shared" session).

This patch is to ignore already established sessions that have unsupported kinds.


Tagar:
@rytis what currently happens if there is already an established session of type "shared"?

https://github.com/apache/incubator-livy/blob/branch-0.6/core/src/main/scala/org/apache/livy/sessions/Kind.scala#L34

It's an interesting idea for `sparkmagic` to support shared session across different notebooks. I see so many use cases for this. 

Thank you.

Issue #458, opened 2018-05-16
pandas 0.23.0 causing error


rsrpsinr:
Pandas 0.23.0 (released to pypi on May 16 2018) causing the following error:

```
The code failed because of a fatal error:
	Failed to register auto viz for notebook.
Exception details:
	"cannot import name 'DataError'".
```

Caused by scripts that contains the following import (changed in Pandas 0.23.0):
`from pandas.core.groupby import DataError`

https://github.com/jupyter-incubator/sparkmagic/search?utf8=%E2%9C%93&q=from+pandas.core.groupby+import+DataError&type=

o-alex:
Is there any workarounds for this issue? I am also getting this error now.

robzor92:
In pandas 0.23.0 the import path for DataError should be "**from pandas.core.base import DataError**"

mauza:
I'm going to try and pull down the change from jimdowling and build from source. Is that the way to go if I want to get around this now and not wait for another release?

mauza:
I guess I could just downgrade pandas.

ivangvasquez:
This is indeed a brand new issue, as sparkmagic 0.12.5 working for us (anaconda 5.0.1 and 5.1.0). Attempted to downgrade pandas and causes all sorts of dependency problems. Can you suggest a temporary workaround? Install an earlier version of sparkmagic perhaps?

ivangvasquez:
Easy enough:
pip install pandas==0.22.0

jimdowling:
Another work around is to 
git clone https://github.com/logicalclocks/sparkmagic.git
cd sparkmagic
pip install --no-cache-dir ./hdijupyterutils 
pip install --no-cache-dir --upgrade ./autovizwidget
pip install --no-cache-dir ./sparkmagic

kevcunnane:
@apetresc or @kkonrad I noticed you merged PRs recently. Can you please merge https://github.com/jupyter-incubator/sparkmagic/pull/459 from @jimdowling since it fixes this issue? It seems like pandas 0.22.0 doesn't work with Python 3.7, so we can't work at all with latest Python + SparkMagic.

Without this type of fix teams will have to fork / maintain private versions of sparkmagic to keep up to date on support, so it'd be great to see some of the community PRs accepted here. Thanks.

kkonrad:
[...]

PR #459, opened 2018-05-18
DataError in Pandas moved from core.groupby to core.base

jimdowling:
An update in Pandas to version 0.23 has caused DataError to move from one package to another, breaking two references in SparkMagic.
Issue #458  described this problem which causes SparkMagic to fail.

I am not sure if DataError is a public module that it is ok for SparkMagic to depend on.


apetresc:
Thank you for this fix, @jimdowling. Confirmed this works with all versions of pandas - since this is a library that end-users are likely to be opinionated about the version of, I agree that the conditional import is a better solution than just bumping our version dependency.

The test failures represent an issue with the testing code itself, so I'm going to override them for now in the interest of getting this merged.

Super-sorry about the delayed reaction.

Issue #460, opened 2018-05-21
SparkMagic and Livy user impersonation


ranjitiyer:
### Background ###

When a Spark notebook is executed in Jupyter, SparkMagic sends code (REST API) to Livy which then creates a Spark job and submits it to a YARN cluster for execution. By default YARN jobs thus submitted run as user **livy** but many enterprise organizations want Jupyter users to be impersonated in Livy. This can be achieved by enabling Livy impersonation and adding the ```proxyUser``` property in the spark magic configuration for each user that needs to be impersonated.

```
  "session_configs": {
    "driverMemory": "1000M",
    "executorCores": 2
    "proxyUser": "bob"
  },
```

The result is of this config change is that if **bob** is the Notebook instance user, they are now also the user running the YARN application.

```
Application-Id	    Application-Name	    Application-Type	      User	     Queue
application_1526925378944_0005	      livy-session-1	               SPARK	       bob	   default
```

### Proposal ###

Since the ```proxyUser``` value cannot be known a-priori, it must be set individually for every user in their spark magic config json. This is not ideal because it increases configuration complexity for a multi-user enterprise requiring them to inject this property when a new user(s) is added to the system.

I'm proposing that SparkMagic support user impersonation by default - meaning it always sends the ```proxyUser``` with its value as the user name of process SparkMagic is running in. This elimiates configuration complexity and makes spark magic more amenable for enterprise use. Users can always override proxyUser by explicitly setting its value in ```sessions_configs``` json object and that will take precedence over the default solution of using the OS user for impersonation.

I envision this to be low complexity change in /sparkmagic/utils/configuration.py. Happy to get some feedback on this proposal.

maziyarpanahi:
This is also an issue when you launch an EMR cluster with Jupyter. When you set the Livy configuration for the cluster, in Jupyter SparkSession cannot be created:

configs:
```
[
    {
      "Classification": "livy-conf",
      "Properties": {
        "livy.impersonation.enabled": "true"
      }
    },
    {
      "Classification": "core-site",
      "Properties": {
        "hadoop.proxyuser.livy.groups": "*",
        "hadoop.proxyuser.livy.hosts": "*"
      }
    }
  ]
```

error:
```
unexpected parameter proxyUser can not be empty.
```

josechudev:
Hey! I am currently needing this change. How can I help with this issue?

PR #462, opened 2018-06-07
Use process user name as proxyUser value for NO_AUTH

ranjitiyer:
This change implements the proposal described [here](https://github.com/jupyter-incubator/sparkmagic/issues/460). It removes the need to manually have to set the proxyUser Livy session property for every user (and then further remembering to repeat the process for new users added in the future)


foscraig:
@apetresc @aggFTW, is it possible to review this PR and help with Travis CI failure?

Issue #492, opened 2018-11-16
Issue with pyspark(3) kernel: TypeError: 'Future' object is not subscriptable


mcanini:
I am having the following issue with version 0.12.6 of sparkmagic and pyspark kernels.

```
[I 00:25:07.005 NotebookApp] Kernel started: fa701309-a461-4d32-b4da-7fccb694f1f1
Traceback (most recent call last):
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.5/dist-packages/sparkmagic/kernels/pysparkkernel/pysparkkernel.py", line 28, in <module>
    IPKernelApp.launch_instance(kernel_class=PySparkKernel)
  File "/home/spark/.local/lib/python3.5/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-139>", line 2, in initialize
  File "/home/spark/.local/lib/python3.5/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/home/spark/.local/lib/python3.5/site-packages/ipykernel/kernelapp.py", line 484, in initialize
    self.init_kernel()
  File "/home/spark/.local/lib/python3.5/site-packages/ipykernel/kernelapp.py", line 389, in init_kernel
    user_ns=self.user_ns,
  File "/home/spark/.local/lib/python3.5/site-packages/traitlets/config/configurable.py", line 412, in instance
    inst = cls(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/sparkmagic/kernels/pysparkkernel/pysparkkernel.py", line 23, in __init__
    language_info, session_language, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/sparkmagic/kernels/wrapperkernel/sparkkernelbase.py", line 42, in __init__
    self._load_magics_extension()
  File "/usr/local/lib/python3.5/dist-packages/sparkmagic/kernels/wrapperkernel/sparkkernelbase.py", line 71, in _load_magics_extension
    log_if_error="Failed to load the Spark kernels magics library.")
  File "/usr/local/lib/python3.5/dist-packages/sparkmagic/kernels/wrapperkernel/sparkkernelbase.py", line 102, in _execute_cell
[...]

PR #508, opened 2019-01-25
Patch SparkMagic for latest IPythonKernel compatibility

jaipreet-s:
**Description**
* The IPython interface was updated to return an asyncio.Future rather than a dict from version 5.1.0. This broke SparkMagic as it still expects a dictionary from the output
* This change updates the SparkMagic base kernel to expect a Future and block on its result.
* This also updates the dependencies to call out the the new IPython version dependency.
* Addresses #492 

**Testing Done**
* Unit tests added
* Validating that the kernel connects successfully
* Validating some basic Spark additional operations on an EMR cluster.


jaipreet-s:
Could a maintainer have a look at this please?  The tests are breaking for for Python2 compatibiltiy, however IPython has dropped support for Python 2 since 5.0.0 [1]. Does it still make sense to support Python2 in SparkMagic?

[1] https://github.com/ipython/ipykernel/blob/master/docs/changelog.rst

Tagar:
@jaipreet-s we still have a bunch of users who are on Python 2.7.. 
IMO we need to support Python 2 and 3 as long as PySpark itself doesn't drop Python 2 support.
It seems even upcoming Spark 3.0 release doesn't have plans to drop Python 2.7 support.

sparkmagic/setup.py
```diff
@@ -85,7 +85,7 @@ def version(path):
           'pandas>=0.17.1',
           'numpy',
           'requests',
-          'ipykernel>=4.2.2,<5',
+          'ipykernel>=5.1.0,<6.0',
           'ipywidgets>5.0.0,<8.0',
           'notebook>=4.2,<6.0',
```
Tagar: if you're upgrading `ipykernel` from `4.2` to `5.x`, 
wouldn't you also need to upgrade `notebook` from `4.2` to `5.x` ?

arunasp: When bumping notebook version in test installation it leads to dependency hell of incompatible versions in other packages. Seems the JupyterHub docker image will also need to follow upgrades as well.

juliusvonkohout:
What about forking into sparkmagic3 which only supports python3, drop all python2 tests and working and the myriads of unanswered pull requests and issues? @jaipreet-s I am using a sparkmagic with notebook ipywidgets plotly ipython on latest versions. and it works fine. for python2 and python3 so far. i will test whether it just breaks test or real functionality for python2.

Tagar:
Spark still supports Python 2. 
It seems even Spark 3 release coming up in a few months has not plans of dropping Python 2 support yet.
We use both Python 3 and 2. 
Would be nice to retain support for Python 2 until end of this year at least - official EOL for Python 2.
Thank you.

juliusvonkohout:
> Spark still supports Python 2.
> It seems even Spark 3 release coming up in a few months has no plans of dropping Python 2 support yet.
> We use both Python 3 and 2.
> Would be nice to retain support for Python 2 until end of this year at least - official EOL for Python 2.
> Thank you.

My packages still work with python2 on the cluster. but this repository badly maintained and the responses are so slow that i just forked it into https://pypi.org/project/sparkmagic3/ I merged the most important and straight forward pull requests, fixed several issues and updated all packages to the latest versions possible. We are using this now in production with livy 0.4 and 0.5.

Tagar:
thanks @juliusvonkohout 

adding rest of `jupyter-incubator` custodians from https://github.com/jupyter-incubator page.
cc @jaipreet-s @Carreau @damianavila @jdfreder @parente @rgbkrk 

is it possible somebody would help with pushing some pull request `jupyter-incubator/sparkmagic` over to master? 

Sparkmagic has broken Python 3 support for over a year, and doesn't support several latest Apache Livy releases.. also there are a lot of PRs that don't get attention from committers for since January.

sparkmagic/setup.py
rgbkrk: The kernel version doesn't have to match that of the notebook, as they're designed to be separate systems.

rgbkrk:
cc @apetresc @aggFTW

If we don't get any responses, we'll see what we can do about current contributors becoming maintainers.

itamarst:
I will try to review this and get it merged.

itamarst:
I would like to get one more release out at least that supports Python 2; the current code doesn't work with Livy 0.6, so would be good to have release that does Python 2 and Livy 0.6, so people stuck with Python 2 can keep using it.

Here's the plan for Spark to drop Python 2; sounds like it's happening next year sometime:

https://spark.apache.org/news/plan-for-dropping-python-2-support.html

itamarst:
I suspect it's possible to support both old and new versions of IPython notebook, so will see if I can manage it, based on this PR.

itamarst:
I'm going to merge this into branch on the main repo, update it, and open new PR off that if I can get it working.

itamarst:
So it occurs to me that Python 2 support conflates two different issues: being able support Python 2 in the notebook, and being able to send Python 2 to Spark.

It's perfectly possible for the notebook to run on Python 3, but still send Python 2 to Spark.

Issue #518, opened 2019-03-18
SparkMagic does not correctly set cell error type


mbrio:
I've been working with Papermill and Sparkmagic, and have noticed that whenever a Spark error occurs while using Sparkmagic in a Jupyter notebook, Papermill does not throw an error which it should. I had posted this error on the Papermill github account and they responded that it may be the kernel not following correct protocols. I dug a little deeper today and found that when a Spark error occurs, the 'output_type' is `stream` and the name is `stderr`, as opposed to 'output_type' being `error`, which I believe is the correct type the cell should be marked as. Do you know if there are guidelines for errors on the IPython codebase? I would like to try and solve this problem, either by helping fix the inconsistency in sparkmagic, or papermill.

mbrio:
I realize, that this is probably because we wouldn't want all spark errors to be fatal errors when running Jupyter as a standard development notebook, but there are situations where running a notebook should cause a fatal error if a spark cell didn't execute correctly. Possibly utilizing the `configuration` class that exists in SparkMagic to ensure that all errors are treated as fatal could be a good solution, I'm looking into implementing this now.

MSeal:
For a little more context: https://github.com/nteract/papermill/issues/330

I can help with kernel messaging questions / guidance if it's needed.

mbrio:
At this point in time, I've engineered a fix that I do not know is correctly implemented or not. I had noticed that for a particular set of errors, the SparkMagic kernel raised an exception, but for Spark statement responses they never raised exceptions.

In my implementation (which I'm about to put a pull request for) I setup two configuration options, one that states that an exception should be raised (it is defaulted to should not since that is the current implementation) and the other that states whether the Livy session should be shutdown if a Spark statement fails.

@MSeal, I think the way in which SparkMagic is handling this is purposefully done, but only speaks to how the scripts should be run in a development environment, not as a production script run through Papermill.

I'm unsure if I agree with it being done this way, as even in a development environment I would want a hard stop if an exception occurred.

In any case, when my configuration is set to True the exception is correctly raised, an error cell is output, and all downstream cells halt execution.

mbrio:
@MSeal, here is my pull request with unit tests, if you want to see what I did, I'm awaiting a response from the sparkmagic team to see if there is an alternative way of handling this that jives better with their source code.

itamarst:
Fixed, should be in next release.

PR #519, opened 2019-03-18
Spark statement errors create ipython error messages when configured to do so

mbrio:
* Introduce SparkStatementException
* When the `spark_statement_errors_are_fatal` configuration is set to `True`
  and an error occurs origininating from the executing spark statement an
  exception is raised within the ipython notebook that correctly inserts an
  `error` output response.
* When the `shutdown_session_on_spark_statement_errors` configuration is set
  to `True` and `spark_statement_errors_are_fatal` is set to `True` when a
  spark statement error occurs the Livy session will be cleaned up.

In reference to: https://github.com/jupyter-incubator/sparkmagic/issues/518

### DIFF
--- a/sparkmagic/sparkmagic/magics/sparkmagicsbase.py
+++ b/sparkmagic/sparkmagic/magics/sparkmagicsbase.py
@@ -19,6 +19,7 @@ from sparkmagic.livyclientlib.sparkcontroller import SparkController
 from sparkmagic.livyclientlib.sqlquery import SQLQuery
 from sparkmagic.livyclientlib.command import Command
 from sparkmagic.livyclientlib.sparkstorecommand import SparkStoreCommand
+from sparkmagic.livyclientlib.exceptions import SparkStatementException
 
 
 @magics_class
@@ -40,6 +41,12 @@ class SparkMagicBase(Magics):
     def execute_spark(self, cell, output_var, samplemethod, maxrows, samplefraction, session_name, coerce):
         (success, out) = self.spark_controller.run_command(Command(cell), session_name)
         if not success:
+            if conf.spark_statement_errors_are_fatal():
+                if conf.shutdown_session_on_spark_statement_errors():
+                    self.spark_controller.cleanup()
+
+                raise SparkStatementException(out)
+
             self.ipython_display.send_error(out)
         else:
             self.ipython_display.write(out)
--- a/sparkmagic/sparkmagic/utils/configuration.py
+++ b/sparkmagic/sparkmagic/utils/configuration.py
@@ -237,6 +237,20 @@ def configurable_retry_policy_max_retries():
     return 8
 
 
+@_with_override
+def spark_statement_errors_are_fatal():
+    # If set to true, any spark statement errors will be considered
+    # fatal and will raise a SparkStatementException
+    return False
+
+
+@_with_override
+def shutdown_session_on_spark_statement_errors():
+    # If set to true, any spark statement errors will cause the Livy
+    # session to be cleaned up
+    return False
+
+
 def _credentials_override(f):
     """Provides special handling for credentials. It still calls _override().
     If 'base64_password' in config is set, it will base64 decode it and returned in return value's 'password' field.
--- a/sparkmagic/sparkmagic/livyclientlib/exceptions.py
+++ b/sparkmagic/sparkmagic/livyclientlib/exceptions.py
@@ -51,6 +51,10 @@ class BadUserDataException(LivyClientLibException):
 class SqlContextNotFoundException(LivyClientLibException):
     """Exception that is thrown when the SQL context is not found."""
 
+
+class SparkStatementException(LivyClientLibException):
+    """Exception that is thrown when an error occurs while parsing or executing Spark statements."""
+
 # == DECORATORS FOR EXCEPTION HANDLING ==
 EXPECTED_EXCEPTIONS = [BadUserConfigurationException, BadUserDataException, LivyUnexpectedStatusException, SqlContextNotFoundException,
                        HttpClientException, LivyClientTimeoutException, SessionManagementException]
--- a/sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
+++ b/sparkmagic/sparkmagic/tests/test_sparkmagicbase.py
@@ -1,4 +1,5 @@
 # -*- coding: UTF-8 -*-
+import sparkmagic.utils.configuration as conf
 from mock import MagicMock
 from nose.tools import with_setup, assert_equals, assert_raises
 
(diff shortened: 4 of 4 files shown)


sparkmagic/sparkmagic/magics/sparkmagicsbase.py
```diff
@@ -40,6 +41,12 @@ def __init__(self, shell, data=None, spark_events=None):
     def execute_spark(self, cell, output_var, samplemethod, maxrows, samplefraction, session_name, coerce):
         (success, out) = self.spark_controller.run_command(Command(cell), session_name)
         if not success:
+            if conf.spark_statement_errors_are_fatal():
```
MSeal: I'm not sure how the repo is treating errors but I would imagine you'd always want to raise here and let the user handle exceptions if they wishes. In the interactive case later evaluation cells could still reinteract with the session just fine.

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
```diff
@@ -40,6 +41,12 @@ def __init__(self, shell, data=None, spark_events=None):
...
+                if conf.shutdown_session_on_spark_statement_errors():
```
MSeal: This seems like a good check. Do be you know if the session is cleaned up when the process exits in ideal circumstances? If so this might be unnecessary as the process / kernel exit would resolve Livy cleanup.

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
```diff
@@ -40,6 +41,12 @@ def __init__(self, shell, data=None, spark_events=None):
...
+                    self.spark_controller.cleanup()
+
+                raise SparkStatementException(out)
+
             self.ipython_display.send_error(out)
```
MSeal: It's possible you'll always want to out out this above any exception raising.

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
mbrio: I agree, but I was worried about changing the current default functionality, I 100% agree with you though.

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
mbrio: It does not, the Livy session stays open and available.

sparkmagic/sparkmagic/magics/sparkmagicsbase.py
mbrio: Again, I agree with you, I was following how the base code here throws before display: https://github.com/jupyter-incubator/sparkmagic/blob/b9413e590d565390c517b016ba7bc436655f9a79/sparkmagic/sparkmagic/livyclientlib/command.py#L40

mbrio: This occurs at: https://github.com/jupyter-incubator/sparkmagic/blob/b9413e590d565390c517b016ba7bc436655f9a79/sparkmagic/sparkmagic/magics/sparkmagicsbase.py#L41 above my added code, above the `send_error`.

jpugliesi:
@mbrio @MSeal Any updates on the status of this PR? As a user of papermill with sparkmagic, this functionality is essential for us to run spark jobs in production via papermill.

In the meantime, how (if at all) are you using sparkmagic with papermill currently?

MSeal:
@itamarst @juliusvonkohout Can one of you review this or help guide what should be changed? I don't really know this repository well.

@jpugliesi Internally we have custom magics to connect to our internal infrastructure with all of it's conventions and patterns where I work. As a result we don't use this repo's magic in particular.

itamarst:
I will try to do some sparkmagic maintenance on Friday.

itamarst:
This looks fine to me, only thing I'd like is some documentation (in the README, say) about how to use Sparkmagic (and in particular these configuration parameters) with Papermill, so that people know they exist.

mbrio:
Is there a reason to include documentation on the README about Papermill? My example was specific to my use of your library with Papermill, but this issue is not specific to Papermill itself.

itamarst:
1. If there's no documentation, most people won't figure out how to use it.
2. Papermill seems like a good example when documenting.

juhoautio:
@mbrio thanks for working on this! What do you think about making the option for automatic cleanup to be always run on exit, no matter if successful or failed? My use case is explained here: https://github.com/nteract/papermill/issues/431.

devstein:
@itamarst @mbrio  Any updates on this? Blocked by this issue and would love to see it merged 😃

mbrio:
Hey all, I’ve been a bit tied up but should have some time tomorrow to look at @juhoautio’s use case and the documentation.

itamarst:
I can just add a bit to the README.

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -237,6 +237,20 @@ def configurable_retry_policy_max_retries():
...
 
+@_with_override
+def spark_statement_errors_are_fatal():
+    # If set to true, any spark statement errors will be considered
+    # fatal and will raise a SparkStatementException
+    return False
+
+
+@_with_override
+def shutdown_session_on_spark_statement_errors():
```
juhoautio: Is there some proper way for setting this (or any sparkmagic configuration) other than the `~/.sparkmagic/config.json` file? 

I couldn't find anything, so went with this hack for now:

```python
from sparkmagic.utils import configuration

def spark_statement_errors_are_fatal():
    return True
# replace the method in imported class:
setattr(configuration, 'spark_statement_errors_are_fatal', spark_statement_errors_are_fatal)
```

itamarst: I'll open an issue for this.

devstein:
@itamarst @mbrio 

Do we need to add `SparkStatementException` to `EXPECTED_EXCEPTIONS`? The current implementation doesn't seem to raise an exception in the iPython notebook, it only shuts down the Livy session. 

I verified my configuration with
```python
from sparkmagic.utils import configuration

print(configuration.shutdown_session_on_spark_statement_errors())
print(configuration.spark_statement_errors_are_fatal())
```

My thought is that [handle_expected_exceptions](https://github.com/jupyter-incubator/sparkmagic/blob/9edc4e17a99daee2739a6340a273d17cb944e6d3/sparkmagic/sparkmagic/livyclientlib/exceptions.py#L63) is swallowing the `SparkStatementException`

itamarst:
@devstein could you file an issue so this doesn't get lost, and I'll try to reproduce/fix when I next do sparkmagic maintenance? Thank you!

devstein:
@itamarst https://github.com/jupyter-incubator/sparkmagic/issues/578 Done!

PR #523, opened 2019-04-01
Fix guaranteed stack trace

ericdill:
I noticed that the original if/else block will always raise a stack trace. I'm not sure what the "ideal" behavior here is for logging, but the way it is currently implemented is a guaranteed NoneType has no attribute logger. 

@apetresc @aggFTW  do you think this naked `print` here is ok? Would you rather us work together to figure out how to get an active logger inject here so we can put this warning into the sparkmagic log file?

### DIFF
--- a/sparkmagic/sparkmagic/livyclientlib/livysession.py
+++ b/sparkmagic/sparkmagic/livyclientlib/livysession.py
@@ -17,8 +17,19 @@ from .exceptions import LivyClientTimeoutException, \
 
 class _HeartbeatThread(threading.Thread):
     def __init__(self, livy_session, refresh_seconds, retry_seconds, run_at_most=None):
+        """
+        Parameters
+        ----------
+        livy_session : LivySession
+        refresh_seconds: int
+            The seconds to sleep between refreshing the Livy session status and info
+        retry_seconds: int
+            The seconds to sleep before retrying on a failed refresh
+        run_at_most: int, optional
+            The max number of loops to execute before ending this thread
+        """
         super(_HeartbeatThread, self).__init__()
-        
+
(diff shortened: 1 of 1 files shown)


Tagar:
cc @itamarst and @rgbkrk

itamarst:
Could you include a test demonstrating the fix, or at least explain what this is trying to fix with an example traceback or reproducer? I'm not sure what this is trying to fix from skimming it.

sparkmagic/sparkmagic/livyclientlib/livysession.py
```diff
@@ -17,38 +17,56 @@
...
+            # a billion iterations should be sufficient
+            run_at_most = int(1e9)
         self.run_at_most = run_at_most
 
     def run(self):
-        i = 0
-        if self.livy_session is not None:
-            self.livy_session.logger.info(u'Starting heartbeat for session {}'.format(self.livy_session.id))
-        else:
-            self.livy_session.logger.info(u'Will not start heartbeat because session is none')
```
ericdill: Hi @itamarst. This is the offending line. On line 29, we are testing for `self.livy_session` being `None`. If the `else` bit of the clause gets executed, then `self.livy_session` is guaranteed to be `None` and so we will get a stack trace saying roughly "None type has no attribute 'logger'", which is a horribly cryptic and confusing error message. This PR fixes that issue, as well as adding some docs, giving better variable names and simplifying some looping logic.

itamarst: Ah hah, makes sense.

itamarst:
This definitely seems like an improvement, so going to merge it. Thank you!

PR #542, opened 2019-06-25
Support newer ipykernels

itamarst:
Fixes #492 

In theory this should work with both Python 2 (using ipykernel 4.10) and Python 3 (newer ipykernels).

Thanks to @jaipreet-s for the bulk of the code.

Again, would appreciate some user testing—does this work for you with Python 2? Python 3? What version ipykernel do you end up with (`pip list | grep ipykernel`).

Please REPLY WITH A COMMENT, a thumbs up isn't very informative.

PR #544, opened 2019-06-26
Update dependencies

juliusvonkohout:
With these dependencies there are no package downgrades if you install sparkmagic after e.g jupyterlab>=0.35.6  or jupyter notebook with python3 or python2 @itamarst


Tagar:
Both Spark and Livy still support Python 2.
Yes, Python 2 EOLs end of this year, and so perhaps we can drop support for Python 2 in Jan 2020?
Even Spark 3.0 release will still work with Python 2, although it'll be just tagged as "deprecated".
Might be best on just hold off with this until Python 2 actually EOL?
This concerns me as our users still have a mix of Python 2 and 3 notebooks.
Thank you.

itamarst:
There are two sets of changes here:

1. Removing the maximum version.
2. Raising the minimum version.

The latter is what makes Python 2 support untenable... and it's also unnecessary. Merely removing the maximum version should fix any issues with newer Jupyter, because pip uses _newest_ available package that meets constraints.

As far as Python 2 support, I want to do at least one reelase that (A) works on Python 2 (B) works with Livy 0.6. So I don't want to drop it yet. However, as mentioned above you don't need to drop Python 2 to solve this problem.

So before this is merged it should remove the increase in minimal versions. Also this should build on #542 which will ensure tests pass (or at least reduce breakage) when maximum version is removed.

juliusvonkohout:
> There are two sets of changes here:
> 
>     1. Removing the maximum version.
> 
>     2. Raising the minimum version.
> 
> 
> The latter is what makes Python 2 support untenable... and it's also unnecessary. Merely removing the maximum version should fix any issues with newer Jupyter, because pip uses _newest_ available package that meets constraints.
> 
> As far as Python 2 support, I want to do at least one reelase that (A) works on Python 2 (B) works with Livy 0.6. So I don't want to drop it yet. However, as mentioned above you don't need to drop Python 2 to solve this problem.
> 
> So before this is merged it should remove the increase in minimal versions. Also this should build on #542 which will ensure tests pass (or at least reduce breakage) when maximum version is removed.

@itamarst Thanks, i also had to change other minimum versions, but it builds now for python2 and python3.

itamarst:
Will get back to this and other PRs when I'm back from vacation.

itamarst:
I ended up removing the max constraints in #542 (testing still ongoing), so I don't think this is needed anymore. Thanks for the push to remove maximum constraints too, hopefully #542 will do what you want.

juliusvonkohout:
> I ended up removing the max constraints in #542 (testing still ongoing), so I don't think this is needed anymore. Thanks for the push to remove maximum constraints too, hopefully #542 will do what you want.

@itamarst Plesse not. Plotly is still limited in your branch and we should upgrade to the latest python2 compatible packages as a minimum requirement like i did here.

sparkmagic/setup.py
```diff
@@ -79,16 +79,16 @@ def version(path):
...
-          'numpy',
-          'requests',
-          'ipykernel>=4.2.2,<5',
-          'ipywidgets>5.0.0,<8.0',
-          'notebook>=4.2,<6.0',
-          'tornado>=4',
-          'requests_kerberos>=0.8.0'
+          'ipython>=5.8.0,<8',
+	  'nose>=1.3.7',
+	  'mock>=3.0.5',
```
Tagar: is this known for sparkmagic to be broken on `mock` 3.0.4?

sparkmagic/setup.py
```diff
@@ -79,16 +79,16 @@ def version(path):
...
+	  'pandas>=0.24.2',
```
Tagar: same - is this known for `sparkmagic` to be broken on 0.24.1 and earlier versions ?

itamarst:
1. There is no need to upgrade minimum versions. pip will _automatically_ choose the latest version.
2. plotly is still an issue, yes, but I've seen indicates of API changes. I'll open a new ticket for that.

itamarst:
To clarify why I think changing minimum versions is unnecessary:

```
(virtualenv) [itamarst@blake tmp]$ echo "ipython>3" > requirements.txt
(virtualenv) [itamarst@blake tmp]$ cat requirements.txt 
ipython>3
(virtualenv) [itamarst@blake tmp]$ pip install -r requirements.txt
....
(virtualenv) [itamarst@blake tmp]$ pip list | grep ipython
ipython          7.6.1
```

There's no point in changing minimum versions to be higher, it'll just break some people's installs and adds no benefit unless we know we actually require higher version.

juliusvonkohout:
@Tagar i am using this dependencies with jupyterlab 0.35 and 1.0.1 against python2 and python3 and livy 0.4 and 0.6 with plotly 3.9.x. I do not use Zeppelin. We should go for the latest versions that work with zeppelin then. 

@itamarst it reduces the maintenance burden. I do not know whether the current minimum versions work together (no matter whether pip thinks that it works), for the ones in this pull request i know it. Or did someone test all possible combinations? Updating the minimum version significantly decreases the number of possible (mis)configurations. But if you do not want to update the minimum versions and trust in pip, please at least unconstrain plotly then. Otherwise i still have to maintain my fork.

Tagar:
@juliusvonkohout 
>> We should go for the latest versions that work with zeppelin then.

I wasn't suggesting to do that at `sparkmagic` level. 
We can easily constrain versions at pip/conda level at install time what works for us.
Zeppelin was just an example. 
If we constrain for Zeppelin, it may breaks somebody else's requirements. 
Again.. folks can easily constrain versions at pip/conda level at install time what works for their environments ... as long as `sparkmagic` itself is not overly restrictive which versions it has put in requirements.

Thank you.

sparkmagic/setup.py
```diff
@@ -79,16 +79,16 @@ def version(path):
...
+	  'numpy>=1.16.3',
+	  'requests>=2.21.0',
+	  'ipykernel>=4.10.0,<6',
+	  'ipywidgets>=7.4.2,<8.0',
+	  'notebook>=5.7.8,<6.0',
+	  'tornado>=5.1.1,<6',
```
Tagar: this `tornado` requirement will break zeppelin for us.
I would not go with this change if possible. 
all other requirements have to be reviewed carefully if they are really necessary. 
thank you

itamarst:
I switched to plotly v3 in #554.

PR #487, opened 2018-11-04
add magic for matplotlib display(pyspark or pyspark3)

wangqiaoshi:
now sparkmagic must be pass data to local  then matplot.  
it is expensive if the data is big

PR #541, opened 2019-06-25
Add progress bar to sparkmagic/sparkmagic/livyclientlib/command.py.

juliusvonkohout:
Tested with livy 0.4-0.6, python2 and python3 superseedes #441 @itamarst 

progress-bar

PR #552, opened 2019-07-12
Server side image rendering

itamarst:
Based on #487.

* [x] Works in PySpark kernel.
* [x] Works in normal notebook.
* [x] Unit tests.
* [ ] Real-world testing by someone else.


itamarst:
Testing is a bit iffy, but honestly only real-world use tests this actually works, and there's no good framework for end-to-end tests at the moment.

sparkmagic/sparkmagic/livyclientlib/command.py
```diff
@@ -65,7 +68,14 @@ def _get_statement_output(self, session, statement_id):
                     return (True, u"")
 
                 if statement_output[u"status"] == u"ok":
-                    return (True, statement_output[u"data"][u"text/plain"])
+                    data = statement_output[u"data"]
+                    png_encoded = data.get("image/png")
+                    if png_encoded:
+                        image = Image(base64.b64decode(png_encoded))
+                        return (True, image)
+                    else:
```
Tagar: IPython.display.Image should also support png/gif?

itamarst: Yeah, probably, but Livy specifically only returns PNG at the moment.

itamarst:
@Tagar does it work for you? There's examples in the notebooks.

Tagar:
@itamarst I am wrapping up my work projects as I am preparing to go for an extended OOO
so don't have capacity to test this unfortunately. 

It would be great if you could merge this since you already did some testing.
More folks can test this as part of a release and 
we can submit fixes for any outstanding issues as necessary later?

Thank you for all those improvements you're submitting. 
Glad to see many of those long-standing spark-magic issues are finally getting resolved.

itamarst:
I expect to be doing more work on Sparkmagic Thursday or Friday, which will include a new release with this and #541.

Tagar:
Both are awesome improvements. Thanks a lot

itamarst:
Thank you @qiaoshi.wang for the original patch!

PR #576, opened 2019-10-02
Support Client Side Cert for Livy

stonezhong:
Our livy is deployed in such a way:
It has a whitelist of CNs and require client to present a client side certificate, only client cert whose CN match the whitelist will be allow to access livy.

I am adding 3 configs to kernel_{language}_credentials:
ssl_client_cert, ssl_client_key and ssl_verify

I have the following config:
```
{
  "kernel_python_credentials" : {
    "username": "my_username",
    "password": "mu_password",
    "url": "https://my_livy_url/",
    "auth": "Basic_Access",
    "ssl_client_cert": "/root/.sparkmagic/sso_client.crt",
    "ssl_client_key":  "/root/.sparkmagic/sso_client.key",
    "ssl_verify":      "/root/.sparkmagic/CombinedDigicertCA.cer"
  },
...
```
Tested it and the change worked for me.

### DIFF
--- a/sparkmagic/sparkmagic/kernels/kernelmagics.py
+++ b/sparkmagic/sparkmagic/kernels/kernelmagics.py
@@ -18,7 +18,7 @@ from sparkmagic.utils.utils import parse_argstring_or_throw, get_coerce_value
 from sparkmagic.utils.sparkevents import SparkEvents
 from sparkmagic.utils.constants import LANGS_SUPPORTED
 from sparkmagic.livyclientlib.command import Command
-from sparkmagic.livyclientlib.endpoint import Endpoint
+from sparkmagic.livyclientlib.endpoint import Endpoint, SSLInfo
 from sparkmagic.magics.sparkmagicsbase import SparkMagicBase
 from sparkmagic.livyclientlib.exceptions import handle_expected_exceptions, wrap_unexpected_exceptions, \
     BadUserDataException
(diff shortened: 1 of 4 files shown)


sparkmagic/sparkmagic/livyclientlib/endpoint.py
```diff
@@ -21,13 +22,37 @@ def __init__(self, url, auth, username="", password="", implicitly_added=False):
...
-        return hash((self.url, self.username, self.password, self.auth))
+        return hash((self.url, self.username, self.password, self.auth, self.ssl_info))
 
     def __ne__(self, other):
         return not self == other
 
     def __str__(self):
         return u"Endpoint({})".format(self.url)
+
+class SSLInfo(object):
```
itamarst: Add a docstring explain what this class does, and what the parameters are.

itamarst: Also, this seems like a good place to start using `attrs` (http://www.attrs.org/en/stable/), it'll make this class much much shorter.

sparkmagic/sparkmagic/kernels/kernelmagics.py
```diff
@@ -389,7 +389,12 @@ def matplot(self, line, cell="", local_ns=None):
     def refresh_configuration(self):
         credentials = getattr(conf, 'base64_kernel_' + self.language + '_credentials')()
         (username, password, auth, url) = (credentials['username'], credentials['password'], credentials['auth'], credentials['url'])
-        self.endpoint = Endpoint(url, auth, username, password)
+        (ssl_client_cert, ssl_client_key, ssl_verify) = (credentials.get('ssl_client_cert'), credentials.get('ssl_client_key'), credentials.get('ssl_verify'),)
```
itamarst: Could we say "TLS" instead of "SSL" everywhere? It's been TLS for 20 years now :)

sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py
```diff
@@ -61,15 +61,37 @@ def _send_request_helper(self, url, accepted_status_codes, function, data, retry
             try:
                 if self._endpoint.auth == constants.NO_AUTH:
                     if data is None:
```
itamarst: This now has a huge amount of duplicate paths and code, which makes me worry about errors. Can it be simplified? E.g.

```
if data is not None:
    data = json.dumps(data)
```

at the top would simplify it quite a lot, and so on.

sparkmagic/sparkmagic/livyclientlib/endpoint.py
```diff
@@ -21,13 +22,37 @@ def __init__(self, url, auth, username="", password="", implicitly_added=False):
...
+    def __init__(self, client_cert, client_key, ssl_verify):
```
itamarst: Having two different places where youo set `ssl_verify`, this and the config, bothers me a little. Is there a reason to have it here and not just rely on the config?

PR #577, opened 2019-10-03
Use JSON true instead of Python True in README Papermill Section

devstein:
Please correct me if I am wrong, but the README says to use Python's `True` in a JSON file, which is invalid, so I updated it to use the valid JSON `true`.


itamarst:
Thank you!

PR #581, opened 2019-10-12
SQLQuery default print error instead of throwing error stack

stczwd:
The default result of the Spark failure is the error message, no exception throwing until user sets spark_statement_errors_are_fatal. The result of running SQL needs this, too. For most users, the error message is sufficient.

### DIFF
--- a/sparkmagic/sparkmagic/livyclientlib/sqlquery.py
+++ b/sparkmagic/sparkmagic/livyclientlib/sqlquery.py
@@ -5,7 +5,7 @@ import sparkmagic.utils.configuration as conf
 import sparkmagic.utils.constants as constants
 from sparkmagic.utils.sparkevents import SparkEvents
 from .command import Command
-from .exceptions import DataFrameParseException, BadUserDataException
+from .exceptions import DataFrameParseException, BadUserDataException, SparkStatementException
 
 
 class SQLQuery(ObjectWithGuid):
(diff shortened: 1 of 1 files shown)


stczwd:
cc @itamarst

PedroRossi:
Hello @stczwd I am studying the issue on throwing errors instead of printing them on the spark kernel (because I really like/use this project :) ), do you think it could be better if this https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/sqlquery.py#L48 could return something like this https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/magics/sparkmagicsbase.py#L44 and the whole idea of 
```python
if not success:
    if conf.spark_statement_errors_are_fatal():
        if conf.shutdown_session_on_spark_statement_errors():
            self.spark_controller.cleanup()
        raise SparkStatementException(out)
```
could be extracted into another function instead of having the same code duplicated in different files.This whole logic also could be extracted into another function in a different file and become easily reusable, but my whole point is to question where this error throwing should be centralized, I think it should be centralized on the `sparkmagicsbase` file since he is the one that controls what appears on the screen and now if the errors are fatal instead of only showing and if the session should be shut down upon statement error.

stczwd:
@PedroRossi Yeap, this is a good suggestion. I will check it out.

stczwd:
@PedroRossi This is a great idea, but it need a lot of changes. I will submit this in a new patch, which we can discuss in it.

itamarst:
Could you describe the status of this PR now? The goal again, and what it tries to do?

stczwd:
@itamarst  This patch is mainly used to control throwing exception in sqlqueries. There will be another patch to handle the problem @PedroRossi talked with me.

sparkmagic/sparkmagic/livyclientlib/sqlquery.py
```diff
@@ -49,13 +51,17 @@ def execute(self, session):
...
         try:
             command = self.to_command(session.kind, session.sql_context_variable_name)
             command_guid = command.guid
             (success, records_text, mimetype) = command.execute(session)
             if not success:
-                raise BadUserDataException(records_text)
-            result = records_to_dataframe(records_text, session.kind, self._coerce)
+                if conf.spark_statement_errors_are_fatal():
+                    raise SparkStatementException(records_text)
+                self.ipython_display.send_error(records_text)
```
devstein: If `conf.spark_statement_errors_are_fatal() == False` and a `SparkStatementException` is raised then wouldn't this be handled and displayed by `handle_expected_exceptions`?

Issue #578, opened 2019-10-08
No iPython Error Thrown When configuration.spark_statement_errors_are_fatal() is True


devstein:
Refer to original comment https://github.com/jupyter-incubator/sparkmagic/pull/519#issuecomment-539150332

The current implementation in https://github.com/jupyter-incubator/sparkmagic/pull/519 doesn't seem to raise an exception when executing a `jupyter` notebook with `papermill`. In my notebook it only shut down the `livy` session

- `shutdown_session_on_spark_statement_errors`
- `spark_statement_errors_are_fatal`

I used to the following code to verify my configuration was correct
```python
from sparkmagic.utils import configuration

print(configuration.shutdown_session_on_spark_statement_errors()) # True
print(configuration.spark_statement_errors_are_fatal()) # True
```

PedroRossi:
I also have the same issue when using papaermill with the sparkkernel.

I tried removing the `wrap_unexpected_exceptions` decorator from the `sparkmagic.kernels` (https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/kernels/kernelmagics.py#L219) and it fixed the error throw on the sparkkernel, I also noticed that this decorator is only used on the `sparkmagic.kernels` and not on the `sparkmagic.magic` (https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/magics/remotesparkmagics.py#L68), is this function really needed to some other motive that I am not understanding? @itamarst

devstein:
@itamarst Any updates on this?

itamarst:
Hi,

Sorry for the delay, I hope to get some work done on SparkMagic in the next couple of weeks. If someone submits a PR (with a test) that'll speed things up.

devstein:
@itamarst Happy to make a PR with a test for this, but do you mind answering @PedroRossi question?

itamarst:
I don't yet know the code well enough to answer offhand... I'll answer when I next to maintenance, if you don't figure it out.

PR #585, opened 2019-10-28
all_errors_are_fatal Configuration to Throw All Expected and Unexpected Exceptions  

devstein:
Closes https://github.com/jupyter-incubator/sparkmagic/issues/578

Creates an `all_errors_are_fatal` configuration to throw any error when executing a notebook. This is useful for executing notebooks via `papermill` in a production setting where an error in any notebook cell should cause the notebook to fail with `exit code 1`

CC @itamarst @PedroRossi


itamarst:
This seems like it would fix the problem, yes, but the additional feature also sounds good if you want to add another flag with another PR.

Alternatively, insofar as `spark_statement_errors_are_fatal` never actually worked, I'd be happy to accept the `all_errors_are_fatal` PR instead of this one.

itamarst:
So next step: either fix failing test, or, change PR to replace it with more general feature (all errors cause processing to stop).

itamarst:
And, what I should've said first: thank you for the PR!

PedroRossi:
@devstein thanks for the `all_errors_are_fatal` catch, this is perfect for production environments using papermill (probably we are doing the same thing :smile: )

devstein:
@PedroRossi Of course! 

@itamarst Update to use a `all_errors_are_fatal` configuration. Do you mind helping with the import error in the `python 2.7.13` test?

README.md
```diff
@@ -81,7 +81,8 @@ If you want Papermill rendering to stop on a Spark error, edit the `~/.sparkmagi
 ```json
 {
     "spark_statement_errors_are_fatal": true,
```
itamarst: I believe this particular config option was added specifically for the Papermill use case, but doesn't actually work—maybe it should be ripped out?

devstein:
@itamarst Good call, do you have a preference on how I removed the unnecessary `spark_errors_are_fatal` flag?


Option 1: Always raise `SparkStatementException` and handle as an expected exception

1. Removed `spark_errors_are_fatal` flag
2. Always raise a `SparkStatementException` on spark errors
3. By default, `SparkStatementException` is an expected exception and is displayed in the notebook as an error 
4. If all_errors_are_fatal flag is enabled, `SparkStatementException`, like all exceptions, will not be caught


Option 2: Only raise `SparkStatementException` on `all_errors_are_fatal`

1. Removed `spark_errors_are_fatal` flag
2. By default output spark errors do not raise an exception and are outputted to the notebook (`ipython_display.send_error(out)`) 
3. If `all_errors_are_fatal` flag is enabled, spark errors will raise a `SparkStatementException`, and like all exceptions, it will not be caught

I prefer option 1 because I don't like the idea of always swallowing spark errors and then raising a new error when an unrelated flag `all_errors_fatal` is set.  But I don't know how raising `SparkStatementException` will affect other parts of the codebase that don't `handle_expected_exceptions`

devstein:
@itamarst I updated the code and the tests to `Option 1` above. Let me know if you need me to make changes.

itamarst:
Thank you! Taking a look now.

itamarst:
@devstein 

Would you be interested in getting commit access to the repo and becoming a maintainer?

You need to:

1. Agree to the code of conduct.
2. Follow best practices (automated tests, manual test, docs, reviews) when adding code.

devstein:
@itamarst 

I'm definitely interested in becoming a maintainer.

Beyond the code of conduct and maintaining best practices, what are the time expectations? How are features and issues prioritized? Is there a feature roadmap or long-term priorities?

itamarst:
@devstein right now I just do a few hours a month, so it can be as little or as many as you want. Given limited time I focus on PRs from others, mostly.

I feel like long-term priority is maintainability, most important being testing which is far too manual at the moment. I opened some issues here: https://github.com/jupyter-incubator/sparkmagic/milestone/7

devstein:
@itamarst Okay, that makes sense. I'm happy to help

PR #595, opened 2019-11-28
New conf option 'cleanup_all_sessions_on_exit'

juhoautio:
New conf option `cleanup_all_sessions_on_exit` (boolean)
- Default: `False`

The existing `shutdown_session_on_spark_statement_errors` conf allows ensuring that livy sessions are shut down only in case of a Spark error.

This new option helps ensuring that livy sessions are not left behind also when there's a non-Spark error, or if the notebook exits gracefully but user forgot to do cleanup.

----

Also added docs to README on how to set conf overrides programmatically.

----

Tested this manually with a minimal Notebook

Näyttökuva 2019-11-29 kello 0 12 38

I also tested starting a session with `%spark add -s smoke -l python -u http://my-host:8998` and shutting down the notebook caused Livy session to be stopped, as desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,35 @@ If you want Papermill rendering to stop on a Spark error, edit the `~/.sparkmagi
 }
 ```
 
+If you want any registered livy sessions to be cleaned up on exit regardless of whether the process exits gracefully or not, you can set:
+ 
+```json
+{
+    "cleanup_all_sessions_on_exit": true,
+    "all_errors_are_fatal": true
+}
+```
+
+### Conf overrides in code
+
+In addition to the conf at `~/.sparkmagic/config.json`, sparkmagic conf can be overridden programmatically in a notebook.
+
+For example:
+```python
+import sparkmagic.utils.configuration as conf
+conf.override('cleanup_all_sessions_on_exit', True)
+```
+
+Same thing, but referencing the conf member: 
+
+```python
+conf.override(conf.cleanup_all_sessions_on_exit.__name__, True)
+```
+
+NOTE: override for `cleanup_all_sessions_on_exit` must be set _before_ initializing sparkmagic ie. before this:
+
+    %load_ext sparkmagic.magics
+
 ## Docker
 
 The included `docker-compose.yml` file will let you spin up a full
--- a/sparkmagic/sparkmagic/livyclientlib/sessionmanager.py
+++ b/sparkmagic/sparkmagic/livyclientlib/sessionmanager.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2015  aggftw@gmail.com
 # Distributed under the terms of the Modified BSD License.
+import atexit
 from sparkmagic.utils.sparklogger import SparkLog
 from sparkmagic.livyclientlib.exceptions import SessionManagementException
 from sparkmagic.utils.constants import MAGICS_LOGGER_NAME
@@ -7,10 +8,13 @@ import sparkmagic.utils.configuration as conf
 
 
 class SessionManager(object):
-    def __init__(self):
+    def __init__(self, ipython_display):
         self.logger = SparkLog(u"SessionManager")
+        self.ipython_display = ipython_display
 
         self._sessions = dict()
+
+        self._register_cleanup_on_exit()
 
     @property
     def sessions(self):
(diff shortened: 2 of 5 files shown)


sparkmagic/sparkmagic/livyclientlib/sessionmanager.py
```diff
@@ -71,3 +75,16 @@ def _remove_session(self, name):
...
+    def _register_cleanup_on_exit(self):
+        """
+        Stop the livy sessions before python process exits for any reason (if enabled in conf)
+        """
+        if conf.cleanup_all_sessions_on_exit():
+            def cleanup_spark_sessions():
+                try:
+                    self.clean_up_all()
+                except Exception:
+                    pass
```
itamarst: Should probably log the error here.

README.md
```diff
@@ -90,6 +90,35 @@ If you want Papermill rendering to stop on a Spark error, edit the `~/.sparkmagi
...
+If you want any registered livy sessions to be cleaned up on exit regardless of whether the process exits gracefully or not, you can set:
+ 
+```json
+{
+    "cleanup_all_sessions_on_exit": true,
+    "all_errors_are_fatal": true
+}
+```
+
+### Conf overrides in code
```
itamarst: Thank you for documenting this!

itamarst:
Once the above lands, would you be interested in getting commit access to the repo and becoming a maintainer?

You need to:

1. Agree to the code of conduct.
2. Follow best practices (automated tests, manual test, docs) which you've done here already.

sparkmagic/sparkmagic/livyclientlib/sessionmanager.py
juhoautio: Ok, done in the latest commit

juhoautio:
@itamarst thanks for that! Right now I don't see any further contributions coming from myself in the near future. But of course I would be very much honoured to be granted those privileges. Considering this, if it still makes sense from your perspective, where can I find the code of conduct?

itamarst:
Thank you for the PR!

Apparently I never added a code of conduct, oops. PR is here: https://github.com/jupyter-incubator/sparkmagic/pull/608

itamarst:
And, you wouldn't necessarily have to contribute code—just reviewing PRs from others is helpful.

juhoautio:
Cool! I agree to the [code of conduct](https://github.com/jupyter-incubator/sparkmagic/blob/c4da58cf69b16bf202f91aba3793eb277df52197/CODE_OF_CONDUCT.md) and would be glad to join as a maintainer.

Thank you for reviewing and including this PR.

jpugliesi:
@juhoautio I am curious how you are using this feature in practice - in your setup, are there multiple sparkmagic users all submitting jobs to a shared livy server? If so, a common issue is that you dont actually want any single user's notebook to delete _everyone else's_ session on the livy server ([as mentioned here](https://github.com/jupyter-incubator/sparkmagic/issues/649#issuecomment-632838914)). 

Am I understanding sparkmagic's behavior correctly - When a given notebook runs the `clean_up_all()` function, is it intended to actually delete _every_ livy session on the server, even the ones not created by the notebook in question?

juhoautio:
@jpugliesi sorry for the late response, I didn't notice your comment before.

> in your setup, are there multiple sparkmagic users all submitting jobs to a shared livy server

Yes.

> Am I understanding sparkmagic's behavior correctly - When a given notebook runs the clean_up_all() function, is it intended to actually delete every livy session on the server, even the ones not created by the notebook in question?

No, that's not what `clean_up_all()` does – or then I would be horribly mistaken.

It only cleans the sessions that were created by the notebook in question.

----

We can examine the code to be sure.

What does `clean_up_all()` do?

https://github.com/jupyter-incubator/sparkmagic/blob/85a9067500a4a44e4c7b5732df7f4155968d7cce/sparkmagic/sparkmagic/livyclientlib/sessionmanager.py#L67-L69

What does `get_sessions_list()` do?

https://github.com/jupyter-incubator/sparkmagic/blob/85a9067500a4a44e4c7b5732df7f4155968d7cce/sparkmagic/sparkmagic/livyclientlib/sessionmanager.py#L23-L24

How is `self._sessions` populated?

https://github.com/jupyter-incubator/sparkmagic/blob/85a9067500a4a44e4c7b5732df7f4155968d7cce/sparkmagic/sparkmagic/livyclientlib/sessionmanager.py#L29-L34

What calls `add_session`? I think it's this:

https://github.com/jupyter-incubator/sparkmagic/blob/814dc52fd0b696382530d1d3c4827c68526374e2/sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py#L90-L91

Issue #614, opened 2020-01-07
Missing Customization on Kerberos Auth Configuration


PedroRossi:
Due to the following line `self._auth = HTTPKerberosAuth(mutual_authentication=REQUIRED)` in this class https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/reliablehttpclient.py#L16, the Kerberos auth constructor is not able to be configured which causes the sparkmagic to always sends 2 requests to Livy, the first sending 401 and the second 200 since the cookie was issued, but since we already have the cookie due to the first 401 why we are getting a new one everytime?

An ideal approach would be, to be able to pass custom args to the kerberos auth constructor  through the `config.json` and to change the `reliablehttpclient.py` to use the requests session (https://requests.readthedocs.io/en/master/user/advanced/#session-objects) instead of plain `requests`

Here is the kerberos auth constructor for reference to a possible solution
```python
class HTTPKerberosAuth(AuthBase):
    """Attaches HTTP GSSAPI/Kerberos Authentication to the given Request
    object."""
    def __init__(
            self, mutual_authentication=REQUIRED,
            service="HTTP", delegate=False, force_preemptive=False,
            principal=None, hostname_override=None,
            sanitize_mutual_error_response=True, send_cbt=True):
```

I am submitting this issue to see if anyone is already working on it, if not my plan is to submit a PR for this problem

PedroRossi:
I have found another issue related to this problem, I don't know why this method is static since it is only used in this class https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py#L113
```python
    @staticmethod
    def _http_client(endpoint):
        return LivyReliableHttpClient.from_endpoint(endpoint)
```
but since the `SparkController` is calling this method for every cell we end up changing the cookie for every cell which again is not needed since in the first request we already have a valid cookie after the first negotiation with kerberos

PR #461, opened 2018-05-25
Kerberos authentication parameters (#388)

silmor:
Allow kerberos mutual authentication and hostname override to be configured via config-file and widgets.

PR #616, opened 2020-01-10
Feat/kerberos auth customization

PedroRossi:
As addressed in issue #614 this PR proposes the following changes
- Using http clients maps on the `SparkController` class to reuse clients since it is not needed a new client per cell
- Using `requests.Session` instead of plain `requests` to ensure we store the generated Kerberos ticket cookie instead of negotiating the ticket in every request
- Add a configuration option to override the default `HTTPKerberosAuth` constructor

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,28 @@ Kerberos support is implemented via the [requests-kerberos](https://github.com/r
 
 Currently, sparkmagic does not support passing a kerberos principal/token, but we welcome pull requests.
 
+### Kerberos Configuration
+
+By default the `HTTPKerberosAuth` constructor provided by the `requests-kerberos` package will use the following configuration
+```python
+HTTPKerberosAuth(mutual_authentication=REQUIRED)
+```
+but this will not be right configuration for every context, so it is able to pass custom arguments for this constructor using the following configuration on the `~/.sparkmagic/config.json`
+```json
+{
+    "kerberos_auth_configuration": {
+        "mutual_authentication": 1,
+        "service": "HTTP",
+        "delegate": false,
+        "force_preemptive": false,
+        "principal": "principal",
+        "hostname_override": "hostname_override",
+        "sanitize_mutual_error_response": true,
+        "send_cbt": true
+    }
+}
+```
+
 ## Papermill
 
 If you want Papermill rendering to stop on a Spark error, edit the `~/.sparkmagic/config.json` with the following settings:
--- a/README.md
+++ b/README.md
@@ -77,8 +77,6 @@ Sparkmagic supports:
 
 Kerberos support is implemented via the [requests-kerberos](https://github.com/requests/requests-kerberos) package. Sparkmagic expects a kerberos ticket to be available in the system. Requests-kerberos will pick up the kerberos ticket from a cache file. For the ticket to be available, the user needs to have run [kinit](https://web.mit.edu/kerberos/krb5-1.12/doc/user/user_commands/kinit.html) to create the kerberos ticket.
 
-Currently, sparkmagic does not support passing a kerberos principal/token, but we welcome pull requests.
-
 ### Kerberos Configuration
 
 By default the `HTTPKerberosAuth` constructor provided by the `requests-kerberos` package will use the following configuration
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -79,6 +79,28 @@ Kerberos support is implemented via the [requests-kerberos](https://github.com/r
 
 Currently, sparkmagic does not support passing a kerberos principal/token, but we welcome pull requests.
```
devstein: Is this still the case?

sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
```diff
@@ -110,6 +114,7 @@ def _livy_session(http_client, properties, ipython_display,
         return LivySession(http_client, properties, ipython_display,
                            session_id, heartbeat_timeout=conf.livy_server_heartbeat_timeout_seconds())
 
-    @staticmethod
-    def _http_client(endpoint):
-        return LivyReliableHttpClient.from_endpoint(endpoint)
+    def _http_client(self, endpoint):
+        if endpoint not in self._http_clients[endpoint]:
```
devstein: Shouldn't this be

```python
if endpoint not in self._http_clients:
```

README.md
PedroRossi: I wasn't able to confirm it since my environment does not use this configuration, but since the constructor accepts a principal I would say it is ok now

sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
PedroRossi: Yes!

devstein:
@PedroRossi From a Kerberos functionality perspective, how does this PR relate to https://github.com/jupyter-incubator/sparkmagic/pull/461/files

PedroRossi:
@devstein the PR you mentioned is already included in mine since I allow changing `mutual_authentication` and `hostname_override` also, but I don't know why it added the option to configure this on the magic widget to add endpoints since organizations usually don't have more than one Kerberos environment which also makes me think it is not a good choice to add Kerberos related config to the `Endpoint` class constructor

itamarst:
@devstein FYI:

1. I've never used Spark for anything in the real world.
2. My knowledge of Sparkmagic codebase isn't particularly extensive.

So I am not An Expert™ :grinning:

sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
```diff
@@ -10,6 +10,10 @@
 
 
 class SparkController(object):
+    # this is to reuse the already created http clients
+    # since the reliablehttpclient uses requests session
+    _http_clients = {}
```
itamarst: Seems to better to have this be an attribute of `SparkController` instances, rather than class attribute? Can see class attribute leading to bugs, or at least confusion.

itamarst:
Other than that one comment seems fine to me, thanks.

sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
PedroRossi: Yes, it is better to have it as an instance attribute

PedroRossi:
> @devstein FYI:
> 
>     1. I've never used Spark for anything in the real world.
> 
>     2. My knowledge of Sparkmagic codebase isn't particularly extensive.
> 
> 
> So I am not An Expert™ grinning

PedroRossi:
@itamarst ended up closing and reopening the PR by mistake, but I just fixed the comment you mentioned

PedroRossi:
@devstein is the PR ok now?

itamarst:
@PedroRossi any interest in becoming a SparkMagic maintainer? Just need to agree to follow the code of conduct and development guidelines (testing etc.).

PedroRossi:
@itamarst I have interest in becoming a maintainer!

PedroRossi:
@itamarst any follow up on becoming a maintainer?

itamarst:
Oops, missed your answer. Will add you now, welcome onboard!

PR #617, opened 2020-01-14
Raise on Failed Session Start if all_errors_are_fatal

devstein:
Currently, `all_errors_are_fatal` was not functioning as expected if the session failed to start because `_do_not_call_start_session` swallowed the exception. This adds maintains existing functionality but also raises the exception if `all_errors_are_fatal` is `True`

- Raise session start exceptions 
- Add regression test


itamarst:
Thank you!

juhoautio:
Sweet! @itamarst: are there any plans on making a new release that would include this?

devstein:
@juhoautio Will be included in the `0.15.0` release

PR #628, opened 2020-02-10
Add a cell magic to cancel statement execution

LinanZheng:
SparkMagic currently does not have the ability to cancel the execution of certain statement and therefore it is quite problematic if the user accidentally wrote runaway code or undesirable long running Spark operations. In this PR, a new cell magic is introduced to cancel either all running/waiting statements from a session (Like SIGINT) or cancel one specific statement from a session.

This is the first draft of what I proposed above and at this moment, the PR is a working prototype stage that you can try out yourself but here are things I would love to get some suggestions/helps/reviews on:
 
1. Cell magic arguments. Are there additional user cases that I haven't thought about? Is the force argument necessary?
2. I implemented the new magic command in KernelMagics but I can see this being ported over to remotesparkmagic as a subcommand. Would you like to see this being migrated over? 

Let me know how do you think of this magic and its design. Thanks!

TODO:
- [ ] Help HTML in KernelMagic
- [ ] Unit Tests
- [ ] RemoteSparkMagic Implementation?
- [ ] Usage in example notebook
- [ ] New issue to track SIGINT callback

### DIFF
--- a/sparkmagic/sparkmagic/kernels/kernelmagics.py
+++ b/sparkmagic/sparkmagic/kernels/kernelmagics.py
@@ -338,6 +338,30 @@ class KernelMagics(SparkMagicBase):
         else:
             self.ipython_display.send_error(u"Include the -f parameter if you understand that all statements executed "
                                             u"in this session will be lost.")
+                                            
+    @magic_arguments()
+    @cell_magic
+    @argument("-f", "--force", type=bool, default=False, nargs="?", const=True, help="If present, user understands.")
+    @argument("-s", "--session", type=int, help="Which session to cancel.")
+    @argument("-t", "--statement", type=int, help="Which statement to cancel.")
+    @wrap_unexpected_exceptions
+    @handle_expected_exceptions
+    @_event
+    def cancel(self, line, cell="", local_ns=None):
+        self._assure_cell_body_is_empty(KernelMagics.cancel.__name__, cell)
+        args = parse_argstring_or_throw(self.cancel, line)
+
+        if args.session is None:
+            self.ipython_display.send_error(u'You must provide a session ID (-s argument).')
+            return
+
+        if args.force:
+            if args.statement is not None:
+                self.spark_controller.cancel_statement_in_session(self.endpoint, args.session, args.statement)
+            else:
+                self.spark_controller.cancel_session_by_id(self.endpoint, args.session)
+        else:
+            self.ipython_display.send_error(u"Include the -f parameter if you understand that statements in execution and waiting for execution will be interrupted")
 
     @cell_magic
     def _do_not_call_start_session(self, line, cell="", local_ns=None):
--- a/sparkmagic/sparkmagic/kernels/wrapperkernel/usercodeparser.py
+++ b/sparkmagic/sparkmagic/kernels/wrapperkernel/usercodeparser.py
@@ -11,7 +11,8 @@ class UserCodeParser(object):
     #    some_input
     _magics_with_no_cell_body = [i.__name__ for i in [KernelMagics.info, KernelMagics.logs, KernelMagics.cleanup,
                                                       KernelMagics.delete, KernelMagics.help, KernelMagics.spark,
-                                                      KernelMagics.send_to_spark]]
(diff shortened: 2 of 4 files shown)


PedroRossi:
Hi @LinanZheng! Thank you for the contribution :tada: 

I have a question regarding this PR, I didn't understand the flow of using the `%cancel`, the user should stop the cell and then later use the magic to cancel the statement? If I got it right, I think it is better to have a callback to the cell stop button so this can trigger Livy cancel (I don't know If the kernel code exposes the stop button behaviour so it can be extended easily, but it is worth searching it, I will search it also later)

LinanZheng:
@PedroRossi 

Thank you for the kind word and you brought up a truly great question. First of all, to answer your question, I think we should have both SIGINT callback and this cancel magic. Having this magic is kind of like the delete session magic which also kick in you restart the kernel. Great to have the callback but there are many use cases that a separate magic that can cancel not just this session but other session/statements would be helpful. 

As for the stop button callback, it is actually the first thing that came to my mind and I have spent nontrivial time researching how to do that. `IPythonKernel` that SparkMagic Kernels inherit actually has a contextmanager function called `_cancel_on_sigint` that deploy a SIGINT coroutine to interrupt the main thread. Unfortunately we don't use asyncio runner so we have a shell running synchronous calls. It is totally possible to wrap our cell with a contextmanager for this custom cancel logic but I don't want to mess with this too much since having this will solve the problem albeit definitely not as easy as callback to interrupt button. Still, I would love to hear more about your research and maybe you have a better way to handle this with your broader knowledge.

PedroRossi:
@LinanZheng we can start with this magic then and create an issue about the `SIGINT` callback

About the questions, I think the session argument is not needed, every cancel should only cancel statements of their own session and the statement argument can be better if we can list the statements also (listing statements can be another magic)

@devstein @itamarst take a look at this later

LinanZheng:
Sound great! I will create an issue regarding `SIGINT` callback later today. Once @devstein and @itamarst had the chance taking a look, I will consolidate all your suggestions into new commits. Thanks!

itamarst:
1. From a UX perspective, I would personally much prefer the SIGINT-style thing where you just press the stop button. I don't think it necessitates a rewrite in asyncio; while HTTP requests from `requests` can't be interrupted, there's a while loop polling status in  `_get_statement_output`.

2. Are you sure this actually works? How can you run code in cell N+1 if cell N is still running?

itamarst:
Assuming it actually works, I'm fine with @PedroRossi 's suggestions though, feel free to proceed.

LinanZheng:
@itamarst Sorry for this late reply. Have been busy with some other stuff. What you said makes sense and I think all of us would agree that pressing the Interrupt button would be ideal. What I was talking about is that I think it is also necessary to have this magic command implemented. Also, it won't work in N+1 when you having cell N running. You need to interrupt it first "locally" from the Jupyter side so that the magic command can be executed.

LinanZheng:
@itamarst @devstein @PedroRossi I think I have all of you's approvals in various forms. I would refactor the MR to incorporate another statement list magic per @PedroRossi 's suggestions and add new unit tests for these changes I made.  If I have some extra time, I will go investigate the approach @itamarst has recommended. Thanks everyone!

itamarst:
BTW @LinanZheng , any interest in becoming a maintainer, since you've already had a PR merged? You'd need to follow the code of conduct and maintainer guidelines (tests, code review).

LinanZheng:
@itamarst I would be honored. Just curious, where can I find the code of conduct and maintainer guidelines you mentioned?

hanyucui:
@LinanZheng This is PR is so close. It would be great if you can push it through finish line.

LinanZheng:
@hanyucui Hi there, thanks for reminding me this PR. I have been busy with job transition these few months and should be able to get back to this very soon.

dylanwilder:
Hey any update here. This can be a painful experience for our users

devstein:
@LinanZheng Closing this in favor of the already merged https://github.com/jupyter-incubator/sparkmagic/pull/671

PR #660, opened 2020-07-27
Fix iteration over the sessions in /reconnectsparkmagic

qlemaire22:
Hi!

I noticed that the `/reconnectsparkmagic` endpoint was broken (with sparkmagic 0.15.0, iPython 7.16.1 and notebook 6.0.3).
I did an investigation and I found out that it was always recreating a new kernel because it could not correctly iterate over the different sessions to find to right kernel to restart. 

So here is a small fix I made to make it work.

### DIFF
--- a/sparkmagic/sparkmagic/serverextension/handlers.py
+++ b/sparkmagic/sparkmagic/serverextension/handlers.py
@@ -99,7 +99,7 @@ class ReconnectHandler(IPythonHandler):
         sessions = self.session_manager.list_sessions()
 
         kernel_id = None
-        for session in sessions:
+        for session in sessions.result():
             if session['notebook']['path'] == path:
                 session_id = session['id']
                 kernel_id = session['kernel']['id']


qlemaire22:
It seems to break the CI, I will do some further investigation to understand why.

qlemaire22:
The test is fixed. I had to create a new function `_get_list_sessions` that I mocked since I didn't find a way to correctly mock `session_manager.list_sessions()` with the new return format [source].(https://github.com/jupyter/notebook/blob/d17caf92c0817e8bc771209a1e867f0c73e34724/notebook/services/sessions/sessionmanager.py#L268)

If you have a solution please for mocking `session_manager.list_sessions(),` tell me :)

devstein:
Hi @qwertwwwe thanks for making a PR! Is this fix still relevant? Can you provide instructions for reproducing the bug without this fix in Jupyterlab?

PR #662, opened 2020-08-14
Add support for custom authentication classes

### LINKED ISSUE #659: Add pluggable authentication support
SparkMagic supports remote access to Hadoop clusters from Jupyter notebooks, but does not support Google authentication unless customers open their Virtual Private Cloud (VPC) networks to outside connections, which poses an untenable security risk. Due to this risk, customers cannot safely use SparkMagic from outside their project. We want to allow users to be able to dynamically connect to an existing Google Dataproc cluster with Livy installed from a remote Jupyter notebook with Google authentication.

In order to add Google Authentication to SparkMagic, it requires that we add a lot of changes that are google specific. So instead of hard-coding in all of these google specific changes, it would be cleaner to allow custom authentication classes. After sparkmagic can support pluggable auth, we can put our Google authentication support in a lightweight wrapper project that has Sparkmagic as a dependency. 

We have considered hardcoding in Google authentication similar to how Sparkmagic hard codes in Kerberos Authentication; however, we want this to be maintainable and adding support for pluggable auth seems to be a better way to ensure that.

alexismacaskilll:
Fixes #659 : Add support for custom authentication classes

Added a base Authenticator class that all custom authentication classes will extend. The base Authenticator is analogous to the "None" auth type Sparkmagic supported previously. The Basic and Kerberos authenticator classes extend the base Authenticator class and will replace Sparkmagic's existing implementation of Basic Access and Kerberos Authentication. All authenticators have a get_widgets() function that creates the widgets that authenticator needs in the endpoint tab of the %manage_spark widget. Instead of creating widgets for all the different auth types, AddEndpointWidget just loads all the Custom authentication classes from the configuration file and initializes them. Removed the username and password attributes of the Endpoint class and changed the auth attribute to be the Authenticator instance instead of the current implementation which uses a string ("None", "Basic_Access", "Kerberos"). Changed the auth attribute of the ReliableHttpClient to be the authenticator instance of the endpoint. Each custom authenticator implements a call function that takes a Request object as an argument, and attaches the authenticator's authentication to the given Request object. Added an initialize_auth function to utils.py which accepts a parsed arg_string from a magic command and returns the instance of the authenticator that was specified with the -t flag. Changed all the subcommands in remotesparkmagics.py to add/cleanup/delete the endpoint with the Authenticator instance returned from the initialize auth function. 

Allowing custom Authentication classes makes it easier for others to add their own custom authenticators without having to upstream a change to Sparkmagic in order to use Sparkmagic with their own authentication type. If several other people wanted to add their own authentication types, this would quickly get out of hand. Another way people to add additional authentication types to Sparkmagic would be to make their new authentication changes into a Sparkmagic fork, but this takes a lot of time and effort to maintain. 

Adding support for custom authentication classes to Sparkmagic will allow others to add their own custom authenticators by creating a lightweight wrapper project that has Sparkmagic as a dependency and that contains their custom authenticator that extends the base Authenticator class.


sparkmagic/sparkmagic/tests/test_configuration.py
```diff
@@ -6,7 +6,6 @@
 from sparkmagic.livyclientlib.exceptions import BadUserConfigurationException
 from sparkmagic.utils.constants import AUTH_BASIC, NO_AUTH
 
-
```
PedroRossi, line 9: I think that there was a lint problem since this file had no changes except for this space, if possible preserve these spaces, so we can a have a contribution later to address any linter issues

sparkmagic/sparkmagic/tests/test_livyreliablehttpclient.py
```diff
@@ -10,6 +10,7 @@
 from sparkmagic.livyclientlib.linearretrypolicy import LinearRetryPolicy
 
 
+
```
PedroRossi, line 13: Linter issue again

PedroRossi (approved):
Really nice PR here! Just added some comments regarding some strange spaces but I will pre-approve this. Hopefully we can have more Authentication classes after this PR. \o/

devstein (approved):
LGTM -- Thank you for the high-quality PR 🥇

itamarst:
@alexismacaskilll any interest in becoming one of the SparkMagic maintainers? Just need to agree to follow code of conduct, and the development process.

alexismacaskilll:
> @alexismacaskilll any interest in becoming one of the SparkMagic maintainers? Just need to agree to follow code of conduct, and the development process.

Sorry for the delay! Yes I am interested in being one of the SparkMagic maintainers.

PR #671, opened 2020-10-09
Cancel statement when KeyboardInterrupt is received

### LINKED ISSUE #629: Implement SIGINT callback to cancel Livy statements
**Is your feature request related to a problem? Please describe.**
This feature request is a follow up to the PR https://github.com/jupyter-incubator/sparkmagic/pull/628. Once the MR is merged, users will be able to cancel statements but having the notebook automatically sending cancel request when the kernel receives SIGINT would be a great addition to the repo. 

**Describe the solution you'd like**
Implement a contextmanager function or other SIGINT handler to send cancel request when the user press the interrupt button in Jupyter Notebook. Cancellation logics has been implemented in the original PR linked above

**Additional context**
Follow up enhancement to PR https://github.com/jupyter-incubator/sparkmagic/pull/628

hanyucui:
This PR was inspired by #628 and resolves #629. The PR handles
`KeyboardInterrupt` by canceling the corresponding statement. After
handlng it, the code also raises a sub-class of `KeyboardInterrupt` to
make sure subsequent cells are not executed.


sparkmagic/sparkmagic/livyclientlib/command.py
```diff
@@ -42,6 +42,19 @@ def execute(self, session):
             response = session.http_client.post_statement(session.id, data)
             statement_id = response[u'id']
             output = self._get_statement_output(session, statement_id)
+        except KeyboardInterrupt as e:
+            msg =  u'Interrupted by user'
```
hanyucui, line 46: Still need to find a good place to print this message.

hanyucui:
@itamarst @PedroRossi @devstein @LinanZheng Please take a look. I wanted to get some feedback before refining it and adding tests.

sparkmagic/sparkmagic/livyclientlib/command.py
```diff
@@ -42,6 +42,19 @@ def execute(self, session):
...
+            if statement_id >= 0:
+                response = session.http_client.cancel_statement(session.id, statement_id)
+                if u'msg' not in response or response[u'msg'] != u'canceled':
+                    msg += u', but Livy returns error'
+                else:
+                    session.wait_for_idle()
+            self._spark_events.emit_statement_execution_end_event(session.guid, session.kind, session.id,
+                                                                  self.guid, statement_id, False, e.__class__.__name__,
+                                                                  str(e))
+            # This is ugly, but raising SparkStatementException or returning does not interrupt subsequent cells.
```
devstein, line 56: This is intended. If you want `raising SparkStatementException` to interrupt, you need to set [all_errors_are_fatal](https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/utils/configuration.py#L256-L259)

sparkmagic/sparkmagic/livyclientlib/command.py
```diff
@@ -42,6 +42,19 @@ def execute(self, session):
             response = session.http_client.post_statement(session.id, data)
             statement_id = response[u'id']
             output = self._get_statement_output(session, statement_id)
+        except KeyboardInterrupt as e:
+            msg =  u'Interrupted by user'
+            if statement_id >= 0:
```
devstein, line 47: Can you check for the unexpected case first (`statement_id < 0`) and raise a more informative error message (i.e "cell interrupted before the statement was created")?

sparkmagic/sparkmagic/livyclientlib/command.py
```diff
@@ -42,6 +42,19 @@ def execute(self, session):
...
+                response = session.http_client.cancel_statement(session.id, statement_id)
+                if u'msg' not in response or response[u'msg'] != u'canceled':
```
devstein, line 49: `livy` always returns `{ "msg": "canceled" }` if it is successful. If it is not successful, `session.http_client.cancel_statement` will raise an `HttpClientException`, so I don't see a reason to check the response here. On success, we should re-raise a helpful error message to inform the user

sparkmagic/sparkmagic/livyclientlib/command.py
devstein: What are the options you've considered?

devstein (changes requested):
@hanyucui Thanks for making a PR! This seems like an intuitive and simple solution. I left some comments

sparkmagic/sparkmagic/livyclientlib/command.py
hanyucui: I only tried returning it, but remaining cells will continue to execute, which I also want to avoid. If I raised an exception, the backtrace is too long and not not helpful at all. Any suggestion?

sparkmagic/sparkmagic/livyclientlib/command.py
hanyucui: Sure, what exception should be raised here? I have yet to fully understand the uses of different exceptions.

sparkmagic/sparkmagic/livyclientlib/command.py
hanyucui: I don't necessarily need to raise `SparkStatementException`. My goal is to interrupt the notebook's execution and print a short message to confirm that. I have yet to find a way to achieve this. Any suggestion would be welcome. I will also dig a bit more.

hanyucui:
> @hanyucui Thanks for making a PR! This seems like an intuitive and simple solution. I left some comments

Thanks for your feedback, @devstein ! I asked some questions, let me know what you think. I am still not very familiar with Sparkamgic's code base and dig a big more.

LinanZheng:
Hey, @hanyucui. Thanks for this MR. My new job is quite demanding so I haven't get the chance to work on 628. I will review your code this week and maybe provide some pointers if I could.

sparkmagic/sparkmagic/livyclientlib/command.py
PedroRossi, line 46: Add this message and the other to the constants file to avoid plain text here.

sparkmagic/sparkmagic/livyclientlib/command.py
PedroRossi: Why not create a new Exception to represent this Interruption and make it always fatal?

hanyucui: In the case of a fatal exception, the backtrace is pretty long and unhelpful. Is there a way to work around this? Since the user is expecting the notebook to be interrupted, a short message to confirm it is preferred over printing the backtrace.

sparkmagic/sparkmagic/livyclientlib/command.py
hanyucui: Now that I think about it, this doesn't seem necessary. When the execution reaches here, the user wants to interrupt the notebook and cancel the statement on Livy. The `statement_id` being invalid means the statement submission was unsuccessful and there is no need to cancel it -- this is what the user wants in most cases and she doesn't need to be informed.

sparkmagic/sparkmagic/livyclientlib/exceptions.py
```diff
@@ -54,6 +55,27 @@ class SqlContextNotFoundException(LivyClientLibException):
 class SparkStatementException(LivyClientLibException):
     """Exception that is thrown when an error occurs while parsing or executing Spark statements."""
 
+
+class SparkStatementCancelledException(KeyboardInterrupt):
```
hanyucui, line 59: I had to inherit from KeyboardInterrupt to stop interrupt the notebook.

hanyucui:
@devstein can you take a another look? Thanks!

sparkmagic/sparkmagic/livyclientlib/command.py
```diff
@@ -42,6 +44,14 @@ def execute(self, session):
             response = session.http_client.post_statement(session.id, data)
             statement_id = response[u'id']
             output = self._get_statement_output(session, statement_id)
+        except KeyboardInterrupt as e:
+            if statement_id >= 0:
+                response = session.http_client.cancel_statement(session.id, statement_id)
+                session.wait_for_idle()
```
LinanZheng, line 50: I noticed here that session will wait for idle here for the statement cancellation. Based on my previous experience with Livy, cancel statement can fail and will always fail for non-spark code due to Livy's implementation (they might have done works after my experiments). You might need to consider that into this MR

hanyucui: Thanks for the suggestion. When the execution reaches here, all the user wants is to interrupt the execution. How about I just try to catch everything here? I can also pass a shorter timeout to `wait_for_idle()`

LinanZheng: Hey Hanyu. Yes I understand that interruption is all we want to achieve but as you mentioned, we might get timeout for idle waiting and what's worse is that Livy Session will become unusable after a "unsupported" cancel call. What I would suggest is that we catch that timeout exception and print helpful exception msg telling user like the session might be dead due to the cancellations and restart might be required.

devstein (approved):
LGTM -- thanks for contributing!

itamarst:
@hanyucui interested in getting commit access? Just have to agree to follow the dev process and code of coduct.

hanyucui:
> @hanyucui interested in getting commit access? Just have to agree to follow the dev process and code of coduct.

Thanks, @itamarst . I'd love to. Although, I perhaps won't be able to do a lot of code reviews.

itamarst:
@devstein don't have a lot of time for this, these days, but skimmed this very very quickly, seems fine at superficial glance. Someone should try this in a notebook before it's merged...

This is the kinda place where end-to-end tests would be really nice, if anyone feels like it.

devstein:
@itamarst Appreciate the input! I manually tested as part of the review and it appeared to work as expected.

I agreed end-to-end tests are very very needed.

hanyucui:
Yeah, an end-to-end test would be great. I manually tested it on Python 2 & 3 with Livy+Spark on EMR.

PR #690, opened 2020-12-10
Fix the bug that reconnectsparkmagic cause 'Error Starting Kernel' in jupyterlab

qwertwwwe:
### Description
Fix the bug that reconnectsparkmagic cause 'Error Starting Kernel' in jupyterlab, add the parameter 'name' in create new session.
The function self._get_kernel_manager_new_session create a new session without passing the 'name' parameter, it creates a session with its name is null, the response for api '/api/sessions' in jupyter notebook contains an item with the name is null, this cause the jupyterlab failed to load the session list. When we refresh the jupyterlab web page, jupyterlab pop a 'Error Starting Kernel' error.
Test in jupyterlab 2.1.3, this pop the error 'Error Starting Kernel'. In jupyterlab 2.2.9, it doesn't pop this error, but failed to load the running sessions in 'Running Terminals and Kernels' subpage.

### Checklist

### DIFF
--- a/sparkmagic/sparkmagic/serverextension/handlers.py
+++ b/sparkmagic/sparkmagic/serverextension/handlers.py
@@ -1,4 +1,5 @@
 import json
+import os
 from notebook.utils import url_path_join
 from notebook.base.handlers import IPythonHandler
 from tornado import web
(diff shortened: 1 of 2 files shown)


devstein:
Hi @qwertwwwe thanks for making a PR! Can you provide instructions for reproducing the bug without this fix in Jupyterlab?

PR #692, opened 2020-12-14
Adding language info and file extension for the spark kernel

### LINKED ISSUE #691: Support LSP by completing the kernel metadata
As reported in https://github.com/krassowski/jupyterlab-lsp/issues/407, [jupyterlab-lsp](https://github.com/krassowski/jupyterlab) (an extension that integrates fancy IDE features with JupyterLab) does not work with spark magic unless manual modifications are made to the kernel, this is adding `file_extension` in `language_info` and changing `language` in [sparkkernel.py](https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/kernels/sparkkernel/sparkkernel.py) so that it looks closer to:

```python
class SparkKernel(SparkKernelBase):
    def __init__(self, **kwargs):
        implementation = 'Spark'
        implementation_version = '1.0'
        language = 'scala'    # or LANG_SCALA ?
        language_version = '0.1'
        language_info = {
            'name': 'scala',
            'mimetype': 'text/x-scala',
            'codemirror_mode': 'text/x-scala',
            'file_extension': '.sc',
            'pygments_lexer': 'scala'
        }

        session_language = LANG_SCALA


```

This is required for the LSP integration using jupyterlab-lsp. We received a [PR instructing user to manually patch sparkmagic](  https://github.com/krassowski/jupyterlab-lsp/pull/415), but we think that it would be ideally addressed here, if it would not be a problem. I do not understand enough about sparkmagic to know why language is set to `'no-op'`, nor if it can be changed to declare the actual language in the standard attribute rather than in the non-standard `session_language` - I would appreciate some feedback here!

skakker:
### Description
The language and file_extension are missing from the spark magic kernel, because of which it has trouble integrating with other features like jupyterlab-lsp. Hence updating the same. 

Closes #691 

### Checklist
- [ ] Added or modified unit tests to reflect my changes
- [ ] Manually tested with a notebook

### DIFF
--- a/sparkmagic/sparkmagic/kernels/pysparkkernel/pysparkkernel.py
+++ b/sparkmagic/sparkmagic/kernels/pysparkkernel/pysparkkernel.py
@@ -14,6 +14,7 @@ class PySparkKernel(SparkKernelBase):
             'name': 'pyspark',
             'mimetype': 'text/x-python',
             'codemirror_mode': {'name': 'python', 'version': 3},
+            'file_extension': '.py'
             'pygments_lexer': 'python3'
         }
 
--- a/sparkmagic/sparkmagic/kernels/pysparkkernel/pysparkkernel.py
+++ b/sparkmagic/sparkmagic/kernels/pysparkkernel/pysparkkernel.py
@@ -13,15 +13,20 @@ class PySparkKernel(SparkKernelBase):
         language_info = {
             'name': 'pyspark',
             'mimetype': 'text/x-python',
-            'codemirror_mode': {'name': 'python', 'version': 3},
-            'file_extension': '.py'
+            'codemirror_mode': {
+                'name': 'python',
+                'version': 3
+            },
+            'file_extension': '.py',
             'pygments_lexer': 'python3'
         }
 
         session_language = LANG_PYTHON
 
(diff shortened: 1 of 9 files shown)


devstein (changes requested):
Thanks for contributing! Can you also make the change for the [PySpark kernel](https://github.com/jupyter-incubator/sparkmagic/tree/master/sparkmagic/sparkmagic/kernels/pysparkkernel) and provide instructions for how to do a manual test?

skakker:
Added for Pyspark and R kernel. 
Thanks :)

sparkmagic/sparkmagic/kernels/pysparkkernel/pysparkkernel.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, **kwargs):
             'name': 'pyspark',
             'mimetype': 'text/x-python',
             'codemirror_mode': {'name': 'python', 'version': 3},
+            'file_extension': '.py'
```
devstein, line 17: Add a `,`

devstein (changes requested):
One quick fix

devstein (approved):

PR #184, opened 2016-03-09
Enormous refactor of Livyclientlib

msftristew:
Huge refactoring to Livyclientlib to make some API's more sensible and the code simpler.  

Closes #185

* New `Command` class. This is a class which represents user code that should be executed against a session. Each `Command` has a GUID for use in triggering events.

* Remove `LivyClient` class (functionality subsumed by `Command`, `SqlQuery`, and `LivySession`)

* Rename classes to incorporate new changes (e.g. `ClientManager` -> `SessionManager`)

* Remove all references to "serialization" (we can add this back in when we decide on an HA story)

* Change API of `LivyReliableHttpClient`, which now has a few Livy-specific methods rather than generic `get` and `post` methods

* Change tests to accomodate these changes

* Change types of session ID string -> int

* Remove `LivySessionState` class (subsumed by `LivySession`)

* Change names of configurations to better reflect their meaning. Now there is a configuration `wait_for_idle_timeout_seconds` (which reflects the amount of time the library will wait to go to idle before executing a statement; defaults to 15 seconds) and `livy_session_startup_timeout_seconds` (which reflects the amount of time that the library will wait for the Livy session to start up).  

* Remove recursion in `wait_for_idle` and in `ReliableHttpClient` implementation

PR #716, opened 2021-06-04
Added one internal magic to enable retry of session creation.

edwardps:
**Description**
Fix two issues:
1. the self.session_started[1] is set as true too early for exceptional case.
2. the session is added too early[2] before session is actually started.

Enhanced the magic kernel by adding one internal magic to allow retry session
creation. Currently, the session creation error is treated as fatal error which
make it impossible[3] to retry after fixing the endpoint info.

[1] https://github.com/jupyter-incubator/sparkmagic/blob/bfabbb39a0249197c2c05c8efe681710fff9151b/sparkmagic/sparkmagic/kernels/kernelmagics.py#L357
[2] https://github.com/jupyter-incubator/sparkmagic/blob/bfabbb39a0249197c2c05c8efe681710fff9151b/sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py#L90
[3] https://github.com/jupyter-incubator/sparkmagic/blob/bfabbb39a0249197c2c05c8efe681710fff9151b/sparkmagic/sparkmagic/kernels/kernelmagics.py#L350

Please see the movitivation for our use cases.
**Motivation**
To enable user to choose an Spark cluster and connect to it at runtime(means the
sparkmagic configuration is not generated in advance), we need to configure the endpoint
after kernel starting and enable retry session creation after fixing the incorrect endpoint.

We discussed this with Alejandro(aggFTW@) and he suggested to leverage the following _do_not_call_*
methods to configure the endpoint dynamically and create session after specifying configuration.
1.    _do_not_call_delete_session
2.    _do_not_call_change_endpoint
3.    _do_not_call_start_session

To conditionally allow the session creation retry, the changes introduced a flag
to allow retry on previous fatal of session creation. By default, the behavior
is kept unchanged and do not allow retry on fatal.

**Testing Done**
1. Added unit test.
2. Also ran the code in a notebook to verify.

**Backwards Compatibility Criteria (if any)**
The enhancement is activated conditionally and no compatibility issue is expected.

### Checklist
- [-] If adding a feature, there is an example notebook and/or documentation in the `README.md` file (N/A as no user visible feature added. The internal magic is added for kernel integration purpose)

### DIFF
--- a/sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
+++ b/sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
@@ -87,8 +87,8 @@ class SparkController(object):
             return
         http_client = self._http_client(endpoint)
         session = self._livy_session(http_client, properties, self.ipython_display)
-        self.session_manager.add_session(name, session)
         session.start()
+        self.session_manager.add_session(name, session)
 
     def get_session_id_for_client(self, name):
         return self.session_manager.get_session_id_for_client(name)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## NEXT RELEASE
+### Features
+Added one internal magic to enable retry of session creation. 
 
 ## 0.18.0
 
--- a/sparkmagic/sparkmagic/kernels/kernelmagics.py
+++ b/sparkmagic/sparkmagic/kernels/kernelmagics.py
@@ -57,6 +57,7 @@ class KernelMagics(SparkMagicBase):
         self.language = u""
         self.endpoint = None
         self.fatal_error = False
+        self.allow_retry_fatal = False
         self.fatal_error_message = u""
         if spark_events is None:
             spark_events = SparkEvents()
@@ -347,17 +348,19 @@ class KernelMagics(SparkMagicBase):
         # No need to add the handle_expected_exceptions decorator to this since we manually catch all
         # exceptions when starting the session.
 
-        if self.fatal_error:
+        if self.fatal_error and not self.allow_retry_fatal:
             self.ipython_display.send_error(self.fatal_error_message)
             return False
 
         if not self.session_started:
             skip = False
             properties = conf.get_session_properties(self.language)
-            self.session_started = True
 
             try:
                 self.spark_controller.add_session(self.session_name, self.endpoint, skip, properties)
+                self.session_started = True
+                self.fatal_error = False
+                self.fatal_error_message = u""
             except Exception as e:
                 self.fatal_error = True
                 self.fatal_error_message = conf.fatal_error_suggestion().format(e)
@@ -370,6 +373,11 @@ class KernelMagics(SparkMagicBase):
                 return False
 
         return self.session_started
+
+    @cell_magic
+    def _do_not_call_allow_retry_fatal(self, line, cell="", local_ns=None):
+        # enable the flag to retry session creation
+        self.allow_retry_fatal = True
 
     @cell_magic
     @handle_expected_exceptions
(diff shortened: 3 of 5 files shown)


devstein (commented):
@edwardps Thank you for the well-written PR. Is the meant to address https://github.com/jupyter-incubator/sparkmagic/issues/709?

edwardps:
Hi Devin,

> @edwardps Thank you for the well-written PR 🥇 Is this also meant to address #709?

No exactly, our case is kind of overlapping with original issue reported by 709 but not exactly the same. 
Issue 709 can be addressed if it's fine to have python code communicating with kernel to invoke the _do_not_call_* methods like below.

```
%%local
# set endpoint
change_endpoint_internal_magic = get_ipython().find_line_magic('_do_not_call_change_endpoint')
change_endpoint_internal_magic('-s http://host:8998 -t Basic_Access -u billy -p welcome123')

# start session
start_session_internal_magic = get_ipython().find_cell_magic('_do_not_call_start_session')
start_session_internal_magic()
```

sparkmagic/sparkmagic/livyclientlib/sparkcontroller.py
```diff
@@ -87,8 +87,8 @@ def add_session(self, name, endpoint, skip_if_exists, properties):
             return
         http_client = self._http_client(endpoint)
         session = self._livy_session(http_client, properties, self.ipython_display)
-        self.session_manager.add_session(name, session)
         session.start()
+        self.session_manager.add_session(name, session)
```
devstein, line 91: @edwardps Can you think of any unintended side-effects of only adding the session after it's started? I agree this change makes sense, but it's been this way for many years and I don't know if there are historical reasons for the ordering

edwardps: @devstein This is a fair callout. If there is timeout issue[1], we might have session leaking. So can clean up the session for edge case caused by e.g. timeout.
Any suggestion? 

```
        try:
            session.start()
        except:
            if session.is_posted(): session.delete() # is_posted() is decided by self.status = constants.NOT_STARTED_SESSION_STATUS
            raise
        else:
            self.session_manager.add_session(name, session)
```
 
[1] - https://github.com/jupyter-incubator/sparkmagic/blob/master/sparkmagic/sparkmagic/livyclientlib/livysession.py#L151

edwardps: Hi Devin,

Made a code fix for the session leaking issue. Please let me know if you have any suggestion or comments. 
https://github.com/edwardps/sparkmagic/commit/b6c89eb03c9440acb6aa9fc9e4fea258c9da5f7b

Will update the PR after confirming the solution. 

Thanks,
Ed

devstein: Hey Ed, thanks for the quick turnaround. The solution looks good to me. Can you add a comment to explain the logic (avoid leaking sessions)? 

Also, cc @aggFTW as he was the last to [edit this logic](https://github.com/jupyter-incubator/sparkmagic/pull/162/files#diff-5b8ca0c4c8dae6a26bd649addcc20b8ed583eb3106629668f9dfb3ac3066877dR236-R237) and [refactored it](https://github.com/jupyter-incubator/sparkmagic/pull/184/files#diff-c9a3377ce7132a4f46b4f99611416ab9cc3e1494dd3d540c2f088d28d80807b6R71). @aggFTW Do you see any issues with changing this logic?

edwardps: Sure thing. I will add more details in the comment when update this PR.

edwardps: Added info about the session leaking edge case handling in the PR description and git submit.

devstein (approved):

Issue #727, opened 2021-08-26
Support IPyKernel >= 6.0.0


juliusvonkohout:
**Describe the bug**
Ipykernel is downgraded from 6.x (jupyterlab 3.1) to 5.x by sparkmagic

#723 

**To Reproduce**
```
pip install jupyterlab==3.1.*
pip install sparkmagic
...

Installing collected packages: ipykernel, qtpy, qtconsole, numpy, jupyter-console, tenacity, pandas, nose, mock, jupyter, pykerberos, plotly, hdijupyterutils, requests-kerberos, docopt, autovizwidget, sparkmagic, hdfs
 Attempting uninstall: ipykernel
Found existing installation: ipykernel 6.2.0
Uninstalling ipykernel-6.2.0:
Successfully uninstalled ipykernel-6.2.0
Successfully installed autovizwidget-0.19.0 docopt-0.6.2 hdfs-2.6.0 hdijupyterutils-0.19.0 ipykernel-5.5.5 jupyter-1.0.0 jupyter-console-6.4.0 mock-4.0.3 nose-1.3.7 numpy-1.19.5 pandas-1.3.2 plotly-5.2.2 pykerberos-1.2.1 qtconsole-5.1.1 qtpy-1.10.0 requests-kerberos-0.12.0 sparkmagic-0.19.1 tenacity-8.0.1
`````

**Expected behavior**
Sparkmagic should support ipykernel 6.x

**Versions:**
 - SparkMagic 1.19.1

PR #737, opened 2021-09-16
Support ipykernel >= 6

devstein:
Closes https://github.com/jupyter-incubator/sparkmagic/issues/727

### Description

- Support `ipykernel>=6` [async def do_execute](https://github.com/ipython/ipykernel/blob/master/ipykernel/ipkernel.py#L294-L403)
- Update `ipykernel` 4 and 5 tests 
- Add test for `ipykernel` 6
- Removes `ipykernel<6` constraints
- Ran `black` formatter on the changed files

### TODOs
- [x] Figure out what other `SparkKernelBase` methods need to be `async`

### Checklist
- [ ] Manually tested with a notebook


juliusvonkohout:
Thanks for your efforts @devstein 

What about the reviewers ? 
@Carreau
@aggFTW
@itamarst
@juhoautio
@PedroRossi

sergiimk:
Hi @devstein I finally had time to give this another go (same setup as I described before).

Initially it looked like kernel has started successfully, but I saw the following error in the log:
```
Exception in callback <TaskWakeupMethWrapper object at 0x7f0212502910>(<Future finis...511"\r\n\r\n'>)
handle: <Handle <TaskWakeupMethWrapper object at 0x7f0212502910>(<Future finis...511"\r\n\r\n'>)>
  File "/opt/conda/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
RuntimeError: Cannot enter into task <Task pending name='Task-1' coro=<HTTP1ServerConnection._server_request_loop() running at /opt/conda/lib/python3.9/site-packages/tornado/http1connection.py:823> wait_for=<Future finished result=b'GET /api/co...7511"\r\n\r\n'> cb=[IOLoop.add_future.<locals>.<lambda>() at /opt/conda/lib/python3.9/site-packages/tornado/ioloop.py:688]> while another task <Task pending name='Task-2' coro=<KernelManager._async_start_kernel() running at /opt/conda/lib/python3.9/site-packages/jupyter_client/manager.py:337>> is being executed.
```
A pretty obscure stack trace...

Attempting to use the notebook then results in errors like:
```
UsageError: Cell magic `%%spark` not found.
```
Which indicates that the above error prevented `sparkmagic` to fully initialize.

I'll take a look at the PR and share ideas if I'll have any.

sergiimk:
You can build this docker image to reproduce the problem yourself:
```dockerfile
FROM jupyter/base-notebook:2021-10-18

USER root
RUN conda install requests-kerberos -y
RUN apt-get update && \
    apt-get -y install git && \
    rm -rf /var/lib/apt/lists/*

USER $NB_USER
RUN pip install --upgrade pip && \
    pip install --upgrade --ignore-installed setuptools && \
    pip install "git+https://github.com/jupyter-incubator/sparkmagic.git@devstein/ipykernel-6#egg=sparkmagic&subdirectory=sparkmagic"

RUN jupyter nbextension enable --py --sys-prefix widgetsnbextension
RUN jupyter-kernelspec install --user $(pip show sparkmagic | grep Location | cut -d" " -f2)/sparkmagic/kernels/pysparkkernel
RUN jupyter serverextension enable --py sparkmagic
```

sergiimk:
Sorry, I just realized that the above is an upstream Jupyter issue: https://github.com/jupyter/notebook/issues/6164
The error shows up even with vanilla `jupyter/base-notebook:2021-10-28` image in a kernel without `sparkmagic`.

This means the `UsageError: Cell magic `%%spark` not found.` error is unrelated.

sparkmagic/sparkmagic/kernels/wrapperkernel/sparkkernelbase.py
```diff
@@ -43,108 +49,168 @@ def __init__(self, implementation, implementation_version, language, language_ve
         # Disable warnings for test env in HDI
         requests.packages.urllib3.disable_warnings()
 
-        if not kwargs.get("testing", False):
-            self._load_magics_extension()
-            self._change_language()
-            if conf.use_auto_viz():
-                self._register_auto_viz()
```
sergiimk, line 50: I don't see where all these functions are now being called from, and I'm guessing this is the reason magics don't get registered upon kernel startup.

Doesn't look like the `IPythonKernel` provides any `async` startup methods to call these from, so I see why you had to remove them...

I wonder if lazy initialization of sorts upon first call to `do_execute` would work here? Perhaps someone with more experience with IPython will suggest a better option.

devstein: @sergiimk I tried

1. Lazily initializing the magics on the first call to `do_execute`
2. Running them synchronously within the `init` via `asyncio`

but neither approach works 😞 .  Let me know if you have any other thoughts.

sergiimk: Sorry for late reply @devstein  - I was struggling to find time to take a closer look at the code.

Running load magics synchronously probably won't work:
- IO loops are non re-entrant, so you cannot block on a coroutine if the loop is already running
- If you create a new loop for those tasks it will have its own separate list of IO descriptors. So if it happens to await for anything that is managed by the original loop - it will block forever.

I tried lazy loading and ended up writing code very similar to your first attempt:
```python
    async def do_execute(...):
        if not self._lazy_initialized:
            self._lazy_initialized = True
            await self._load_magics_extension()
            await self._change_language()
            if conf.use_auto_viz():
                await self._register_auto_viz()
```
And it **mostly** worked in my notebooks - magics, auto viz was running as expected, but for some reason the images (loaded via markdown and via HTML) would not display. Only after killing the kernel a few times and switching back and forth to vanilla Python3 kernel the PySpark kernel would suddenly start displaying images. I suspect though that this is not related to this PR but a side effect of jupyter/notebook#6164.

Can you tell me what did not work for you when you tried lazy initializing the magics?

juliusvonkohout:
@devstein what is keeping this from being merged?

devstein:
Hi @juliusvonkohout the main blocker is I have not been able to get this PR to work with https://github.com/nteract/papermill (cc @MSeal). I have not had time to debug to understand the root cause, so I worry about the implications for other common notebook libraries that I'm unaware of. Any help would be appreciated!

juliusvonkohout:
> Hi @juliusvonkohout the main blocker is I have not been able to get this PR to work with https://github.com/nteract/papermill (cc @MSeal for help 🙏 ). I have not had time to debug to understand the root cause, so I worry about the implications for other common notebook libraries that I'm unaware of. Any help would be appreciated!

Sadly i have zero experience with writing kernels or papermill and I do not use the old and deprecated "Notebook" interface anyway, just jupyterlab 3.2. I just fixed some small things in sparkmagic. If if works properly in a recent jupyterlab i would merge it and not care about old deprecated stuff. I am completely occupied until May by my employer and private stuff, so I can only build it myself and test whether it works in Jupyterlab until then.

itamarst:
@Carreau might you have a few minutes to glance at this and see if it's using IPython APIs appropriately?

Carreau:
I had a quick looks but didn't see anything obviously wrong. I'll try to to dig deeper but these days I barely find the time to follow all that is happening on ipykernel and have barely enough bandwidth for IPython itself.

devstein:
@sergiimk @Qiuzhuang  @Carreau  @itamarst @juliusvonkohout  Thanks for the support and patience. I was finally able to take a day to tackle this. 

The latest implementation takes a page from [jupyter/notebook](https://github.com/jupyter/notebook/pull/6221/files) and uses [nest_asyncio](https://github.com/erdewit/nest_asyncio) and a custom `run_sync` function wrapper for `loop.run_until_complete` to avoid changing every kernel method to `async`. 

The tests are passing and I have seen no issues when doing integration testing manual locally. Please let me know if any of you have concerns with this approach; otherwise, I'll merge this in by end of week (May 1) and release.

PR #751, opened 2022-03-10
Add configurable progress and session info indicators

amitschang:
### Description
Configuration options are added to allow override of the spark statement progress indicator and for the html table of
session info upon startup. Both come with defaults that implement the same function as was done prior to this change, so
there would be no impact in case options are not set.

In my use-case, I wanted to create a indicator that sits at the top of the page and includes a link to the UI, and thus can be seen no matter where in the page your are focused. I think there are many other potential uses, such as users wanting to do simpler things like setting the width of the widget, or choosing a different widget for progress, etc.

*Testing*: Manually tested with notebook both without the new options and confirming the existing behavior and with overridden classes to customize both progress and startup display. Existing unit tests pass

*Note*: my editor is set to be opinionated about trailing white-space, so the PR includes some white-space fixes. I can certainly prune those out if it is not desired to include unrelated cleanup. Otherwise, the diff can be viewed with white-space hidden.

### Checklist
- [x] If adding a feature, there is an example notebook and/or documentation in the `README.md` file

### DIFF
--- a/sparkmagic/sparkmagic/utils/configuration.py
+++ b/sparkmagic/sparkmagic/utils/configuration.py
@@ -96,7 +96,12 @@ def session_configs():
 
 @_with_override
 def kernel_python_credentials():
-    return {u'username': u'', u'base64_password': u'', u'url': u'http://localhost:8998', u'auth': NO_AUTH}
+    return {
+        "username": "",
+        "base64_password": "",
+        "url": "http://localhost:8998",
+        "auth": NO_AUTH,
+    }
 
 
 def base64_kernel_python_credentials():
@@ -299,13 +304,16 @@ def cleanup_all_sessions_on_exit():
 def kerberos_auth_configuration():
     return {"mutual_authentication": REQUIRED}
 
+
 @_with_override
 def progress_indicator_class():
-    return 'sparkmagic.utils.progress.HorizontalFloatProgressWidgetIndicator'
+    return "sparkmagic.utils.progress.HorizontalFloatProgressWidgetIndicator"
+
 
 @_with_override
 def startup_info_display_class():
-    return 'sparkmagic.utils.startupinfo.HTMLTableStartupInfoDisplay'
+    return "sparkmagic.utils.startupinfo.HTMLTableStartupInfoDisplay"
+
 
 def _credentials_override(f):
     """Provides special handling for credentials. It still calls _override().
--- a/sparkmagic/sparkmagic/utils/progress.py
+++ b/sparkmagic/sparkmagic/utils/progress.py
@@ -1,5 +1,6 @@
 from ipywidgets.widgets import FloatProgress, Layout
 
+
 class ProgressIndicator:
     def __init__(self, session, statement_id):
         pass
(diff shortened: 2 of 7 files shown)


devstein (commented):
Thanks for contributing and the helpful PR description @amitschang! I'll try to review this week.

sparkmagic/sparkmagic/utils/progress.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, session, statement_id):
+        pass
+
+    def update(self, value):
+        pass
+
+    def close(self):
+        pass
+
+class defaultProgressIndicator(ProgressIndicator):
```
devstein, line 13: Two minor changes:

1. Please change to `TitleCase`
```suggestion
class DefaultProgressIndicator(ProgressIndicator):
```

2. Give it a more meaningful name, one idea would be

```suggestion
class HorizontalProgressBarIndicator(ProgressIndicator):
```

sparkmagic/sparkmagic/utils/startupinfo.py
```diff
@@ -0,0 +1,20 @@
...
+        self.sessions_info = sessions_info
+        self.current_session_id = current_session_id
+
+    def write_msg(self, msg):
+        pass
+
+    def display_info(self):
+        pass
+
+class defaultStartupInfoDisplay(StartupInfoDisplay):
```
devstein, line 15: Same naming comments 
```suggestion
class BasicStartupInfoDisplay(StartupInfoDisplay):
```

sparkmagic/sparkmagic/utils/configuration.py
```diff
@@ -271,6 +271,13 @@ def kerberos_auth_configuration():
         "mutual_authentication": REQUIRED
     }
 
+@_with_override
+def progress_indicator_class():
+    return 'sparkmagic.utils.progress.defaultProgressIndicator'
+
+@_with_override
+def startup_info_display_class():
+    return 'sparkmagic.utils.startupinfo.defaultStartupInfoDisplay'
```
devstein, line 280: Reminder to change names here as well

devstein, line 280: Can you add unit tests that will tell us if these paths point to a class that doesn't exist? This will be helpful for future refactors

sparkmagic/sparkmagic/utils/startupinfo.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, ipython_display, sessions_info, current_session_id):
+        self.ipython_display = ipython_display
+        self.sessions_info = sessions_info
+        self.current_session_id = current_session_id
+
+    def write_msg(self, msg):
+        pass
+
+    def display_info(self):
+        pass
```
devstein, line 13: I found it confusing that `defaultProgressIndicator` calls `display` in its `init` function, but the `defaultStartupInfoDisplay` has `display_info` method that is called after initialization.

Can we consolidate on one pattern? What do you think?

devstein (changes requested):
Overall looks good. Just need to fix some issue around naming and class design

devstein:
@amitschang Merge conflicts from #758 an easy way to fix is

```shell
git remote add upstream git://github.com/jupyter-incubator/sparkmagic.git
git fetch upstream
# choose local changes over remotes
git merge upstream/master -X ours
# run black on all files
black .
```

amitschang:
Thanks @devstein! I will carve out some time to work on requested changes probably early next week, cheers :)

amitschang:
@devstein, I've addressed all your comments I believe. Please let me know what you think. I've also merged in the lastest upstream and ran black. There are still a small number of whitespace fixes left from my changes, for non-py files - I hope they are OK left in. The tests were added to the configuration tests since they have to to with default configuration and check whether the defaults are subclasses of the appropriate indicator classes (even though otherwise there is no attempt at enforcing this - which I think is probably OK).

amitschang:
a bit of a bump on this. Still something you think could be included?

PR #775, opened 2022-09-26
Change import paths for dataerror

GaspardBT:
### Description
Fixes import paths after pandas latest release (1.5.0)
DataError is now in `pandas.errors` change [here](https://github.com/pandas-dev/pandas/pull/47047/files)
### Checklist
- [x] Formatted my code with [`black`](https://black.readthedocs.io/en/stable/index.html)
- [ ] Added or modified unit tests to reflect my changes

### DIFF
--- a/autovizwidget/autovizwidget/plotlygraphs/graphbase.py
+++ b/autovizwidget/autovizwidget/plotlygraphs/graphbase.py
@@ -4,7 +4,13 @@
 from plotly.graph_objs import Figure, Layout
 from plotly.offline import iplot
 
-from pandas.errors import DataError
+try:
+    from pandas.core.base import DataError
+except ImportError:
+    try:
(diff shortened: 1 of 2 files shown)


lucasdurand:
This change would require the user to be on `pandas>=1.5.0` , but that isn’t specified in the package requirements, so this change should also:
1. Pin `pandas>=1.5.0` in the setup.py (which I’m thinking we don’t want to do)
2. Instead, extend the existing try/except block to handle the new `DataError` location, while still supporting the existing fallbacks

GaspardBT:
Thanks for the review 🙇 
I make the modification for the second option, is it right?

devstein (approved):
@GaspardBT Thank you for contributing! LGTM 👍

devstein:
and @lucasdurand thank you for reviewing!

PR #925, opened 2024-07-31
fix: remove duplicate fields

jeffeth-donaldson:
### Description
This fixes an issue where the fields are duplicated on newer versions of jupyter.

### Checklist
- [X] Wrote a description of my changes above 
- [X] Formatted my code with [`black`](https://black.readthedocs.io/en/stable/index.html)
- [X] Added a bullet point for my changes to the top of the `CHANGELOG.md` file
- [X] Added or modified unit tests to reflect my changes
- [X] Manually tested with a notebook
- [ ] ~If adding a feature, there is an example notebook and/or documentation in the `README.md` file~

### DIFF
--- a/sparkmagic/sparkmagic/controllerwidget/abstractmenuwidget.py
+++ b/sparkmagic/sparkmagic/controllerwidget/abstractmenuwidget.py
@@ -35,8 +35,8 @@ class AbstractMenuWidget(Box):
             self._repr_html_()
 
     def _repr_html_(self):
-        for child in self.children:
-            self.ipython_display.display(child)
+        # for child in self.children:
+        #     self.ipython_display.display(child)
         return ""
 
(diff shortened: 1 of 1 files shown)


jeffeth-donaldson:
Apologies, @devstein, I'm not terribly familiar with IPywidgets.

I believe the issue is that `_repr_html_` is supposed to return an html string whereas here `display` was called which actually draws the widget to the screen.

I'm guessing in newer versions of Jupyter, the `display` method gets called recursively on the children already which is why the duplication occurs. I can see if returning all the children's html as a string works on older versions of Jupyter, however, I no longer require sparkmagic so I'm not sure how quickly I'll get around to it. I also do not know what version of Jupyter introduced this issue, so I'd have to figure that out for testing.

PR #959, opened 2025-05-27
Fix support for IPython 8 and 9

### LINKED ISSUE #957: [BUG] IPython.core.display import issue
**Describe the bug**
I'm setting up a clean conda environment with the following dependencies:

```
[tool.poetry.dependencies]
python = "^3.11"
requests = "^2.32.3"
pyspark = "^3.5.5"
sparkmagic = "^0.22.0"
ipywidgets = "^8.1.6"
```

When I run jupyter notebook and create a notebook with pyspark as a kernel, if I execute a cell with "2+2", it errors saying that it can't import display from IPython.core.display. I believe display has moved to IPython.display a while back. Seems to be an issue with an incompatible version of ipython? The installed version of ipython is 9.2.0.

**To Reproduce**
1. Set up an env with the dependencies mentioned above. 
2. Run jupyter notebook and create a notebook for pyspark. 
3. Execute any command in a cell.

**Expected behavior**
Cell results should show.

**Versions:**
As above.

ddrinka:
### Description

A public API for `IPython.display` was introduced in `ipython==4.1.2`.  Access to the tool via the internal API was deprecated in 7.14, warned against through 8.*, and removed in 9.0.

With this change, hdijupyterutils targets the public API which enables support for the most recent versions of IPython, but requires a minor version change for the minimum required version.

Closes #957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Next Release
 
+### Bug Fixes
+
+- Add support for `ipython>=9.0.0`
+- Remove support for `ipython<4.1.2`
+
 ## 0.22.0
 
 ### Updates
--- a/hdijupyterutils/hdijupyterutils/ipythondisplay.py
+++ b/hdijupyterutils/hdijupyterutils/ipythondisplay.py
@@ -1,4 +1,4 @@
-from IPython.core.display import display, HTML
+from IPython.display import display, HTML
 from IPython import get_ipython
 import sys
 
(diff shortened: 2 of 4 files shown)


devstein (commented):
Thank you for fixing @ddrinka! Do you know what version of `ipython` introduced `IPython.display` over `IPython.core`? I'm wondering if we need to bump the min `ipython` requirement

ddrinka:
I've adjusted the required IPython version and added a Changelog entry since this now introduces an external-facing change.

Issue #963, opened 2025-10-12
[BUG] Sparkmagic hanging indefinitely when connecting to livy from vscode-jupyter


001ben:
**Describe the bug**
When trying to connect from VS-code Jupyter to an AWS EMR-serverless livy server - a number of members of our team have noticed intermittent hanging.

**To Reproduce**
1. Create an EMR-Serverless application, enable the interactive livy endpoint
2. Configure local sparkmagic kernel using the AWS guide: 
https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/interactive-workloads-livy-endpoints.html
3. For PySpark to show up in VScode-jupyter - replace the python path in kernel.json with an absolute path to the python binary (e.g. on a mac located at `/Users/<username/Library/Jupyter/kernels/pysparkkernel/kernel.json`)
4. Start EMR-Serverless application and then open kernel

**Expected behavior**
Expected: Spark application starts and kernel is able to be used interactively.
Actual: Kernel intermittently hangs indefinitely. It's not every time, but it's often enough to be very annoying. Reconnecting a number of times/restart emr-s application is observed to eventually allow a connection through.

**Versions:**
 - SparkMagic: 0.22.0
 - Livy (if you know it): 0.8.0-incubating
 - EMR-S: emr-7.8.0 (x86_64)
 - Spark: 3.5.4-amzn-0

**Additional context**
On investigating the issue and logging additional information such as the commands being sent to livy, it was observed that VSCode sends some python bootstrapping logic when a kernel is started (likely to enable some of the features like reporting local variables).

It's believed that this logic is causing an error that isn't being gracefully handled during the initial connection and is the source of the hanging. By adding a fix to usercodeparser to ensure any VS Code bootstrapping logic is not sent to livy, the hanging error disappears entirely (from experience)

Another user arrived at a similar solution posted here: https://github.com/microsoft/vscode-jupyter/issues/15977#issuecomment-3116541795

PR #964, opened 2025-10-12
Fix vscode-jupyter hanging by executing vscode commands locally

001ben:
### Description
Implement fix for issue described in https://github.com/jupyter-incubator/sparkmagic/issues/963

Simplest fix is to not ad %%spark to any code containing `_VSCODE_` as this code should only be intended for running locally.

See example VScode bootstrapping code here: https://github.com/jupyter-incubator/sparkmagic/issues/963

### Checklist
- [x] Formatted my code with [`black`](https://black.readthedocs.io/en/stable/index.html)
- [x] If adding a feature, there is an example notebook and/or documentation in the `README.md` file


devstein (commented):
Thank you for the PR @001ben! I don't have local set up to reproduce, but think this looks good. 

I reached out to a few folks that have run into this issue to see if this fix works for them as well.

alex-stankey:
Hello @devstein! Any update on this review? 

I've had the same issue, and it looks like this fix works locally for me.

